PR #28, opened 2019-12-17
textile docker

andrewxhill:
with this setup, you can run,

`docker-compose -f docker-compose.yml up` from the project dir and it will stand up the threads service + a textiled server in docker. 

i've pulled the thread server out of this project, so instead it expects to always just talk to it over grpc. that feels right still.

but probably, the way you'll want to run the project is just,

1. in my PR of go-textile-threads (https://github.com/textileio/go-textile-threads/pull/169) we'll start pushing tag builds to docker-hub
2. then locally, you can just do `docker run -p 9090:9090 -i textile/threads-service` (or whatever name)
3. then in this project just fire it up with `textiled --addrThreadsApi=/ip4/0.0.0.0/tcp/9090` since the port will be available over localhost like that, it'll all just work.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+Dockerfile
+
+.github
+.circleci
+dist
+examples
+.threads*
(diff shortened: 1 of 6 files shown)


sanderpick:
Outdated...

Issue #32, opened 2019-12-20
Move testing token to env var


andrewxhill:
Can you use an ENV var instead? Doesn't feel right to have a param scoped to tests in here... maybe, `TXTL_EMAIL_TOKEN`? Just assume, if that var is found, use it, otherwise, generate a single one.

_Originally posted by @sanderpick in https://github.com/textileio/textile/pull/29_

PR #36, opened 2019-12-23
Uses an enum to set token type, Static or Dynamic (default)

andrewxhill:
fixes https://github.com/textileio/textile/issues/32

let me know how this looks @sanderpick


sanderpick:
Outdated...

PR #46, opened 2020-01-02
Integrate filecoin client and wallet address

asutula:
Changed the focus of this PR a bit to make first things first:

1. First we need to integrate the filecoin client into the project.
2. Next we need to associate each textile project with a filecoin wallet id.

This PR does those two things.

Storing data using the associated wallet will be for the next PR


api/service.go
```diff
@@ -347,7 +349,12 @@ func (s *service) AddProject(ctx context.Context, req *pb.AddProjectRequest) (*p
 		log.Fatal("scope required")
 	}
 
-	proj, err := s.collections.Projects.Create(ctx, req.Name, scope)
+	addr, err := s.fcClient.Wallet.NewWallet(ctx, "bls")
```
asutula: Ok, so not sure how tightly coupled we want project creation to be to filecoin. Right now, it's very tightly coupled: The `fcClient` must exist and it must be pointed at a running filecoin grpc service. Do we want to make this more flexible?

asutula:
Real thing to discuss here is the coupling to the `filecoin` project. The tests currently fail because there is no filecoin service and lotus running.

go.mod
```diff
@@ -31,10 +31,10 @@ require (
 	github.com/rs/cors v1.7.0
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/viper v1.3.2
+	github.com/textileio/filecoin v0.0.0-20200108152132-f1b938b219a6
```
sanderpick: 🎉

core/core.go
```diff
@@ -109,6 +111,11 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 		return nil, err
 	}
 
+	fcClient, err := fc.NewClient(conf.AddrFilecoinApi)
```
sanderpick: I can't remember, does this have a `Close` method?

sanderpick: If so, we should probably close it and `threadsClient` (oops) in the Close method below.

api/service.go
sanderpick: Maybe for now we could just check to see if the `fcClient` is `nil`, if not, then create the address. That would be nice for development / tests. In the future, I could see it being a requirement. Dunno though.

cmd/textiled/main.go
```diff
@@ -142,6 +146,12 @@ func init() {
 		flags["addrGatewayUrl"].DefValue.(string),
 		"Public gateway address")
 
+	// Filecoin settings
+	rootCmd.PersistentFlags().String(
+		"addrFilecoinApi",
+		flags["addrFilecoinApi"].DefValue.(string),
+		"Filecoin gRPC API listen address")
```
sanderpick: Is it the listen address or just the address? Listen sounds like you're binding to it, where here you're just using it to create a client. Just a personal bias, not important.

collections/projects.go
```diff
@@ -11,11 +11,12 @@ import (
...
-	Name    string
-	Scope   string // user or team
-	StoreID string
-	Created int64
+	ID              string
+	Name            string
+	Scope           string // user or team
+	StoreID         string
+	Created         int64
+	FCWalletAddress string
```
sanderpick: 👍

api/service.go
asutula: sounds good. required a little special handling in `GetProject` as well to only make the query for balance if the fc account != "", but done

cmd/textiled/main.go
asutula: good call.

core/core.go
asutula: Yep, good catch. I'll add that and close the `threadsClient` as well.

asutula: Ok, I added the `fcClient.Close()` to the `server.Close()` since the `Server` manages the fc client and `Textile` keeps no reference to it. I added `t.threadsClient.Close()` to `Textile.Close()`.

asutula:
Ok, for now what controls whether or not a filecoin client is created and used for creating/inspecting projects is whether or not `NewTextile()` is called with `AddrFilecoinApi` set in its config struct. The tests don't set it, `textiled` does.

PR #50, opened 2020-01-07
docs: update readme

sanderpick:
Adds a standard readme, removes filecoin spec as that stuff now lives in the filecoin repo.


sanderpick:
Merged quick to expose the readme now that this is public.

PR #52, opened 2020-01-07
More Docker

sanderpick:
Poking around in here some more...
- Cleans up the docker file, updating with latest go-ipfs and removing the contain_daemon script, which wasn't doing anything.
- Leverages cobra environment variables instead of default flags, allowing users to set them to whatever they want. 

In order to use the compose file with the email functionality in textile, create a `.env` file at the repo root:
```
TXTL_LOG_DEBUG=true
TXTL_EMAIL_API_KEY=<MAILGUN_API_KEY>
```
That would also enable debug logging.


sanderpick:
FYI, this requires a docker image with textile _from this branch_, which I didn't push because it gets built by docker hub on merge.

sanderpick:
Merging it.

PR #55, opened 2020-01-10
App tokens and users

### LINKED ISSUE #56: App tokens and users
Developers will create tokens that get embedded in client SDKs. These SDKs will then be able to make threads, store files, etc. associated with a given project.

This will require: 
- Add, list, remove token handling (API, client, cmd)
- App user registration
- API auth interceptors for threads (and future store API)

sanderpick:
Closes #56 

- App token and user collections
- App token add, list, remove API methods
- App token add, list, remove client methods and tests
- App token add, list, remove commands (`textile tokens`)
- Add user registration is handled through the gateway (POST JSON) so as to minimize the (yet to be written) client SDK code (it can have the threads client + a thin wrapper for associating with a project)
- Leverages interceptors to the threads API, check for an app user session (requires https://github.com/textileio/go-threads/pull/211, after this is in, I will update go.mod)


api/client/client_test.go
```diff
@@ -450,44 +462,278 @@ func TestRemoveProject(t *testing.T) {
...
+	})
+
+	t.Run("test remove app token", func(t *testing.T) {
+		if err := client.RemoveAppToken(context.Background(), token.ID, Auth{Token: user.SessionID}); err != nil {
+			t.Fatalf("remove app token should succeed: %v", err)
+		}
+	})
+}
+
+func TestRegisterAppUser(t *testing.T) {
```
sanderpick: This test will give an overview of what's happening here.

api/client/client_test.go
```diff
@@ -450,44 +462,278 @@ func TestRemoveProject(t *testing.T) {
...
+			t.Fatalf("expected status code 200, got %d", res.StatusCode)
+		}
+
+		// Setup a threads client for app user tests.
+		ttarget, err := tutil.TCPAddrFromMultiAddr(conf.AddrThreadsApi)
+		if err != nil {
+			t.Fatal(err)
+		}
+		tclient, err := threadsclient.NewClient(ttarget, grpc.WithInsecure(),
+			grpc.WithPerRPCCredentials(collections.TokenAuth{}))
```
sanderpick: This is meant to emulate a client SDK that would wrap the threads client.

api/client/client_test.go
```diff
@@ -450,44 +462,278 @@ func TestRemoveProject(t *testing.T) {
...
+		defer tclient.Close()
+
+		t.Run("test create new store without client token", func(t *testing.T) {
+			if _, err := tclient.NewStore(context.Background()); err == nil {
+				t.Fatal("test create new store without client token should fail")
+			}
+		})
+
+		t.Run("test create new store with client token", func(t *testing.T) {
+			storeID, err := tclient.NewStore(collections.AuthCtx(context.Background(), session))
```
sanderpick: Here, an app user is using a session from registration to create their own store.

sanderpick:
Some things to consider:
- Generally speaking, what does it mean for an app user to create stuff under a project? 
- Should the user only be able to interact with existing stores, e.g., create / modify models? If we think they should be able to create their own stores (via the app, via the developer's code), we'll have to associate those stores with the project some how. It would be nice if there was a post-request interceptor that could catch whatever was done and connect the dots with the users associated project.

api/client/client_test.go
asutula: Cool. What does it mean to have a user store? Is it somehow restricted access to just that user?

api/pb/api.proto
```diff
@@ -135,4 +157,8 @@ service API {
     rpc GetProject (GetProjectRequest) returns (GetProjectReply) {}
     rpc ListProjects (ListProjectsRequest) returns (ListProjectsReply) {}
     rpc RemoveProject (RemoveProjectRequest) returns (RemoveProjectReply) {}
+
+    rpc AddAppToken (AddAppTokenRequest) returns (AddAppTokenReply) {}
+    rpc ListAppTokens (ListAppTokensRequest) returns (ListAppTokensReply) {}
+    rpc RemoveAppToken (RemoveAppTokenRequest) returns (RemoveAppTokenReply) {}
```
asutula: This looks good for now, but just fyi... I started building grpc components in different packages in the `filecoin` project. Feels pretty nice and keeps the server and client and proto code from getting monolithic. Maybe worth considering at some point.

api/server.go
```diff
@@ -158,11 +155,13 @@ func (s *Server) authFunc(ctx context.Context) (context.Context, error) {
...
 	}
-	newCtx = context.WithValue(newCtx, reqKey("scope"), scope)
 
 	if err := s.service.collections.Sessions.Touch(ctx, session); err != nil {
 		return nil, err
 	}
 
+	newCtx := context.WithValue(ctx, reqKey("session"), session)
+	newCtx = context.WithValue(newCtx, reqKey("user"), user)
+	newCtx = context.WithValue(newCtx, reqKey("scope"), scope)
```
asutula: Want to get a brain dump from you about how `Context` is used by gRPC (or other servers) to manage session data like this. Looks powerful but I don't totally get it.

collections/appusers.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if stat.Message() != s.ErrNotFound.Error() {
+			return
+		}
+	}
+	user = &AppUser{
+		ID:        deviceID,
+		ProjectID: projectID,
+	}
+	if err := u.threads.ModelCreate(ctx, u.storeID.String(), u.GetName(), user); err != nil {
```
asutula: just really love seeing threads used to build Textile! so cool

asutula: we could really start pointing to this repo as an example of using threads

cmd/textiled/main.go
```diff
@@ -240,6 +245,7 @@ var rootCmd = &cobra.Command{
 			EmailFrom:            emailFrom,
 			EmailDomain:          emailDomain,
 			EmailApiKey:          emailApiKey,
+			ThreadsInternalToken: uuid.New().String(),
```
asutula: I see this token spreading deep into things (but maybe not, maybe the other ones are actually user/app user tokens now that I think about it). But what is this one used for?

api/pb/api.proto
sanderpick: Cool, makes sense. Might make more sense over there where there's a clean break between modules.

api/server.go
sanderpick: Cool. I'm mostly just following examples, but there's an `PerRPCCredentials` interface that has this method:
```
// GetRequestMetadata gets the current request metadata, refreshing
// tokens if required. This should be called by the transport layer on
// each request, and the data should be populated in headers or other
// context. If a status code is returned, it will be used as the status
// for the RPC. uri is the URI of the entry point for the request.
// When supported by the underlying implementation, ctx can be used for
// timeout and cancellation. Additionally, RequestInfo data will be
// available via ctx to this call.
// TODO(zhaoq): Define the set of the qualified keys instead of leaving
// it as an arbitrary string.
GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error)
```
This essentially maps context values (k/v pairs on any Go context) to headers.

cmd/textiled/main.go
sanderpick: This is used as a means to identify requests to the thread server _from textile itself_, like projects, teams, etc. Otherwise, those requests would return as unauthenticated. The interceptor simply lets all requests go through if they come with this authorization token, which is only ever visible to `textile` itself (not visible outside the box / process).

andrewxhill:
Looking great! The only thing that occurred to me in this PR was that it might be good to get 

user of cli => (app, project, team)  => users of project/app

terminology clearer. feels like most of this PR could be renamed developers? 

developers (use the CLI) => are part of teams => own apps/projects => service users

sanderpick:
Yeah nice, and can do a renaming as a follow up.

PR #64, opened 2020-01-23
Store Files

### LINKED ISSUE #57: Add basic store endpoint backed by IPFS
The pins will have to be associated with the project or user.

sanderpick:
Closes #57 

Still have to do:
- [x] Add a store command that leverages `client.Store`
- [x] Track stored files in a project
- [x] List stored files (and command)
- [x] Remove files (and command)


.github/workflows/test.yml
```diff
@@ -10,6 +10,15 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    services:
```
sanderpick: Damn, pretty easy to run IPFS in CI like this.

api/pb/api.proto
```diff
@@ -140,6 +140,58 @@ message RemoveAppTokenRequest {
 
 message RemoveAppTokenReply {}
 
+message AddFileRequest {
```
sanderpick: I changed from "store" to "add files", which feels more intuitive on the command line as well.

api/service.go
```diff
@@ -514,6 +525,199 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+	defer close(eventCh)
+	go func() {
+		for e := range eventCh {
+			event, ok := e.(*iface.AddEvent)
+			if !ok {
+				log.Error("unexpected event type")
+				continue
+			}
+			if event.Path == nil { // This is a progress event
+				sendEvent(&pb.AddFileReply_Event{
```
sanderpick: Pumping progress events here.

api/service.go
```diff
@@ -514,6 +525,199 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+	if err != nil {
+		return err
+	}
+
+	_, err = s.collections.Files.Create(server.Context(), pth, name, proj.ID)
+	if err != nil && !strings.Contains(err.Error(), store.ErrUniqueExists.Error()) {
+		return err
+	}
+
+	sendEvent(&pb.AddFileReply_Event{
```
sanderpick: Then, send the last event with the final path.

api/service.go
```diff
@@ -514,6 +525,199 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+	file, err := s.getFileWithScope(ctx, req.Path, scope)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = s.collections.Files.Delete(ctx, file.ID); err != nil {
+		return nil, err
+	}
+
+	if err = s.ipfsClient.Pin().Rm(ctx, path.New(req.Path)); err != nil {
```
sanderpick: Unpin.

collections/files.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/ipfs/interface-go-ipfs-core/path"
+	"github.com/textileio/go-threads/api/client"
+	s "github.com/textileio/go-threads/store"
+)
+
+type File struct {
```
sanderpick: I'd like to add a `Parent` in a follow up PR, which can be used to build a directory structure.

collections/collections.go
```diff
@@ -121,7 +139,12 @@ func (c *Collections) addCollection(ctx context.Context, col Collection, key dat
 		if err != nil {
 			panic(err)
 		}
-		if err = c.threads.RegisterSchema(ctx, storeID.String(), col.GetName(), string(schema)); err != nil {
+		if err = c.threads.RegisterSchema(
+			ctx,
+			storeID.String(),
+			col.GetName(),
+			string(schema),
+			col.GetIndexes()...); err != nil {
```
sanderpick: Indexes are working nice 👍

api/client/client.go
```diff
@@ -161,13 +171,158 @@ func (c *Client) RemoveAppToken(ctx context.Context, tokenID string, auth Auth)
...
+			return nil, err
+		}
+		if err = stream.Send(&pb.AddFileRequest{
+			Payload: &pb.AddFileRequest_Chunk{
+				Chunk: buf[:n],
+			},
+		}); err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, err
```
jsign: Should in both returns call `stream.CloseSend()`?

api/service.go
jsign: `sendEvent` will only log if an error happened while sending. That sounds good for progress events since they aren't a critical thing. But I wonder what might happen if this last event fails on sending. Apart from logging it, this `sendEvent` assumes all goes well. 

Also, the `s.collections.Files.Create` was already called at this point, so that could be a problem?

Considering not missing anything; possible solution: 
- make `sendEvent` return an `error`. While sending progress, ignore it. Not ignore it on this last send. 
- Switch first sending event, and then `.Create()` call? Not sure about this last point since maybe both orderings are problematic in both ways, so just mentioning it if you have a more clear picture if one of them might be better than the other.

Extra question: if last `sendEvent` fails, the channel with receiver will close anyway and they'll understand that something bad happened since was closed before receiving the last result, right? (just checking they won't block forever waiting for it).

api/service.go
jsign: Thought: I guess GC is mainly the responsibility of the hosting IPFS actor.

collections/files.go
```diff
@@ -0,0 +1,89 @@
...
+	return &File{}
+}
+
+func (f *Files) GetIndexes() []*s.IndexConfig {
+	return []*s.IndexConfig{{
+		Path:   "Path",
+		Unique: true,
+	}, {
+		Path: "ProjectID",
+	}}
```
jsign: Nice

jsign: @carsonfarmer ^

api/client/client.go
```diff
@@ -161,13 +171,158 @@ func (c *Client) RemoveAppToken(ctx context.Context, tokenID string, auth Auth)
...
+// AddFileOptions defines options for adding a file.
+type AddFileOptions struct {
+	Progress chan int64
+}
+
+// AddFileOption specifies an option for adding a file.
+type AddFileOption func(*AddFileOptions)
+
+// WithProgress writes progress updates to the given channel.
+func WithProgress(ch chan int64) AddFileOption {
```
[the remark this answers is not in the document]
sanderpick: 👍

api/client/client.go
```diff
@@ -161,13 +171,158 @@ func (c *Client) RemoveAppToken(ctx context.Context, tokenID string, auth Auth)
...
+
+	stream, err := c.c.AddFile(authCtx(ctx, auth))
+	if err != nil {
+		return nil, err
+	}
+	if err = stream.Send(&pb.AddFileRequest{
+		Payload: &pb.AddFileRequest_Header_{
+			Header: &pb.AddFileRequest_Header{
+				Name:      filepath.Base(file.Name()),
+				ProjectID: projID,
```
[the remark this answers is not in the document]
sanderpick: Sounds good 👌. This would also allow the remote to tell the client it already has the file _before_ uploading it. Will make an issue.

api/service.go
sanderpick: >     * make `sendEvent` return an `error`. While sending progress, ignore it. Not ignore it on this last send.

Good call! 

>     * Switch first sending event, and then `.Create()` call? Not sure about this last point since maybe both orderings are problematic in both ways, so just mentioning it if you have a more clear picture if one of them might be better than the other.

Right yeah, I was wondering the same about the order of things. Not ideal in either case. Though, the remote uses the event with final `Path` to indicate to the client that they can stop receiving. In this setup, the `sendEvent` must come last, otherwise the client would have already left the room by time the remote calls `Create`, ignoring any errors that happen there. 

> Extra question: if last `sendEvent` fails, the channel with receiver will close anyway and they'll understand that something bad happened since was closed before receiving the last result, right? (just checking they won't block forever waiting for it).

In this case, (currently) the remote would end as normal, causing an EOF on the client. This sends to its wait channel, which is used to wrap things up. However, we can catch the error as mentioned above, in which case it would be used to wrap things up on the client.

.github/workflows/test.yml
sanderpick: Totally, maybe it just used compose under the hood, or at least puts all services on the same network. @carsonfarmer you might try this regarding your questions yesterday. You could probably run two `go-threads` services this way.

api/service.go
jsign: SGTM

api/client/client.go
sanderpick: Ah yeah, good catch.

sanderpick:
Thanks for the solid review as always, @jsign 🙏

andrewxhill:
Very nice feature! 

storeFile

it's going to pair so nicely with threads, some early testers already asking about it

sanderpick:
👍! Btw, it does use threads to track the files internally.

PR #74, opened 2020-02-06
Folders

### LINKED ISSUE #68: Include parent in File model
The `File` model can have a parent field, which can be used to indicate that it's part of a directory.

### LINKED ISSUE #69: Add an entire directory of files
This is a prereq to a `push` style command that syncs an entire project directory, which is a prereq to a public folder.

sanderpick:
Closes #68 (different implementation than originally expected)
Closes #69 (different implementation than originally expected)

This adds folder handling to the Files API, which involved removing the `Files` collection and the `File` model in place of `Folders` and a `Folder` model. Files are handled under the hood by the IPFS Files API. Folders are just IPFS Directories. 

### Demos
- Adding folders and files: https://asciinema.org/a/cBKjOToGtrUOFQbE5HZfPXaYi
- Inspecting folders and catting files: https://asciinema.org/a/1hvJGUDrgFLcNLJAiGRcrfRq1
- Removing files: https://asciinema.org/a/f1Ci5O6pDUuT7EhQkOlvRIwwA (notice folder root path changes)

### Next Steps
This PR adds some low-ish-level APIs and commands for managing files and folder. Next, I'd like to add some higher level commands for syncing folders (these may require some merge handling, not sure yet):
- `textile push <folder>` (push latest)
- `textile pull <folder>` (pull whatever the remote has)


api/client/client_test.go
```diff
@@ -773,26 +776,172 @@ func TestClient_ListFiles(t *testing.T) {
...
+		}
+		defer file.Close()
+		progress := make(chan int64)
+		go func() {
+			for p := range progress {
+				t.Logf("progress: %d", p)
+			}
+		}()
+		pth, err := client.AddFile(
+			context.Background(), "myfolder/more/stuff/file1.jpg", file, Auth{Token: user.SessionID},
```
sanderpick: Adding a file under a nested path. The path will be created if it doesn't yet exist.

api/client/client.go
```diff
@@ -171,6 +169,37 @@ func (c *Client) RemoveAppToken(ctx context.Context, tokenID string, auth Auth)
 	return err
 }
 
+// AddFolder adds a folder by name.
+func (c *Client) AddFolder(ctx context.Context, projID, name string, public bool, auth Auth) (*pb.AddFolderReply, error) {
```
sanderpick: Here, `public` indicates that the folder should be viewable over the public gateway. Currently, not handled.

api/client/client.go
```diff
@@ -216,8 +240,7 @@ func (c *Client) AddFile(
 	if err = stream.Send(&pb.AddFileRequest{
 		Payload: &pb.AddFileRequest_Header_{
 			Header: &pb.AddFileRequest_Header{
-				Name:      filepath.Base(file.Name()),
-				ProjectID: projID,
+				Path: filePath,
```
sanderpick: This path (and others used to interact with files) is of the form: `<folder_name>/path/to/file/name.ext`.

cmd/textile/files.go
```diff
@@ -83,45 +86,100 @@ var addFileCmd = &cobra.Command{
...
+			data[i] = []string{
+				e.Path, strconv.Itoa(int(e.Size)), strconv.FormatBool(e.IsDir),
+			}
 		}
-		cmd.RenderTable([]string{"name", "path"}, data)
+		cmd.RenderTable([]string{"path", "size", "dir"}, data)
 	}
 
-	cmd.Message("Found %d files", aurora.White(len(files.List)).Bold())
+	cmd.Message("Found %d files", aurora.White(len(folder.Entries)).Bold())
```
sanderpick: Files will include directories.

cmd/textile/main.go
```diff
@@ -61,6 +61,7 @@ var (
 	cmdTimeout     = time.Second * 10
 	loginTimeout   = time.Minute * 3
 	addFileTimeout = time.Minute
+	getFileTimeout = time.Minute
```
sanderpick: Dunno what this should really be. Maybe no timeout? User should be able to download massive files. Same goes for `addFileTimeout`.

collections/collections.go
```diff
@@ -25,7 +25,7 @@ var (
 	dsAppTokensKey = datastore.NewKey("/apptokens")
 	dsAppUsersKey  = datastore.NewKey("/appusers")
 
-	dsFilesKey = datastore.NewKey("/files")
+	dsFoldersKey = datastore.NewKey("/folders")
```
sanderpick: We only need to track folders in the thread. The files are available from the IPFS API. This also helps decouple Threads from the underlying Pinning API... we only have to track to roots, not each file.

collections/folders.go
```diff
@@ -0,0 +1,105 @@
...
+func (f *Folders) GetInstance() interface{} {
+	return &Folder{}
+}
+
+func (f *Folders) GetIndexes() []*s.IndexConfig {
+	return []*s.IndexConfig{{
+		Path: "Path",
+	}, {
+		Path:   "Name",
+		Unique: true,
```
sanderpick: A project can only have one folder by a given name. Question: Maybe we can use "bucket" instead of "folder"? Too close to S3?

api/service.go
```diff
@@ -524,6 +528,158 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+		log.Fatal("scope required")
+	}
+	proj, err := s.getProjectForScope(ctx, req.ProjectID, scope)
+	if err != nil {
+		return nil, err
+	}
+
+	pth, err := s.ipfsClient.Unixfs().Add(
+		ctx,
+		ipfsfiles.NewMapDirectory(map[string]ipfsfiles.Node{}),
```
sanderpick: Here we just create an empty directory.

api/service.go
```diff
@@ -524,6 +528,158 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+		Name:      folder.Name,
+		Public:    folder.Public,
+		ProjectID: folder.ProjectID,
+		Created:   folder.Created,
+		Entries:   entries,
+	}, nil
+}
+
+func dirToPbSlice(node ipfsfiles.Node) (entries []*pb.GetFolderReply_Entry, err error) {
+	err = ipfsfiles.Walk(node, func(path string, n ipfsfiles.Node) error {
```
sanderpick: When getting a folder (used to list files in the CLI), walk the node structure and collect a flat list of paths.

api/service.go
```diff
@@ -524,6 +528,158 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+	folder, err := s.getFolderWithScope(ctx, req.Name, scope)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = s.collections.Folders.Delete(ctx, folder.ID); err != nil {
+		return nil, err
+	}
+
+	if err = s.ipfsClient.Pin().Rm(ctx, path.New(folder.Path)); err != nil {
```
sanderpick: There's only one pin at the root that needs to be removed.

api/service.go
```diff
@@ -621,15 +777,22 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
+		options.Unixfs.Pin(false),
 		options.Unixfs.Progress(true),
 		options.Unixfs.Events(eventCh))
 	if err != nil {
 		return err
 	}
 
-	_, err = s.collections.Files.Create(server.Context(), pth, name, proj.ID)
-	if err != nil && !strings.Contains(err.Error(), store.ErrUniqueExists.Error()) {
+	dirpth, err := s.ipfsClient.Object().
```
sanderpick: Add the link under the node. The trick here is to use to object API, not the Files API 🤷‍♂

api/service.go
```diff
@@ -621,15 +777,22 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
+		AddLink(server.Context(), folderPath, fileName, pth, options.Object.Create(true))
+	if err != nil {
+		return err
+	}
+	if err = s.ipfsClient.Pin().Update(server.Context(), folderPath, dirpth); err != nil {
```
sanderpick: Move the pin to the root.

api/service.go
```diff
@@ -621,15 +777,22 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
+		return err
+	}
+	if err = s.collections.Folders.UpdatePath(server.Context(), folder, dirpth); err != nil {
```
sanderpick: Finally, update the folder path to the new root.

api/service.go
```diff
@@ -703,16 +902,21 @@ func (s *service) RemoveFile(ctx context.Context, req *pb.RemoveFileRequest) (*p
...
-	file, err := s.getFileWithScope(ctx, req.Path, scope)
+	folderName, fileName := parsePath(req.Path)
+	folder, err := s.getFolderWithScope(ctx, folderName, scope)
 	if err != nil {
 		return nil, err
 	}
+	folderPath := path.New(folder.Path)
 
-	if err = s.collections.Files.Delete(ctx, file.ID); err != nil {
+	dirpth, err := s.ipfsClient.Object().RmLink(ctx, folderPath, fileName)
```
sanderpick: Similar to adding a file, remove the link and then move the pin to the new root.

api/service.go
```diff
@@ -536,19 +692,19 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
-		name = payload.Header.Name
+		filePath = payload.Header.Path
 	default:
-		return fmt.Errorf("project ID is required")
+		return fmt.Errorf("add file header is required")
 	}
-
-	proj, err := s.getProjectForScope(server.Context(), projID, scope)
+	folderName, fileName := parsePath(filePath)
+	folder, err := s.getFolderWithScope(server.Context(), folderName, scope)
```
sanderpick: Here we have the folder name from the path and can see if the current user session is authorized to manage this folder.

api/service.go
jsign: So `node` has the real bytes data of files in the folder?

Curious since walking seems like a function that doesn't use some API to make this incremental, so I guess is all the data? (that means... GBs if the folder is big? or is a light structure?)

jsign: Nice

api/service.go
```diff
@@ -524,6 +528,158 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
...
+	for i, folder := range folders {
+		node, err := s.ipfsClient.Unixfs().Get(ctx, path.New(folder.Path))
+		if err != nil {
+			return nil, err
+		}
+		list[i], err = folderToPbFolder(folder, node)
+		if err != nil {
+			return nil, err
+		}
+		node.Close()
```
jsign: Should be `defer`ed and moved above to be called on early returns?

api/service.go
jsign: Does unpinning also unpin _orphaned_ children of the unpinned cid?

api/service.go
jsign: Huh?, that's weird. I wouldn't expect that adding a Link in some other API will have the same meaning in UnixFS. I mean, sounds reasonable but weird.

Sounds that would be natural to have some `UnixFS().Link(..)`. :/

api/service.go
```diff
@@ -621,15 +777,22 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
 	pth, err := s.ipfsClient.Unixfs().Add(
 		server.Context(),
 		ipfsfiles.NewReaderFile(reader),
-		options.Unixfs.Pin(true),
+		options.Unixfs.Pin(false),
```
jsign: Uhm, so the root-folder was pinned at creation but a priori here isn't pinned.

api/service.go
jsign: Is this because adding a link changed the CID, so the original one should be unpinned because has _stalled_ data?

api/service.go
```diff
@@ -640,10 +803,20 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
 	return nil
 }
 
+func parsePath(pth string) (folder, name string) {
+	pth = strings.TrimPrefix(pth, "/")
+	parts := strings.Split(pth, "/")
+	folder = parts[0]
+	if len(parts) > 1 {
+		name = filepath.Join(parts[1:]...)
+	}
```
jsign: Maybe `strings.SplitN(..., 2)` can be used to avoid the `if`.

api/service.go
```diff
@@ -652,47 +825,73 @@ func (s *service) GetFile(ctx context.Context, req *pb.GetFileRequest) (*pb.GetF
...
+	folder, err := s.getFolderWithScope(ctx, folderName, scope)
+	if err != nil {
+		return nil, err
+	}
+	pth := path.New(filepath.Join(folder.Path, fileName))
+	if err = pth.IsValid(); err != nil {
+		return nil, err
+	}
+
+	node, err := s.ipfsClient.Unixfs().Get(ctx, pth)
```
jsign: So this traverses the `Links` added with the `Object().AddLink` and resolves the path starting from the project folder?

api/service.go
```diff
@@ -652,47 +825,73 @@ func (s *service) GetFile(ctx context.Context, req *pb.GetFileRequest) (*pb.GetF
...
+	node, err := s.ipfsClient.Unixfs().Get(server.Context(), pth)
 	if err != nil {
-		return nil, err
-	}
-	list := make([]*pb.GetFileReply, len(files))
-	for i, file := range files {
-		list[i] = fileToPbFile(file)
+		return err
 	}
+	defer node.Close()
```
jsign: Nit: Maybe extract to avoid DRY in some parts?.

api/service.go
```diff
@@ -652,47 +825,73 @@ func (s *service) GetFile(ctx context.Context, req *pb.GetFileRequest) (*pb.GetF
...
 
-	return &pb.ListFilesReply{List: list}, nil
+	file := ipfsfiles.ToFile(node)
```
jsign: Oh, this doesn't return an error if this path happens to be a directory?
Or maybe fails earlier?

api/service.go
asutula: I've found the go `path` package useful for these types of things.

cmd/textile/main.go
asutula: Yea seems like something really high like a day or three or no timeout.

api/client/client.go
```diff
@@ -171,6 +169,37 @@ func (c *Client) RemoveAppToken(ctx context.Context, tokenID string, auth Auth)
...
+
+// ListFolders returns a list of folders under the current project.
+func (c *Client) ListFolders(ctx context.Context, projID string, auth Auth) (*pb.ListFoldersReply, error) {
+	return c.c.ListFolders(authCtx(ctx, auth), &pb.ListFoldersRequest{
+		ProjectID: projID,
+	})
+}
+
+// RemoveFolder removes a folder by name.
+func (c *Client) RemoveFolder(ctx context.Context, name string, auth Auth) error {
```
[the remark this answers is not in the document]
sanderpick: Good call, will do.

api/client/client_test.go
sanderpick: Cool, will do. Yep, creating the root folder must be done explicitly... but actually, I guess there's no reason we couldn't do that. This would be sorta like `git checkout -b mybranch` in that your creating a new branch before checking it out. We'd need a way to manipulate a folder's public status after creating it, maybe `textile folders <name> public true` or something.

api/pb/api.proto
```diff
@@ -140,15 +140,59 @@ message RemoveAppTokenRequest {
...
 
+message AddFolderRequest {
+    string name = 1;
+    bool public = 2;
+    string projectID = 3;
+}
+
+message AddFolderReply {
+    string ID = 1;
+    string path = 2;
```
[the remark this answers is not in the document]
sanderpick: Yep, correct.

api/service.go
```diff
@@ -524,6 +528,158 @@ func (s *service) RemoveAppToken(ctx context.Context, req *pb.RemoveAppTokenRequ
 	return &pb.RemoveAppTokenReply{}, nil
 }
 
+// AddFolder handles an add folder request.
+func (s *service) AddFolder(ctx context.Context, req *pb.AddFolderRequest) (*pb.AddFolderReply, error) {
+	log.Debugf("received add folder request")
```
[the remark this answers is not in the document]
sanderpick: In the sense that, we should ensure folders aren't added concurrently? I suppose that applies to files as well. I can make an issue.

api/service.go
sanderpick: Yes, this will always return the same node since it's just an empty directory. Below, we pin it, but it won't be pinned again if it's already pinned, say, by another project. However, we should ensure that it's not _unpinned_ since we don't want to unpin if another project is relying on it. In theory, this applies to any directory... perhaps we should explore some way of making all directories not conflict across projects. There's a non-zero chance that you and I create the exact same directory (name and contents).

api/service.go
sanderpick: Bytes aren't included—the node will be of type _File_ or _Directory_, both of which are `io.Reader`.

api/service.go
sanderpick: Ah yeah, good catch. I'd opt for just calling it explicitly in the return case to avoid the `defer` in loop, sound okay?

api/service.go
sanderpick: The children don't ever get pinned. There could be a cross folder file that is under two different pins from different roots. The second one would remain as expected.

api/service.go
```diff
@@ -621,15 +782,33 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
 	pth, err := s.ipfsClient.Unixfs().Add(
```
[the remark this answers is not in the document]
sanderpick: Yep!

api/service.go
sanderpick: This just means, don't create an _additional_ pin for the actual file since below we're going to place a pin above it at the new folder root.

api/service.go
sanderpick: Agreed, yeah. Not very intuitive! `UnixFS` just has `Add`, `Get`, and `Ls`. `Add` doesn't let you manipulate existing nodes.

api/service.go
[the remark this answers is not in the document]
jsign: Mosty two additions where one might cancel other since was based in a root that will change or be unpined or similar? Actually was just a note yep.

api/service.go
sanderpick: Yep!

api/service.go
jsign: Uhm, cool. I wonder how that is resolved. Maybe the `ipfsfiles.Node` hold a reference to the `ipfshttpclient` (or DagResolver) that was used to fetch them? Sounds like magic.

api/service.go
jsign: Oh, I see. So this operation always returns the same empty folder for all _starting_ new folders.
But as soon as they add files, that _node_ would be _mutated_ and will start becoming an _onwed noded_ of the project.  I guess the `Pin().Update(..)` call to move the pinning wouldn't hurt. Since the next created folder will create this empty directory again. (I hope this makes sense).

Regarding creating identical graphs in different folders, maybe instead of creating an empty folder to start, it can have a file with some random content, and that file is always ignored in listings or can't be removed. That would make the root cid always unique per project-folder. Something like creating a `.git` when you do `git init` (first idea in my mind).

api/service.go
jsign: Yep.

api/service.go
jsign: Yep, actually I tried to exclude that case with the _ophaned_ word (which I don't know if is correct, took it from ORM worlds :P).

If a child is orphaned (no root links to it), I guess should be unpinned. I guess this indirectly will behave as how IPFS understands unpinning hierarchies, so I guess its ok.

api/service.go
sanderpick: Doesn't `strings.SplitN(..., 2)` return _at most 2 substrings_? As in, it still could be just one?

api/service.go
jsign: So to clarify: pinning the base directory is recursive.

sanderpick: Yep. To further add confusion, the pins under the root are not normal type pins... they are _recursive_ type pins, which makes them easy to remove when you remove the root pin.

api/service.go
jsign: I made my comment quite sloppy, I meant the thing inside the if.
Yep, will return at most two substrings. 
The second one would be the `filepath.Join(...)` for free. (`Split` separated things that it later rejoins again).

cmd/textile/main.go
sanderpick: Cool, I'll make them a day for now.

api/service.go
sanderpick: Nice, I like that! I'll create an issue.

api/service.go
sanderpick: Yep, they have this concept of a "recursive" type pin which is used to unpin as well.

api/service.go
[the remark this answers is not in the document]
sanderpick: Cool- made an issue regarding project locks.

api/service.go
sanderpick: Must be something like that... I just embraced the magic! Wild though.

api/service.go
sanderpick: This will just get the node at the path. If it's a directory, you can call `Ls` on it to list it links.

api/service.go
sanderpick: Oh yeah! Of course 🤦‍♂

api/service.go
sanderpick: Yep, will do 👍

api/service.go
sanderpick: Oops! Yep, file will be nil if this is a directory (this method is just a helper that does an assertion). I added a `nil` check.

sanderpick:
Thanks for the review, guys! I've addresses or created issues for the relevant comments.

PR #79, opened 2020-02-09
Push 'n pull bucket paths

### LINKED ISSUE #75: Auto-create root folders if they don't exist
Currently, creating the root folder must be done explicitly... but actually, I guess there's no reason we couldn't do that. This would be sorta like git checkout -b mybranch in that your creating a new branch before checking it out. We'd need a way to manipulate a folder's public status after creating it, maybe textile folders <name> public true or something.

### LINKED ISSUE #76: Avoid folder conflicts across projects
From @jsign 

> Regarding creating identical graphs in different folders, maybe instead of creating an empty folder to start, it can have a file with some random content, and that file is always ignored in listings or can't be removed. That would make the root cid always unique per project-folder. Something like creating a .git when you do git init (first idea in my mind).

### LINKED ISSUE #78: Push and pull folders
https://github.com/textileio/textile/pull/74 added some low-ish-level APIs and commands for managing files and folder. Next, we need some higher level commands for syncing folders (these may require some merge handling, not sure yet). If we also moved to `bucket(s)` in the CLI, I think we could consolidate the `folders` and `files` commands, replacing it all with:

- `textile buckets (ls/list)`  (list all top level dirs, aka, buckets)
- `textile buckets inspect <path/to/file/or/dir>`  (show file or entire directory details, including root bucket)
- `textile buckets push <path/to/file/or/dir>`  (add a file or entire directory to a bucket, create bucket if needed)
- `textile buckets rm <path/to/file/or/dir>` (remove a file or entire directory from a bucket, delete bucket if empty)
- `textile buckets pull <path/to/file/or/dir>` (pull down the file or entire directory, including root bucket)

sanderpick:
Closes #75 
Closes #76 
Closes #78

After feeling the UX of the `folders` and `files` commands, I wanted to try an experience where the user doesn't have to know / care if a path is a bucket root or just a child path. The goal of the UX is:
- Push anything (files/dirs) to any path, regardless of whether or not the path already exists
- List anything (files/dirs) at any path, showing relevant info
- Pull anything in the same fashion, writing files/dirs locally
- Remove / delete any path (the root bucket will be auto-delete if its empty)

### The `push` help docs
```
Push files and directories to a bucket path. All paths will be created if they don't exist.

File structure is mirrored in the bucket. For example, given the directory:
    foo/one.txt
    foo/bar/two.txt
    foo/bar/baz/three.txt

These 'push' commands result in the following bucket structures.

'textile buckets push foo mybuck':
    mybuck/foo/one.txt
    mybuck/foo/bar/two.txt
    mybuck/foo/bar/baz/three.txt

'textile buckets push foo/bar mybuck':
    mybuck/bar/two.txt
    mybuck/bar/baz/three.txt

'textile buckets push foo/bar/baz mybuck':
    mybuck/baz/three.txt

'textile buckets push foo/bar/baz/three.txt mybuck':
    mybuck/three.txt

'textile buckets push foo/* foo':
    foo/one.txt
    foo/bar/two.txt
    foo/bar/baz/three.txt
```

###  The `pull` help docs
```
Pull files and directories from a bucket path. All paths will be created if they don't exist.

Bucket structure is mirrored locally. For example, given the bucket:
    foo/one.txt
    foo/bar/two.txt
    foo/bar/baz/three.txt

These 'pull' commands result in the following local structures.

'textile buckets pull foo mydir':
    mydir/foo/one.txt
    mydir/foo/bar/two.txt
    mydir/foo/bar/baz/three.txt

'textile buckets pull foo/bar mydir':
    mydir/bar/two.txt
    mydir/bar/baz/three.txt

'textile buckets pull foo/bar/baz mydir':
    mydir/baz/three.txt

'textile buckets pull foo/bar/baz/three.txt mydir':
    mydir/three.txt

'textile buckets pull foo .':
    foo/one.txt
    foo/bar/two.txt
    foo/bar/baz/three.txt
```


api/service.go
```diff
@@ -541,148 +585,147 @@ func (s *service) AddFolder(ctx context.Context, req *pb.AddFolderRequest) (*pb.
...
-		Name:      folder.Name,
-		Public:    folder.Public,
-		ProjectID: folder.ProjectID,
-		Created:   folder.Created,
-		Entries:   entries,
-	}, nil
-}
-
-func dirToPbSlice(node ipfsfiles.Node) (entries []*pb.GetFolderReply_Entry, err error) {
-	err = ipfsfiles.Walk(node, func(path string, n ipfsfiles.Node) error {
```
sanderpick: After playing around with listing paths, I realized a full walk is not really what you want. Just list directories as single items and let the client walk if needed.

api/service.go
```diff
@@ -785,88 +834,76 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
 	if err != nil {
 		return nil, err
 	}
-	defer node.Close()
 
-	size, err := node.Size()
+	pth, err := s.ipfsClient.Unixfs().Add(
+		ctx,
+		ipfsfiles.NewMapDirectory(map[string]ipfsfiles.Node{
+			".seed": ipfsfiles.NewBytesFile(seed), // Ensure the directory is unique w/ reasonable certainty
```
sanderpick: Here's @jsign's random file idea.

api/service.go
```diff
@@ -785,88 +834,76 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
+			Public:  false,
+		},
 	}); err != nil {
 		return err
 	}
 
-	log.Debugf("added file %s to folder with new path: %s", pth.String(), dirpth.String())
+	log.Debugf("pushed %s to bucket: %s", fileName, buck.Name)
 	return nil
 }
```
jsign: Q: `PushBucketPath` should be understood as an overriding operation. What it was asked, will be done.

api/service.go
```diff
@@ -893,91 +930,60 @@ func (s *service) CatFile(req *pb.CatFileRequest, server pb.API_CatFileServer) e
...
-}
 
-// getProjectForScope returns a project if the scope is authorized.
-func (s *service) getProjectForScope(ctx context.Context, projID, scope string) (*c.Project, error) {
-	proj, err := s.collections.Projects.Get(ctx, projID)
-	if err != nil {
-		return nil, err
+	var cnt int
+	for l := range links {
+		fmt.Println(l)
```
jsign: Delete?

api/service.go
```diff
@@ -893,91 +930,60 @@ func (s *service) CatFile(req *pb.CatFileRequest, server pb.API_CatFileServer) e
...
-	return team, nil
-}
 
-// getProjectForScope returns a project if the scope is authorized.
-func (s *service) getProjectForScope(ctx context.Context, projID, scope string) (*c.Project, error) {
-	proj, err := s.collections.Projects.Get(ctx, projID)
-	if err != nil {
-		return nil, err
+	var cnt int
+	for l := range links {
```
jsign: Uhm, I wonder how the creators of the API thought about a closed channel because there was some error. Seems that that case can't be distinguished. I don't see some `.Err()` (similar to how it's done in other places).
Anyway, not much we can do. But mentioning it because I may be missing something.

api/service.go
```diff
@@ -893,91 +930,60 @@ func (s *service) CatFile(req *pb.CatFileRequest, server pb.API_CatFileServer) e
...
+	buckName, fileName := parsePath(req.Path)
+	buck, err := s.getBucketWithScope(ctx, buckName, scope)
 	if err != nil {
 		return nil, err
 	}
-	folderPath := path.New(folder.Path)
 
-	dirpth, err := s.ipfsClient.Object().RmLink(ctx, folderPath, fileName)
+	buckPath := path.New(buck.Path)
+	dirpth, err := s.ipfsClient.Object().RmLink(ctx, buckPath, fileName)
```
jsign: I wonder if we should check that someone isn't deleting `.seed`.
This check might also apply to pushing that file. Despite its main intention was to make buckets unique per project, if we don't error in that case we might get some unwanted consequence in a later _Remove_ operation. Since we consider that `.seed` is ephemeral and will be wiped-out if its the remaining file in the folder.

I guess `.seed` is fine, but maybe prefixing it with something extra may lower the chances that someone has some real `.seed` file and this gets in their way. e.g: `.textilebucketseed`?.

cmd/textile/buckets.go
```diff
@@ -0,0 +1,385 @@
...
+
+		for i := range names {
+			addFile(projectID, names[i], paths[i])
+		}
+
+		cmd.Success("Pushed %d files to %s", aurora.White(len(names)).Bold(), aurora.White(bucketPath).Bold())
+	},
+}
+
+// @todo: Support Stdin
```
jsign: Nice idea.

cmd/textile/buckets.go
```diff
@@ -0,0 +1,385 @@
...
+		api.WithPushProgress(progress)); err != nil {
+		cmd.Fatal(err)
+	}
+	bar.Finish()
+}
+
+var pullBucketPathCmd = &cobra.Command{
+	Use:   "pull",
+	Short: "Pull a bucket path",
+	Long: `Pull files and directories from a bucket path. All paths will be created if they don't exist.
```
jsign: Nit: Clarifying what would happen if that file already exist.

api/pb/api.proto
```diff
@@ -255,13 +235,9 @@ service API {
...
     rpc RemoveAppToken (RemoveAppTokenRequest) returns (RemoveAppTokenReply) {}
 
-    rpc AddFolder (AddFolderRequest) returns (AddFolderReply) {}
-    rpc GetFolder (GetFolderRequest) returns (GetFolderReply) {}
-    rpc ListFolders (ListFoldersRequest) returns (ListFoldersReply) {}
-    rpc RemoveFolder (RemoveFolderRequest) returns (RemoveFolderReply) {}
-
-    rpc AddFile (stream AddFileRequest) returns (stream AddFileReply) {}
-    rpc GetFile (GetFileRequest) returns (GetFileReply) {}
-    rpc CatFile (CatFileRequest) returns (stream CatFileReply) {}
```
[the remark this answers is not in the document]
sanderpick: Totally, and if the path is a directory, maybe we could just "cat" the raw node or return an error saying "blah is a directory".

api/service.go
```diff
@@ -541,148 +585,147 @@ func (s *service) AddFolder(ctx context.Context, req *pb.AddFolderRequest) (*pb.
...
-	return
+	}
+	return item, nil
 }
 
-// ListFolders handles a list folders request.
-func (s *service) ListFolders(ctx context.Context, req *pb.ListFoldersRequest) (*pb.ListFoldersReply, error) {
-	log.Debugf("received list folders request")
+// GetBucketPath handles a get bucket path request.
+func (s *service) GetBucketPath(ctx context.Context, req *pb.GetBucketPathRequest) (*pb.GetBucketPathReply, error) {
```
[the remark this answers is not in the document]
sanderpick: Yep. Another distinction is that `ListBuckets` won't ever return a top-level node w/ entries, just the entries.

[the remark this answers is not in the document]
sanderpick: Nice, sounds good. Actually, we could even consolidate them into one, such that an empty path is taken to be the "virtual" project root. The entry there could be returned as empty with a list.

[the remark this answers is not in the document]
sanderpick: (As in, consolidate `ListBuckets` and `GetBucketPath` into a single `ListBucketPath`)

api/service.go
sanderpick: Oopl, yep.

api/service.go
sanderpick: Yeah, strange. The API used to return a list of links. Not sure why they changed to channel.

api/service.go
```diff
@@ -693,31 +736,37 @@ func (s *service) AddFile(server pb.API_AddFileServer) error {
...
-	folderName, fileName := parsePath(filePath)
-	folder, err := s.getFolderWithScope(server.Context(), folderName, scope)
+	buckName, fileName := parsePath(filePath)
+	buck, err := s.getBucketWithScope(server.Context(), buckName, scope)
 	if err != nil {
-		return err
+		if status.Convert(err).Code() != codes.NotFound {
+			return err
+		}
+		buck, err = s.createBucket(server.Context(), buckName, false, projectID)
```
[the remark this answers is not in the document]
sanderpick: Yep. It's not totally clear to me how to deal with "public" yet. I think a public bucket just means it's viewable at a static path over the gateway. However, some addition thoughts here / a proposal: 

It would be nice to be able to mark _any_ path as public, much like S3. This might involve an additional badger key space in `textile` that just tracks a path (as key) and a boolean public value. So, everything is implied to be private, sans for the keys given in the "public" keyspace. When traversing files/dirs in the gateway, a fast check would be made to ensure a path or a parent of the path has a public entry.

api/service.go
sanderpick: Yep. As a future thing, maybe we can enable a "force" type flag that is required to overwrite paths.

api/service.go
sanderpick: Ah yeah, good catch. I'll add some additional checks for deletes and pushes, and change the name.

cmd/textile/buckets.go
sanderpick: Hmm, I need to check what happens actually (and make a note in the doc). Most likely it will currently error. If so, I'll make it overwrite in the meantime... the "force" idea would be good here as well.

api/service.go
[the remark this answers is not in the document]
jsign: Sounds good.
Today I was also thinking about if the `.seed` the idea can be extended further and enable folder-configuration _embedded_ in the filesystem. It may serve for permissions, or even flagging it _public_ or _private_ too. 

So, someone can change the visibility just by editing the folder configuration file?
Maybe that's easier for people to change things since is just editing the file and pushing changes. But that would add some work to: validate changes, check format, etc.

Just brainstorming too.

api/service.go
jsign: Yeah, overriding policies is another big story. 

Brainstorm: An intermediate idea before some fully-mergable magic algorithm: send the last known bucket root while pushing... the server can check is overriding from that rootcid. If that isn't the case, fail since will override some not-known other change. (Similar of a fast-forward check, if not -> fail). Sounds that might cover most of the bad-cases and would be quite easy.

cmd/textile/buckets.go
jsign: Sounds good!

sanderpick: Ok, currently everything is overwritten locally. I added a note to reflect this.

api/service.go
sanderpick: Nice, I like it. I'll create a ticket.

api/service.go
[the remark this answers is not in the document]
sanderpick: Nice, that does sound like a more extensible solution. Quite like S3 bucket policy files: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html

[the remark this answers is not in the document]
sanderpick: Created https://github.com/textileio/textile/issues/84 to continue convo / work.

PR #90, opened 2020-02-11
Gateway Buckets

### LINKED ISSUE #81: Consolidate ListBuckets and GetBucketPath
... into a single `ListBucketPath`. An empty path can be taken to be the "virtual" project root. The entry there could be returned as empty with a list.

### LINKED ISSUE #86: View buckets over the gateway
This will involve mapping the dynamic bucket path to a file browsing view.

Related to bucket policies: #84

sanderpick:
Closes #81 
Closes #86 

This renders bucket files and directories. 

### Top-level buckets
Screenshot 2020-02-10 23 00 56

### Bucket paths
Screenshot 2020-02-10 23 01 18

Raw files are written directly to the response writer.


sanderpick:
Doh, looks like I'll have to move the `gateway` package out out from under the `api` package.

sanderpick:
Jeesh, that was a little time consuming. The solution was to move the gateway management up to core, and use a dedicated `client_test` package, since `gateway` now needs it's own API `client`.

I also snuck in some fixes for bucket reading and writing with progress.

sanderpick:
(And removing an entire bucket)

andrewxhill:
suuuuper! could be handy to provide the static hashes (or raw gateway links) in the cloud explorer for each node in the bucket

PR #92, opened 2020-02-11
ci: adds release step for cli

andrewxhill:


andrewxhill:
cool, just fixed an error in the install script i found working with threads. will merge after pass

Issue #65, opened 2020-01-23
Refactor core components


andrewxhill:
I was about to fix this notice about scope switching real quick (https://github.com/textileio/textile/issues/42) but wondered if we could clean up the layers at the beginning.

Summary of current flow is:

- user login
- creates a new 'user' namespace locally linked to their email
- user can then create teams
- user can switch between teams by `textile teams switch ...`
- user can switch back to personal account by `textile switch...`

A simplification here that could solve some issues I had when I first started playing with the current layouit,

- user login
- creates a new 'user' but immediately creates a team for that user. that could be 'personal' or drawn from a random list of words, it doesn't matter so much
- user than can create new teams
- only one switch, `textile teams switch ...` nothing really operates for now at the root user layer (until we have some features that require it at least)

Some things this solves

- only onboard people with one concept to switch around on, a team
- when i onboarded and started creating projects, really quickly i wanted to invite people to those projects, but realizing that i have to go back and create a team first, then projects (first removing old project), then invite.
- if every project has a team by default we can do another enhancement

additional enhancement:

- when a user cc's into a folder where a Project is already initialized, it could immediately switch the user to working within that team. team id could just be added to the yaml to make this quick enough. perhaps there are ways to provide queues in the terminal about the team/project name at all times, e.g. `~/Textile/textile(andrew/minor-fixes)$` my git setup always showing me the current branch in git.

sanderpick:
> Some things this solves 
>     * only onboard people with one concept to switch around on, a team

We could just get rid of the `team switch` command (the root one covers it anyway) if this is confusing. 

>     * when i onboarded and started creating projects, really quickly i wanted to invite people to those projects, but realizing that i have to go back and create a team first, then projects (first removing old project), then invite.

This would require a pretty big refactor. If adding people to projects like this is a needed feature, I'd propose just severing the notion that a project is owned by a team or person... then you could add people or teams to projects. Again, a good amount of work, but sounds like a reasonable change. 

>     * when a user cc's into a folder where a Project is already initialized, it could immediately switch the user to working within that team. team id could just be added to the yaml to make this quick enough. perhaps there are ways to provide queues in the terminal about the team/project name at all times, e.g. `~/Textile/textile(andrew/minor-fixes)$` my git setup always showing me the current branch in git.

The current scope is held on the session, as in, on the server. So, `cd` alone won't be able to alter the session.

andrewxhill:
[...]

PR #93, opened 2020-02-12
Collections Cleanup

### LINKED ISSUE #71: Add indexes to collections
Now that we have index support, we can update the existing collections.

### LINKED ISSUE #88: Add project sub-command, move app tokens to project tokens

### LINKED ISSUE #91: Global project names
A project name makes a much nicer URL slug than a project UUID. Now that we have unique indexes, this will be pretty straightforward.

sanderpick:
Closes #71 
Closes #91 
Closes #88 

Also,
- Cleans up the project config, removing scope and store (precursor to a larger refactor #65)
- `Users` -> `Developers`
- `AppUsers` -> `Users`
- `AppTokens` -> `Tokens`


collections/users.go
```diff
@@ -31,21 +33,46 @@ func (u *Users) GetInstance() interface{} {
...
-	user := &User{
-		Email:   email,
-		Teams:   []string{},
-		Created: time.Now().Unix(),
+	user, err = u.GetByDeviceID(ctx, deviceID)
+	if user != nil {
+		return
 	}
-	if err := u.threads.ModelCreate(ctx, u.storeID.String(), u.GetName(), user); err != nil {
+	if err != nil {
```
jsign: Shouldn't the `err` be checked before `user`?

sanderpick: Ah yeah, thanks! Previously, this was a get by ID, which errors if not found... but we didn't want to stop at that error. Fixed.

PR #95, opened 2020-02-13
Bit more project cleanup

sanderpick:
This migrates project handling in the API to use names, and some other cleanups. I pulled this out of the DNS stuff so we can deploy it on cloud.textile.io, which will help me finish that work.


gateway/gateway.go
```diff
@@ -109,8 +109,8 @@ func (g *Gateway) Start() {
 	router.GET("/confirm/:secret", g.confirmEmail)
 	router.GET("/consent/:invite", g.consentInvite)
 
-	router.GET("/p/:project", g.bucketHandler)
-	router.GET("/p/:project/*path", g.bucketHandler)
+	router.GET("/dashboard/:project", g.dashHandler)
+	router.GET("/dashboard/:project/*path", g.dashHandler)
```
andrewxhill: nice

PR #98, opened 2020-02-13
cli: adds project pull command to grab a team project config

andrewxhill:
I'm guessing this command goes away or changes over time. But for the sake of today's demo it will be really helpful. 

You can accept a team invite, then `projects ls`. if you see a project that you just want to start sharing buckets, you can just pull the config and start using,

```sh
andrewhill@~/Textile/temp/test2$ textile projects ls
NAME     	ID                                   	STORE                                 
heeeeeyo	02ac4a54-257e-4c36-8539-681de0b9849c	d5c0bd11-0fd1-415c-98bc-fa2336965707	
heeeeyo 	8bc982dc-6b68-4eac-9dfb-6c37631e5b82	4dcfbfb1-3a77-45ad-82fc-66a552c19b91	
heeeey  	ad5fcabe-1740-497e-98c2-941de1fb9e1b	52c8d9eb-4d38-4199-90e4-4f89045b10f0	

> Found 3 projects for current scope
andrewhill@~/Textile/temp/test2$ textile project pull heeeeeyo --path heeeeeyo
> Pulled heeeeeyo project to heeeeeyo/.textile
andrewhill@~/Textile/temp/test2$ cd heeeeeyo/
andrewhill@~/Textile/temp/test2/heeeeeyo$ textile buckets ls
NAME    	SIZE 	DIR  	ITEMS 	PATH                                                  
recipes	0   	true	2    	/ipfs/QmSDSmX3Q5QBynw1wBMExmw8A1aFJGRiE4si7AUjjv197G	

> Found 1 items
```

### DIFF
--- a/cmd/textile/projects.go
+++ b/cmd/textile/projects.go
@@ -18,9 +18,14 @@ import (
 func init() {
 	rootCmd.AddCommand(projectsCmd)
 	rootCmd.AddCommand(initProjectCmd) // alias to root
-	projectsCmd.AddCommand(initProjectCmd, lsProjectsCmd, inspectProjectCmd, rmProjectCmd)
+	projectsCmd.AddCommand(initProjectCmd, pullProjectCmd, lsProjectsCmd, inspectProjectCmd, rmProjectCmd)
 
 	initProjectCmd.Flags().String(
+		"path",
+		".",
+		"Project path")
+
+	pullProjectCmd.Flags().String(
 		"path",
 		".",
 		"Project path")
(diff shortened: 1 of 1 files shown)


sanderpick:
This isn't really pulling the project though, is it? There's no server call. The user may think they can pull a project they don't have access to. 

I would just rely on the existing command, `textile init [name]`, which will try to create the project on the server. If it exists and you have access, the server can just return it to you as normal.

andrewxhill:
well... i had thought `textile project pull` should be the whole shebang. you should put in your directory the project config and pull all buckets. but realized I don't have time.

so what this PR does is just creates the project config for you locally. project init wont work because it will try to create a project on the server, who's name is already taken (i think that's what i saw).

an alternative would be a `textile init <name> --remote` or similar to indicate not to register something new

PR #101, opened 2020-02-13
Bucket CNAMES and Pages

sanderpick:
- Creates a CNAME for each (global) bucket
- The gateway leverages the client to implement a simple _bucket file system_ for rendering static assets via a static middleware

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -139,7 +139,7 @@ func (c *Client) ListProjects(ctx context.Context, auth Auth) (*pb.ListProjectsR
 	return c.c.ListProjects(authCtx(ctx, auth), &pb.ListProjectsRequest{})
 }
 
-// RemoveProject removes a project by ID.
+// RemoveProject removes a project by name.
 func (c *Client) RemoveProject(ctx context.Context, name string, auth Auth) error {
 	_, err := c.c.RemoveProject(authCtx(ctx, auth), &pb.RemoveProjectRequest{
 		Name: name,
@@ -275,7 +275,9 @@ func (c *Client) PushBucketPath(
 				Payload: &pb.PushBucketPathRequest_Chunk{
 					Chunk: buf[:n],
 				},
-			}); err != nil {
+			}); err == io.EOF {
+				break
+			} else if err != nil {
 				_ = stream.CloseSend()
 				return nil, nil, err
 			}
--- a/api/service.go
+++ b/api/service.go
@@ -697,9 +697,9 @@ func (s *service) getBucketWithScope(ctx context.Context, name, scope string) (*
 	if buck == nil {
 		return nil, status.Error(codes.NotFound, "Bucket not found")
 	}
-	if _, err := s.getProjectForScope(ctx, buck.ProjectID, scope); err != nil {
-		return nil, err
-	}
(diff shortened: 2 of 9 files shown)


api/client/client.go
```diff
@@ -275,7 +275,9 @@ func (c *Client) PushBucketPath(
 				Payload: &pb.PushBucketPathRequest_Chunk{
 					Chunk: buf[:n],
 				},
-			}); err != nil {
+			}); err == io.EOF {
+				break
```
jsign: :sweat_smile:

api/service.go
```diff
@@ -908,6 +914,15 @@ func (s *service) createBucket(ctx context.Context, proj *c.Project, name string
 		return nil, err
 	}
 
+	if s.dnsManager != nil {
+		parts := strings.SplitN(s.gatewayUrl, "//", 2)
+		if len(parts) > 1 {
+			if _, err := s.dnsManager.NewCNAME(name, parts[1]); err != nil {
+				return nil, err
+			}
+		}
```
jsign: The `else` case seems like a malformed url, so preferred to ignore?

gateway/bucketfs.go
```diff
@@ -0,0 +1,85 @@
...
+// serveBucket returns a middleware handler that serves files in a bucket.
+func serveBucket(fs serveBucketFileSystem) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		bucket, err := bucketFromHost(c.Request.Host, fs.ValidHost())
+		if err != nil {
+			return
+		}
+		if fs.Exists(bucket, c.Request.URL.Path) {
+			c.Writer.WriteHeader(http.StatusOK)
+			ctype := mime.TypeByExtension(filepath.Ext(c.Request.URL.Path))
```
sanderpick: Would be nice to use more advanced content type detection, but we'd need the client to implement a `ReadSeeker`.

gateway/bucketfs.go
```diff
@@ -0,0 +1,85 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), f.timeout)
+	defer cancel()
+	return f.client.PullBucketPath(ctx, path.Join(bucket, pth), writer, f.auth)
+}
+
+func (f *bucketFileSystem) ValidHost() string {
+	return f.host
+}
+
+func bucketFromHost(host, valid string) (buck string, err error) {
```
sanderpick: This prevents bucket rendering under a different domain.

api/client/client.go
sanderpick: Oh yeah haha. I can remove that when digging into the nginx-related. steam error I mentioned in Slack.

api/service.go
sanderpick: As in, ignore the request?

jsign: I mean, if `s.gatewayUrl` parsing doesn't have `len(parts)>1`, the current code ignores that and doesn't register the CNAME. Is that something that should happen? I wonder if a malformed `s.gatewayUrl` can backfire somewhere else (or in the future).

sanderpick: Oh, I see. Makes sense. We could try to parse the URL on start, that fails -> fatal?

jsign: It may be an option, yeah. I'll leave if that may worth it to you :P

PR #104, opened 2020-02-14
Better bucket exists error

sanderpick:
Not much to see here, just showing a better user error when pushing to a bucket that already exists for a different scope.

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -892,28 +892,33 @@ func TestClient_RemoveBucketPath(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	file, err := os.Open("testdata/file1.jpg")
+	file1, err := os.Open("testdata/file1.jpg")
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer file.Close()
+	defer file1.Close()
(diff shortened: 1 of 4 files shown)


api/client/client_test.go
```diff
@@ -892,28 +892,33 @@ func TestClient_RemoveBucketPath(t *testing.T) {
...
 		t.Fatal(err)
 	}
-	file, err := os.Open("testdata/file1.jpg")
+	file1, err := os.Open("testdata/file1.jpg")
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer file.Close()
+	defer file1.Close()
+	file2, err := os.Open("testdata/file2.jpg")
```
sanderpick: Just making this remove bucket path test a little more comprehensive.

gateway/gateway.go
```diff
@@ -215,6 +215,8 @@ func (g *Gateway) bucketHandler(c *gin.Context) {
 
 	for _, item := range rep.Item.Items {
 		if item.Name == "index.html" {
+			c.Writer.WriteHeader(http.StatusOK)
+			c.Writer.Header().Set("Content-Type", "text/html")
```
sanderpick: May as well...

PR #113, opened 2020-02-27
Generates account admin lib from protobufs

### LINKED ISSUE #110: Generate JS client from API

andrewxhill:
not sure how much of this will change due to upcoming system changes, so happy to hold off merge but wanted to capture this. just let me know @sanderpick 

fixes #110


sanderpick:
Looks good, but can we only change the package name, not the filenames? We use the api patterns in a few places to indicate gRPC APIs.

andrewxhill:
ah yep, name change not required

sanderpick:
Hrm, I'm not sure "account" is the right name here... it does more than account stuff, which is why it was called Textile API

andrewxhill:
There is a bit of a conflict in purely a Textile lib though. 

1. There are is js-textile, which can cover most use cases for use auth tokens etc. Seems to me like the more nature `textile`
2. This thing, which is something for managed internal apis linked to developer accounts. Probably only used by our own code. 

what do you think is best for calling it?

sanderpick:
Isn't `js-textile` actually threads though?

api/client/client.go
```diff
@@ -382,11 +382,11 @@ func (t tokenAuth) GetRequestMetadata(ctx context.Context, _ ...string) (map[str
 	md := map[string]string{}
 	token, ok := ctx.Value(authKey("token")).(string)
 	if ok && token != "" {
-		md["Authorization"] = "Bearer " + token
+		md["authorization"] = "bearer " + token
```
sanderpick: Why's this?

api/pb/Makefile
```diff
@@ -1,16 +1,19 @@
 PB = $(wildcard *.proto)
 GO = $(PB:.proto=.pb.go)
+PROTOC_GEN_TS_PATH = "./javascript/node_modules/.bin/protoc-gen-ts"
 
 all: $(GO)
 
 %.pb.go: %.proto
-	protoc -I=. -I=$(GOPATH)/src \
+	protoc -I=. \
+	--plugin="protoc-gen-ts=${PROTOC_GEN_TS_PATH}" --ts_out="service=grpc-web:javascript/src/." \
```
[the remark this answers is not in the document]
sanderpick: Why does `-I=$(GOPATH)/src` get removed?

api/pb/javascript/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@textile/account",
+  "version": "0.0.1",
+  "description": "A client for interacting Textile account APIs.",
```
sanderpick: Typo

api/client/client_test.go
```diff
@@ -1012,7 +1012,7 @@ func makeTextile(t *testing.T) (conf core.Config, shutdown func()) {
 		RepoPath: dir,
 
 		AddrApi: util.MustParseAddr(fmt.Sprintf("/ip4/127.0.0.1/tcp/%d", apiPort)),
-
```
sanderpick: Can you maintain the white space that was here? Makes it easier to read.

sanderpick: As in, things that are alike are grouped.

api/server.go
```diff
@@ -112,11 +143,21 @@ func NewServer(ctx context.Context, conf Config) (*Server, error) {
 		}
 	}()
 
+	go func() {
+		s.rpcWebProxy.ListenAndServe()
```
sanderpick: You need to handle errors here... see the threads API: https://github.com/textileio/go-threads/blob/master/api/server.go#L84 to line 124

api/server.go
```diff
@@ -154,8 +196,8 @@ func (s *Server) authFunc(ctx context.Context) (context.Context, error) {
 	if err != nil {
 		return nil, status.Error(codes.PermissionDenied, "User not found")
 	}
-
-	scope := metautils.ExtractIncoming(ctx).Get("X-Scope")
+	// @TODO check if this messes up other request types
+	scope := metautils.ExtractIncoming(ctx).Get("x-scope")
```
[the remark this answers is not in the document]
sanderpick: Okay cool.

andrewxhill:
while building it, I had imagined `js-textile` was going to become everything related to using your project tokens. so remote threads, thread followers, remote pins/buckets, etc from an app a dev is building. so maybe this one brings the stuff from js-textile into it and we get rid of that separate repo? 

otherwise i had imagined this pr was for generally internal apis that we wouldn't document for developers.

sanderpick:
Nice, sounds good. Since `js-textile` has some non-grpc generated stuff, I think it makes sense to keep it on its own. 

We could call this api "cloud"? So, in JS it would be a package called `@textile/cloud`?

andrewxhill:
cloud, nice! updates coming

api/client/client.go
andrewxhill: same thing as the x-scope. the grpc client lib in java and javascript lowercase these no matter my input, so I wanted to ensure we matched here. can change it back if you prefer, just felt like the grpc preference for lowercase

api/pb/api.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 package pb;
 
 option java_multiple_files = true;
-option java_package = "io.textile.textile_grpc";
+option java_package = "io.textile.account_grpc";
```
sanderpick: Cloud?

api/pb/Makefile
```diff
@@ -1,16 +1,19 @@
...
 	protoc -I=. -I=$(GOPATH)/src \
+	--plugin="protoc-gen-ts=${PROTOC_GEN_TS_PATH}" --ts_out="service=grpc-web:javascript/src/." \
 	--go_out=\
 	plugins=grpc:\
 	. $<
 
 clean:
 	rm -f *.pb.go
-	rm -f *pb_test.go
+	rm -f ./javascript/src/api_pb*
```
sanderpick: This one needed?

api/server.go
```diff
@@ -25,6 +27,7 @@ import (
 var (
 	log = logging.Logger("textileapi")
 
+	// @TODO recompile pb without the 'api.' prefix, fix this
```
sanderpick: Still needed?

api/server.go
```diff
@@ -112,11 +143,23 @@ func NewServer(ctx context.Context, conf Config) (*Server, error) {
 		}
 	}()
 
+	errc := make(chan error)
+	go func() {
+		errc <- s.rpcWebProxy.ListenAndServe()
```
sanderpick: It doesn't look like you're reading the error channel. In any case, you can just do a normal error check... like @jsign mentioned on the fil-tools PR. You'll just have to check for and ignore the server closed error.

api/pb/Makefile
andrewxhill: just tested, i guess not necessary no

Issue #111, opened 2020-02-22
Rendering of website sub paths not always working


andrewxhill:
Here is an example,

https://gatsby-ipfs-blog-377515556.textile.cafe

That site renders fine. If you go to the link of the first blog post on the webpage, it will also render fine. If you now copy that URL (https://gatsby-ipfs-blog-377515556.textile.cafe/textile-buckets) and paste it in a new browser tab to render it directly, it will 404

PR #115, opened 2020-03-03
Allows direct rendering of subdirectories in bucket gateway

andrewxhill:
fixes https://github.com/textileio/textile/issues/111

### DIFF
--- a/gateway/bucketfs.go
+++ b/gateway/bucketfs.go
@@ -16,7 +16,7 @@ import (
 )
 
 type serveBucketFileSystem interface {
-	Exists(bucket, pth string) bool
+	Exists(bucket, pth string) (bool, string)
 	Write(bucket, pth string, writer io.Writer) error
 	ValidHost() string
 }
@@ -35,7 +35,9 @@ func serveBucket(fs serveBucketFileSystem) gin.HandlerFunc {
 		if err != nil {
 			return
 		}
-		if fs.Exists(bucket, c.Request.URL.Path) {
+
+		exists, target := fs.Exists(bucket, c.Request.URL.Path)
+		if exists {
 			c.Writer.WriteHeader(http.StatusOK)
 			ctype := mime.TypeByExtension(filepath.Ext(c.Request.URL.Path))
 			if ctype == "" {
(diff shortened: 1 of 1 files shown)


gateway/bucketfs.go
```diff
@@ -47,22 +49,38 @@ func serveBucket(fs serveBucketFileSystem) gin.HandlerFunc {
...
 	defer cancel()
 	rep, err := f.client.ListBucketPath(ctx, "", path.Join(bucket, pth), f.auth)
 	if err == nil && !rep.Item.IsDir {
-		return true
+		return true, ""
+	}
+	if rep.Item.IsDir {
+		for _, item := range rep.Item.Items {
+			if item.Name == "index.html" {
+				return false, item.Name
```
andrewxhill: this could return `(bool, bool)` but i left it open here intentionally to support different default render paths. at least `.htm` and `.html` came to mind

gateway/bucketfs.go
```diff
@@ -47,22 +49,38 @@ func serveBucket(fs serveBucketFileSystem) gin.HandlerFunc {
 			} else {
 				c.Abort()
 			}
+		} else if target != "" {
+			c.Writer.WriteHeader(http.StatusOK)
+			c.Writer.Header().Set("Content-Type", "text/html")
```
andrewxhill: just renders a directory's default file if it exists. target should equal `index.html` for now

sanderpick: would, `ctype := mime.TypeByExtension(filepath.Ext(c.Request.URL.Path))` not work here?

gateway/bucketfs.go
sanderpick: To clarify, the idea here is to allow rendering a website in a bucket directory, yes? Are relative URLs broken when you do this?

gateway/bucketfs.go
andrewxhill: yeah maybe, it would actually be `mime.TypeByExtension(filepath.Ext(fmt.Printf("%s/%s", c.Request.URL.Path, target)))` or similar though

gateway/bucketfs.go
andrewxhill: i tested by recreating this bucket's data locally: https://gatsby-ipfs-blog.textile.cafe/

navigating links from the homepage worked fine, all embedded assets worked fine (imgs, js, etc)

also, direct links to subdirectories (the issue on master) was fixed. so linking to https://gatsby-ipfs-blog.textile.cafe/new-beginnings/ works on this branch (not on production)

gateway/bucketfs.go
sanderpick: Nice, should we do that? Just to max out on flex?

gateway/bucketfs.go
sanderpick: Ahh, I see... I had forgotten the core issue. Sweet.

gateway/bucketfs.go
andrewxhill: yep! will do

andrewxhill: k, done

PR #116, opened 2020-03-10
buckets: fix err handling and dir rendering

andrewxhill:


gateway/bucketfs.go
```diff
@@ -72,8 +72,9 @@ func (f *bucketFileSystem) Exists(bucket, pth string) (bool, string) {
 	ctx, cancel := context.WithTimeout(context.Background(), f.timeout)
 	defer cancel()
 	rep, err := f.client.ListBucketPath(ctx, "", path.Join(bucket, pth), f.auth)
-	if err == nil && !rep.Item.IsDir {
-		return true, ""
+	if err != nil {
+		log.Errorf("error listing bucket path: %v", err)
```
andrewxhill: not confident that this log is actually useful... thoughts?

sanderpick: Why's that? It's good practice to log an error if you're not returning it.

sanderpick: That is to say, looks like a nice addition.

andrewxhill: cool, all good then

PR #139, opened 2020-04-23
BucketFS: Updates for new buckets

sanderpick:
This updates the gateway and bucketfs logic to handle the new buckets setup.


cmd/textiled/main.go
```diff
@@ -53,10 +53,6 @@ var (
 			Key:      "addr.gateway.url",
 			DefValue: "http://127.0.0.1:8006",
 		},
-		"addrGatewayBucketDomain": {
```
sanderpick, line 56: I can't remember why I added this... the DNS domain setting can be used instead.

PR #144, opened 2020-04-25
Bucket key and API updates

sanderpick:
- Migrates a bucket instance ID to an IPNS compatible key.
- Removes the unique constraint on bucket names and the auto-generated slug from the name, which is no longer needed in a bucket URL.
- Adds some needed improvements to the buckets API:
   - Adds an `Init` method, which returns a new bucket root. Previously, this init process was done as needed (magically) on the first push, but this gets pretty sloppy when bucket slugs go away. The API should not rely on free from text as the main identifier. Now, you init a bucket and get a key that is used with all other requests, which feels a lot cleaner.
- Since the bucket path magic is now gone, some new APIs were needed:
  - `List` all buckets (in the current context / thread)
  - `Remove` is the opposite of `Init`, which was previously handled by a magical heuristic: "If the bucket is empty after a `RevmovePath` call, then remove the entire bucket", which doesn't feel right along with the more explicit `Init` the CWD flow.


sanderpick:
Merging this now so I can get it up on staging later.

PR #146, opened 2020-04-26
Rename js packages to match existing conventions

carsonfarmer:
Tiny 'nit' PR. If there is an argument against this, that's also fine. We use `*-grpc` in our other JS libs, it it seemed like a good idea to do the same here. Also, seems useful to reserve the top-level lib names (`hub`, `user`, `buckets`) for those actual clients (if we ever needed them), plus this avoids confusion when folks are installing via npm (they might try to install @textile/hub and get something unexpected).

### DIFF
--- a/api/buckets/pb/javascript/package.json
+++ b/api/buckets/pb/javascript/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@textile/buckets-grpc",
   "version": "0.0.0",
-  "description": "Interfaces/classes for interacting with the Textile Buckets gRPC API.",
+  "description": "A client for interacting with the Textile Buckets gRPC API.",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/textileio/textile.git"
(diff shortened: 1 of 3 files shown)


api/buckets/pb/javascript/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@textile/buckets",
+  "name": "@textile/buckets-grpc",
   "version": "0.0.0",
-  "description": "A client for interacting with the Textile Buckets gRPC API.",
+  "description": "Interfaces/classes for interacting with the Textile Buckets gRPC API.",
```
sanderpick, line 4: It's a client for interacting with the gRPC API, no?

carsonfarmer: Is it? These don't include the client libs, just the proto classes etc right? Or maybe I have an overly strict definition of client?

carsonfarmer: I would consider the thing in @textile/textile the _client_.

PR #148, opened 2020-04-27
run dev locally in single commands

jsign:
Created a `dev` compose file that doesn't rely on compiled `textile`, but local code. So you can change the code and run things again. (No .env things needed). This is a similar idea of the `make embed` in Powergate, which re-runs the docker-build on changed code.

The run is clean in the sense that *nothing* about the docker-compose run is mapped to the local HDD, all lives inside docker volumes.

I added new `make` commands:
- `make local-up`: spin up things with `textile` running local code.
- `make local-stop`: to stop things, but keep volumes (state).
- `make local-clean`: clean all things, so next `local-up` is a fresh world. We could go a step further and also do `rm -r ~/.textile` to rm any client folder... but preferred to leave the host folder untouched just in case the client wasn't being run with defaults.

My actual use:
`make local-up`
`tt --api "localhost:3006: init`
`tt buckets init`
...
eventually if want to start fresh: `make local-clean local-up`

Also, the email-session-secret is fixed so `init`ing can be done fast using `textilesession` as the email session for validating email.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,11 @@
 textile:
 	go install ./...
+
+local-up:
+	docker-compose -f docker-compose-dev.yml up
+
+local-stop:
+	docker-compose stop
+
+local-clean:
+	docker-compose down -v
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,0 +1,27 @@
+version: "3"
+services:
+  textile:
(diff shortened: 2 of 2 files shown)


asutula:
Been using this, but it seems that my server side grpc service changes aren't being picked up when I clean an up again

PR #150, opened 2020-04-28
enable race detector & fix data races

jsign:
- Enable race detector
- Fix data race in tests
- Fix data race in buckest `PushPath`


api/buckets/service.go
```diff
@@ -388,6 +388,7 @@ func (s *Service) PushPath(server pb.API_PushPathServer) error {
 		return err
 	}
 
+	size := <-chSize
```
jsign, line 391: Tricky one. `size` was being used in `pb.PushPathReply_Event` but was not being waited correctly for the progress goroutine. This resulted in some garbage size value.

sanderpick: Nice, good catch 💪

PR #151, opened 2020-04-28
Powergate integration

jsign:
Starting with the integration, some hidden things appeared which needed fixes: 
- https://github.com/textileio/powergate/pull/363
- https://github.com/textileio/go-threads/pull/324
- https://github.com/textileio/textile/pull/150
- https://github.com/textileio/powergate/pull/366
- https://github.com/textileio/powergate/pull/368
- https://github.com/textileio/powergate/pull/373 
- ~Some other dep manual replacement that will be explained. Nothing too bad.~
- https://github.com/textileio/powergate/pull/389
- https://github.com/textileio/powergate/pull/397
- https://github.com/textileio/powergate/pull/402

Target:
- #160 
- #161
- #162

---

A bit of a big PR, not only regarding code changes in this repo but the listed things above that were needed to move forward. Feels like a good v1 for the integration.

In short, here's the story:
- If the textile server is configured with a Powergate addr, the archive feature will be enabled. If not, will be disabled. So, assuming that is configured correctly.
- When a bucket is created, an FFS instance is created on Powergate.
- The authapi is stored in a Mongo collection, so to be used in further API calls.
- A user can create an archive doing `tt bucket archive`
- This will queue a PushConfig on the Powergate side.
- The user can `tt bucket archive status` to query the current status of the last pushed archive. 
- It can also do `tt bucket archive status -w` to do the same as above, plus seeing realtime logs of what PG is doing while executing the archiving.
- On success, deal data is stored in a collection in the ThreadDB of the underlying thread of the bucket, so that it can be considered public information to retrieve the archive.

---

But this is already covering border cases such as archiving too fast, so having more than one archive queued, or cancelling already queued archives in favor of a new archive, etc. Mostly resolved by PG features in the scheduler.

Most prob, still some rough edges. But preferred to put a breaking point and merge this. This will keep evolving.


.github/workflows/test.yml
```diff
@@ -43,4 +43,16 @@ jobs:
         run: |
           go get -v -t -d ./...
       - name: Test
-        run: go test -race ./...
+        run: go test -race `go list ./... | grep -v integrationtest`
```
jsign, line 46: Separate the usual tests, from the heaviest one that also spin up PG, IPFS (from PG) and devnet.

Both sets of tests are run, but in different jobs.

api/apitest/apitest.go
```diff
@@ -25,6 +25,11 @@ var SessionSecret = NewUsername()
 func MakeTextile(t *testing.T) (conf core.Config, shutdown func()) {
```
jsign, line 25: Some small refactor to provide a custom config for running tests.
Needed in the integration test to tune up configs for it.

api/buckets/buckets.go
```diff
@@ -22,20 +37,56 @@ var (
...
-	Name      string `json:"name"`
-	Path      string `json:"path"`
-	DNSRecord string `json:"dns_record,omitempty"`
-	CreatedAt int64  `json:"created_at"`
-	UpdatedAt int64  `json:"updated_at"`
+	Key       string   `json:"_id"`
+	Name      string   `json:"name"`
+	Path      string   `json:"path"`
+	DNSRecord string   `json:"dns_record,omitempty"`
+	Archives  Archives `json:"archives"`
```
jsign, line 44: New attr.

api/buckets/buckets.go
```diff
@@ -22,20 +37,56 @@ var (
...
+	CreatedAt int64    `json:"created_at"`
+	UpdatedAt int64    `json:"updated_at"`
+}
+
+type Archives struct {
+	Current Archive   `json:"current"`
+	History []Archive `json:"history"`
+}
```
jsign, line 52: Mongo world:
We save the current archive (latest), and also the history. The history isn't used yet but sounds like it's useful.

api/buckets/buckets.go
```diff
@@ -22,20 +37,56 @@ var (
...
+	threads  *dbc.Client
+	pgClient *powc.Client
+
+	lock   sync.Mutex
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
+
+func New(t *dbc.Client, pgc *powc.Client, col *collections.FFSInstances) *Buckets {
```
jsign, line 80: Preferred to create a constructor here to control a bit how a `Buckets` should be constructed since got a bit more complex.

jsign: Took advantage of that and made attributes unexported to be a bit safer from external changes.

api/buckets/buckets.go
```diff
@@ -62,16 +114,40 @@ func (b *Buckets) Create(ctx context.Context, dbID thread.ID, key, name string,
 		return nil, err
 	}
 	bucket.Key = ids[0]
+
+	if err := createFFSInstance(ctx, b.ffsCol, b.pgClient, key); err != nil {
```
jsign, line 118: Create FFS instance on bucket creation.

api/buckets/buckets.go
```diff
@@ -106,23 +182,258 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+	b.lock.Lock()
+	defer b.lock.Unlock()
+	ffsi, err := b.ffsCol.Get(ctx, key)
+	if err != nil {
+		return fmt.Errorf("getting ffs instance data: %s", err)
+	}
+
+	ctxFFS := context.WithValue(ctx, powc.AuthKey, ffsi.FFSToken)
+	var jid ffs.JobID
+	if ffsi.Archives.Current.JobID != "" {
```
jsign, line 209: We make here the distinction to call `PushConfig` if it's the first archive, or `Replace` if its a further one.

api/buckets/buckets.go
```diff
@@ -106,23 +182,258 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+	ffsi.Archives.Current = collections.Archive{
+		Cid:       c.String(),
+		CreatedAt: time.Now().Unix(),
+		JobID:     jid.String(),
+		JobStatus: int(ffs.Queued),
+	}
+	if err := b.ffsCol.Replace(ctx, ffsi); err != nil {
+		return fmt.Errorf("updating ffs instance data: %s", err)
+	}
+	go b.trackArchiveProgress(util.NewClonedContext(ctx), dbID, key, jid, ffsi.FFSToken, c, opts...)
```
jsign, line 238: `Archive` call finish fast.
It delegates to a go-routine to track the progress and keep updating in the Mongo state the status of the Job.

api/buckets/buckets.go
```diff
@@ -106,23 +182,258 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+
+	return nil
+}
+
+// ArchiveStatus returns the last known archive status on Powergate. If the return status is Failed,
+// an extra string with the error message is provided.
+func (b *Buckets) ArchiveStatus(ctx context.Context, key string) (ffs.JobStatus, string, error) {
```
jsign, line 245: We only query the Mongo database.
The background go-routine is updating the Job status... so if clients get crazy querying status, it will only hit Mongo, not PG.

api/buckets/buckets.go
```diff
@@ -106,23 +182,258 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+	ffsi, err := b.ffsCol.Get(ctx, key)
+	if err != nil {
+		return ffs.Failed, "", fmt.Errorf("getting ffs instance data: %s", err)
+	}
+
+	if ffsi.Archives.Current.JobID == "" {
+		return ffs.Failed, "", ErrNoCurrentArchive
+	}
+	current := ffsi.Archives.Current
+	if current.Aborted {
```
jsign, line 255: The `Aborted` indicates a situation in that the background goroutine ended unexpectedly and the Job status might be in an indefinite state. This should be re-handled eventually, or similar.
The summary is: if it's aborted, we're not sure how the Job ended. Doesn't necessarily means it failed, just that maybe the network conn exploded and we lost track of the status.

It's important eventually to add further code to recover these cases, etc.

api/buckets/buckets.go
```diff
@@ -106,23 +182,258 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+		Deals: deals,
+	}
+
+	if err = b.threads.Save(ctx, dbID, CollectionName, dbc.Instances{buck}, db.WithTxnToken(args.Token)); err != nil {
+		return fmt.Errorf("saving deals in thread: %s", err)
+	}
+	return nil
+}
+
+// updateArchiveStatus save the last known job status. It also receives an
```
jsign, line 395: Similar to what was said above. Prefered to keep that documented.

cmd/textiled/main.go
```diff
@@ -253,9 +253,9 @@ var rootCmd = &cobra.Command{
 		addrGatewayHost := cmd.AddrFromStr(configViper.GetString("addr.gateway.host"))
 		addrGatewayUrl := configViper.GetString("addr.gateway.url")
 
-		var addrFilecoinApi ma.Multiaddr
-		if str := configViper.GetString("addr.filecoin.api"); str != "" {
-			addrFilecoinApi = cmd.AddrFromStr(str)
+		var addrPowergateApi ma.Multiaddr
+		if str := configViper.GetString("addr.powergate.api"); str != "" {
+			addrPowergateApi = cmd.AddrFromStr(str)
```
jsign, line 258: This config is optional. If not set, archive feature isn't available... so you're not forced to have a PG instance running.

cmd/tt/bucket.go
```diff
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
```
jsign, line 5: Not a ninja on CLI things, so some reasonable setup :P

collections/ffs.go
```diff
@@ -0,0 +1,70 @@
+package collections
+
+import (
+	"context"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+type FFSInstance struct {
```
jsign, line 10: So now we have a new Mongo collection: FFSInstance.
It contains all the information of a FFS instance which corresponds to a bucket.

Information here isn't available for the user (of course, it's in Mongo)... mostly, this is for tracking JobIDs, current status, and "ephemeral/temporal" things that we need to archiving.

Plus, of course, the auth token.

integrationtest/pg/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
+version: '3.7'
```
jsign, line 1: Whole stack for integration test.

integrationtest/pg/pg_test.go
```diff
@@ -0,0 +1,264 @@
+package pg
```
jsign, line 1: Integration test to cover most things.

integrationtest/pg/pg_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+	_, err = client.Init(ctx, "bucky")
+	require.Nil(t, err)
+
+	lst, err = powc.FFS.ListAPI(ctx)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(lst))
+}
+
+func TestArchiveBucketWorkflow(t *testing.T) {
```
jsign, line 46: Might be good to look at this test.
Kind of covering the usual workflow.

integrationtest/pg/powcli/docker-compose-pow.yml
```diff
@@ -0,0 +1,56 @@
+version: '3.7'
```
jsign, line 1: Similarish thing as used in integration testing.
But allows you to spin up the whole stack, and play with the CLI.
Basically that magic `make` that I did in the demo video, and then used the CLI locally.

util/ctxutil.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import (
+	"context"
+	"time"
+)
+
+// NewClonedContext returns a context with the same Values
```
jsign, line 8: Allows to clone a context values, without inheriting its cancellation setup.
Handly for spinning goroutines out of the request scope.

integrationtest/pg/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+  
+  powergate:
+    image: textile/powergate:sha-77c885e
+    ports:
+      - 8889:8889
+      - 8888:8888
+      - 6060:6060
+      - 5002:5002
+      - 6002:6002
+    depends_on:
```
sanderpick, line 16: 🤘

util/ctxutil.go
sanderpick: Clever!

integrationtest/pg/pg_test.go
sanderpick: So much going on under the hood, this makes it look too easy 👍

integrationtest/pg/pg_test.go
```diff
@@ -0,0 +1,264 @@
...
+func TestFailingArchive(t *testing.T) {
+	_ = spinup(t)
+	ctx, client := setup(t)
+
+	b, err := client.Init(ctx, "bucky")
+	require.Nil(t, err)
+	time.Sleep(4 * time.Second)
+	// Store a file that is bigger than the sector size, this
+	// should lead to an error on the PG side.
+	addDataFileToBucket(ctx, t, client, b.Root.Key, "Data3.txt")
```
sanderpick, line 143: Have you set the sector size to be small somewhere? Not sure how that all works.

integrationtest/pg/pg_test.go
```diff
@@ -0,0 +1,264 @@
...
+	require.Nil(t, err)
+
+	user := apitest.Signup(t, hubclient, conf, apitest.NewUsername(), apitest.NewEmail())
+	ctx := common.NewSessionContext(context.Background(), user.Session)
+	id := thread.NewIDV1(thread.Raw, 32)
+	ctx = common.NewThreadNameContext(ctx, "buckets")
+	err = threadsclient.NewDB(ctx, id)
+	require.Nil(t, err)
+	ctx = common.NewThreadIDContext(ctx, id)
+	t.Cleanup(func() {
```
sanderpick, line 211: Oh, well that's handy

api/buckets/buckets.go
```diff
@@ -22,20 +37,56 @@ var (
...
+	History []Archive `json:"history"`
+}
+
+type Archive struct {
+	Cid   string `json:"cid"`
+	Deals []Deal `json:"deals"`
+}
+
+type Deal struct {
+	ProposalCid string `json:"proposalCid"`
```
sanderpick, line 60: Nit: Camelcase to match others?

collections/ffs.go
```diff
@@ -0,0 +1,70 @@
...
+	Archives  Archives
+}
+
+type Archives struct {
+	Current Archive
+	History []Archive
+}
+
+type Archive struct {
+	Cid        string
```
sanderpick, line 22: Mongo will bonk if you make this `cid.Cid`?

collections/ffs.go
jsign: Correct.

collections/ffs.go
jsign: Good question, will try. I suppose that will call `MarshalJSON` sooo it shouldn't be a problem.
Will report back here.

core/core.go
```diff
@@ -78,14 +78,14 @@ type Textile struct {
...
-	AddrGatewayURL  string
-	AddrFilecoinAPI ma.Multiaddr
-	AddrMongoURI    string
+	AddrAPI          ma.Multiaddr
+	AddrAPIProxy     ma.Multiaddr
+	AddrThreadsHost  ma.Multiaddr
+	AddrIPFSAPI      ma.Multiaddr
+	AddrGatewayHost  ma.Multiaddr
+	AddrGatewayURL   string
+	AddrPowergateApi ma.Multiaddr
```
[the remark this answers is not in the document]
jsign: Agree, will change.

collections/ffs.go
sanderpick: Maybe the same goes for some of the other types here... not sure if they are representing some higher type like `Cid`

integrationtest/pg/pg_test.go
jsign: The `textileio/lotus-devnet` has a fixed 2KB sector size. I'm not sure that's tunable without breaking some other sectorbuilder-mock implementation assumptions. So you can imagine that being _baked-in_. 
But now that you mention that I could try and see what happen to maybe make that tunable if magically works.

integrationtest/pg/pg_test.go
jsign: Yep!

collections/ffs.go
jsign: Oh, doesn't work since Mongo allows to provide serialization of BSON:
```
// Marshaler is an interface implemented by types that can marshal themselves
// into a BSON document represented as bytes. The bytes returned must be a valid
// BSON document if the error is nil.
type Marshaler interface {
	MarshalBSON() ([]byte, error)
}
```
So that would need some augmenting type.

But, took advantage and switched to store bytes... sounds that would be better (even in BSON) than storing a base32 string cid.

api/buckets/buckets.go
jsign: You mean:
```
ProposalCID string `json:"proposalCid"`
```
Or:
```
ProposalCid string `json:"proposal_cid"`
```
Or some other combination? :P

sanderpick: Oop, the second one!

jsign: Sure, will change.

collections/ffs.go
sanderpick: Sounds good. For the other collections, I opted for the much more painful approach of handling the encoding / decoding internally so that the consumer doesn't have to do something extra before using the values. I think that's pretty nice, but I'm not sure if it's worth it in this case.

sanderpick: Yep, for thread IDs etc, I've just been storing bytes, and then decoding into the right type on the way out.

jsign: Well, to make that happen the API can use a struct with `cid.Cid`, and then save internally another struct with `[]bytes`. That would only mean moving that custom `.Bytes()` internally, and still use the automatic encoding/decoding which might be a bit less painful (would be kind of a direct mapping of fields). Mentioning that as another middle-point option.

sanderpick: That sounds like a good option. Pretty close to the other collections, but they're using the `bson.M` instead of an additional struct. That struct can carry `bson` tags with JSON looking tags (snakecase), which will make playing with data in the mongo CLI a bit more intuitive since all key patterns would match.

jsign:
> The way you have embedded the archive info _into_ the bucket JSON is very nice for encapsulating _all bucket info / history_, but it would be nice to be able to directly address an archive via the thread ID pattern: `/<collection_name>/<instance_id>`. For that to work, an archive would have to be in it's own collection. Now that I think about it though, this would be a lot of extra stuff for a simple feature.
> 
> Another solution: Make the gateway smarter such that it can traverse instance JSON via query params like, `?archives.current`

Oh, I forgot to mention this.
That's a good observation that I hadn't in mind while making the decision. Looks like we can move to its own collection, or make that custom-walker which might be useful for more things. A good thing to think about.

PR #154, opened 2020-04-29
chore(docs): update doc strings

andrewxhill:
fixes https://github.com/textileio/textile/issues/152

### DIFF
--- a/cmd/tt/buckets.go
+++ b/cmd/tt/buckets.go
@@ -35,20 +35,10 @@ func init() {
 	}
 }
 
-<<<<<<< HEAD
 var bucketCmd = &cobra.Command{
 	Use:   "bucket",
 	Short: "Manage a bucket",
 	Long:  `Init a bucket and push and pull files and folders.`,
-=======
-var bucketsCmd = &cobra.Command{
-	Use: "bucket",
-	Aliases: []string{
-		"buckets",
-	},
-	Short: "Bucket management",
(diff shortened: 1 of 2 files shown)


cmd/tt/buckets.go
```diff
@@ -36,12 +36,12 @@ func init() {
 }
 
 var bucketsCmd = &cobra.Command{
-	Use: "buckets",
+	Use: "bucket",
```
andrewxhill, line 39: i read this suggestion from @jsign

cmd/tt/buckets.go
```diff
@@ -203,6 +203,11 @@ File structure is mirrored in the bucket. For example, given the directory:
 
 These 'push' commands result in the following bucket structures.
 
+'tt buckets push foo/ mybuck':
+    mybuck/one.txt
+    mybuck/bar/two.txt
+		mybuck/bar/baz/three.txt
+
```
andrewxhill, line 210: moved this up to the top of the list since the first thing i typically do in a new bucket is push some dir to the root of the bucket

jsign: I think the `mybuck/` prefixes in these results should be deleted. Or maybe missing something?

sanderpick: Yep, +1

andrewxhill: right, cool, will fix

PR #158, opened 2020-04-29
IPNS: use key cids for bucket instance id, adds ipns manager

sanderpick:
- Uses base32 IPNS key CIDs as bucket instance IDs
- Adds IPNS publishing on init, pushpath, and removepath
- Updates deps


api/buckets/ipnsmanager.go
```diff
@@ -0,0 +1,102 @@
...
+	"errors"
+	"sync"
+
+	iface "github.com/ipfs/interface-go-ipfs-core"
+	opt "github.com/ipfs/interface-go-ipfs-core/options"
+	"github.com/ipfs/interface-go-ipfs-core/path"
+	c "github.com/textileio/textile/collections"
+)
+
+// IPNSManager handles bucket name publishing to IPNS.
```
sanderpick, line 14: This thing publishes to IPNS when the bucket root changes.

api/buckets/ipnsmanager.go
```diff
@@ -0,0 +1,102 @@
...
+// Cancel all pending publishes.
+func (m *IPNSManager) Cancel() {
+	m.lock.Unlock()
+	defer m.lock.Unlock()
+	for _, cancel := range m.ctxs {
+		cancel()
+	}
+}
+
+func (m *IPNSManager) publish(pth path.Path, keyID string) {
```
sanderpick, line 43: Some points about the goals here:
- Only one publish per key at a time
- We only want to most recent path to be published
- If a new publish request comes in, cancel the pending one (for example, when pushing a directory of files)

api/buckets/service.go
```diff
@@ -73,33 +77,31 @@ func (s *Service) Init(ctx context.Context, req *pb.InitRequest) (*pb.InitReply,
...
 		ipfsfiles.NewMapDirectory(map[string]ipfsfiles.Node{
-			bucketSeedName: ipfsfiles.NewBytesFile(seed),
+			seedName: ipfsfiles.NewBytesFile(seed),
 		}),
 		options.Unixfs.Pin(true))
 	if err != nil {
 		return nil, err
 	}
 
+	key, err := s.IPFSClient.Key().Generate(ctx, util.MakeToken(keyLen), opt.Key.Type(options.Ed25519Key))
```
sanderpick, line 94: Use the key API to generate the key. We still need our own mapping to thread ID.

api/buckets/service.go
```diff
@@ -403,6 +406,8 @@ func (s *Service) PushPath(server pb.API_PushPathServer) error {
 		return err
 	}
 
+	go s.IPNSManager.publish(dirpth, buck.Key)
```
sanderpick, line 409: Kick off a publish.

gateway/bucketfs.go
```diff
@@ -78,12 +78,12 @@ func serveBucket(fs serveBucketFileSystem) gin.HandlerFunc {
 	}
 }
 
-func (f *bucketFileSystem) GetThread(ctx context.Context, key string) (id thread.ID, err error) {
-	ik, err := f.keys.Get(ctx, key)
+func (f *bucketFileSystem) GetThread(ctx context.Context, bkey string) (id thread.ID, err error) {
+	key, err := f.keys.GetByCid(ctx, bkey)
```
sanderpick, line 82: Very much the same as before, but the key is indexed by name (for the core IPFS API), but we want CID here.

api/buckets/ipnsmanager.go
```diff
@@ -0,0 +1,102 @@
...
+		cancel()
+		m.ctxsLock.Lock()
+		delete(m.ctxs, keyID)
+		m.ctxsLock.Unlock()
+		<-ptl
+	default:
+		m.ctxsLock.Lock()
+		cancel, ok := m.ctxs[keyID]
+		if !ok {
+			return
```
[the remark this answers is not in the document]
jsign: I think I'd structure this method differently.
But a quick solution for now could be:
- Poll `m.ctx[keyID]` a bit with a really small sleep to solve the situation above. So, never return on `!ok`. `!ok`== sleep&retry.
- Call `cancel()` as is in 74
- Avoid de recursive 75, and switch this method to a `for` loop. If we loop more than X times, abort mission since something bad is happening. At least we don't have a StackOverflow panic with the recursive call.

Another idea is after canceling doing `ptl <-struct{}{}`, so the next line knows that the cancelation finished, so can `<-ptl` and retry as is.

We can create an issue to find a more elegant solution later... throwing ideas now to unblock if this in hotpath.

api/buckets/service.go
```diff
@@ -116,6 +118,7 @@ func (s *Service) createBucket(ctx context.Context, dbID thread.ID, dbToken thre
 			}
 		}
 	}
+	go s.IPNSManager.publish(pth, buck.Key)
```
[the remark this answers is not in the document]
sanderpick: I think that would mean having a service level context, since using the request contexts are short lived compared to what's needed for publishing. 

There's a `Cancel` method on the ipns manager that was meant to solve this issue, but I forgot to wire it up core's close method.

api/buckets/ipnsmanager.go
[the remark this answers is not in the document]
sanderpick: Nice, good eye! It's a tricky pattern... like a FILO stack the should always be size one. 

I added a loop to the select statement, negating the recursion, and cleaned up the missed unlock. The default case is allowed to come up empty handed in the _cancel func map_ only `maxCancelPublishTries` times. However, it will run a bunch of times while the publish context is actually cancelling (as you mentioned above, it takes some time... about 2 millis on my machine). I did some testing with consecutive `go pushlish()` calls, and pushing large directories. I've only seen `try` bump to 1. So, I have arbitrarily set `maxCancelPublishTries` to 10.

PR #167, opened 2020-04-30
Allow init existing bucket

sanderpick:
- Adds some handling to deal with the presence or `--key` and `--thread` when initing a bucket. 
- Fixes interacting with orgs that have a name that doesn't equal the URL slug, e.g., "my org" vs. "my-org"... this was causing the "Not a member error"
- Hides textile configs in the bucket UI


cmd/tt/bucket.go
```diff
@@ -76,52 +76,78 @@ Existing configs will not be overwritten.
...
+			var err error
+			name, err = prompt.Run()
+			if err != nil {
+				cmd.End("")
+			}
+		}
 
-		var dbID thread.ID
-		if selected.ID == "Create new" {
+		if !dbID.Defined() {
```
sanderpick, line 110: You can use `--thread` without key, but not `--key` without thread

sanderpick:
When you init _any_ bucket, we _could_ make sure the config file is pushed to the remote, since the config file will be needed in the gateway to make some decisions like public/private, whitelisted domains, etc. However, I opted to not do this now since it's actually not currently read on the remote.

PR #168, opened 2020-04-30
Materialize Buckets

asutula:
Ok spun my wheels on this for a bit, but the solution was pretty simple and clean. Basically, you can still manually set flags on `tt bucket init` to existing thread and key, or now, you can just pass the `--existing` flag and it will interactively prompt you to choose a remote bucket for whatever org you're using (org is set with the `--org` flag like before)


cmd/tt/bucket.go
```diff
@@ -76,6 +78,62 @@ Existing configs will not be overwritten.
...
+					}
+					bucketInfos = append(bucketInfos, bucketInfo{ID: id, Name: name, Key: root.Key})
+				}
+			}
+
+			prompt := promptui.Select{
+				Label: "Which exiting Bucket do you want to init from?",
+				Items: bucketInfos,
+				Templates: &promptui.SelectTemplates{
+					Active:   fmt.Sprintf(`{{ "%s" | cyan }} {{ .Name | bold }} {{ .Key | faint | bold }}`, promptui.IconSelect),
```
asutula, line 121: in the case of an unnamed bucket, not sure which is best to display as a backup way to differentiate buckets... key or thread id?

asutula:
We can still make a sub command that simple prints your bucket thread ids and keys so you can use them as flags, but this is a simpler ux for now. Let me know if I should add that, though.

cmd/tt/bucket.go
sanderpick: I'd say key as it's unique to the bucket

cmd/tt/main.go
```diff
@@ -58,6 +58,10 @@ var (
 			Key:      "thread",
 			DefValue: "",
 		},
+		"existing": {
```
sanderpick, line 61: Should this live on in the bucket config?

asutula: it's only used for determining how to create the local bucket, so it doesn't need to be saved anywhere after that. I tried just making it a normal flag not bound with viper, but had some problem. I can try that again.

sanderpick: It should work if you put the flag definition _after_ calling _cmd.BindFlags_

asutula: All good, I think previously I was getting the flag value from viper instead of `c`. All fixed now.

PR #178, opened 2020-05-01
add a cmd to print bucket links any time

asutula:
I think this is good, but I'm having trouble testing it locally because docker doesn't seem to pick up my server side code changes. Might as well get some feedback now tho.


api/buckets/client/client.go
```diff
@@ -47,6 +47,12 @@ func (c *Client) Init(ctx context.Context, name string) (*pb.InitReply, error) {
 	})
 }
 
+func (c *Client) Links(ctx context.Context, key string) (*pb.LinksReply, error) {
```
sanderpick, line 50: Nit: Doc string?

asutula: oops, adding

PR #192, opened 2020-05-05
Adds destroy bucket command and some other nits

### LINKED ISSUE #171: Add a destroy bucket command

### LINKED ISSUE #179: cli: --org sometimes has no effect
when passed to the `tt threads ls` command

### LINKED ISSUE #191: Encourage use of existing db for new buckets
As a first pass, just show the existing dbs above _Create new_ in the select list.

sanderpick:
Closes #171 
Closes #179  
Closes #191 

- Improves the UX around choosing a threadDB for a new bucket:
  1. If the account (dev/org) has no DBs, they _are not_ given the option of naming the DB... it will be called "default"
  2. Next time they need to choose a DB, the "default" one is shown on top
  3. If they choose to create a new one, they are then asked what to name it
- Adds `tt bucket destroy`
- Improves UX with _dangerous_ actions like,
  - `tt destroy`: Dev must type their username
  - `tt orgs destroy`: Dev must type the full name of the org
  - `tt bucket destroy`: Dev must confirm
- Fixes `tt threads` and `tt keys` to use the `--org` flag when present (it will still pick up the config setting within a bucket, but the flag will always override)
- Adds a conditional on account deletion: _Owned orgs_ must first be deleted
- An account cleanup fix: Invites from the deleted account were not properly removed

### DIFF
--- a/cmd/tt/orgs.go
+++ b/cmd/tt/orgs.go
@@ -236,11 +236,11 @@ var destroyOrgsCmd = &cobra.Command{
 	Short: "Destroy an org",
 	Long:  `Destroy an organization and all associated data (interactive). You must be the org owner.`,
 	Run: func(c *cobra.Command, args []string) {
-		selected := selectOrg("Remove org", aurora.Sprintf(
-			aurora.BrightBlack("> Removing org {{ .Name | white | bold }}")))
+		selected := selectOrg("Destroy org", aurora.Sprintf(
+			aurora.BrightBlack("> Destroying org {{ .Name | white | bold }}")))
 		configViper.Set("org", selected.Slug)
 
-		cmd.Message("Are you absolutely sure? This action cannot be undone. The org and all associated data will be permanently deleted.")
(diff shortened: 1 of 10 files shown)


cmd/tt/orgs.go
```diff
@@ -257,3 +230,35 @@ func selectOrg(label, successMsg string) *orgItem {
 	}
 	return items[index]
 }
+
+var destroyOrgsCmd = &cobra.Command{
```
asutula, line 234: Super duper nit, but just for clarity should this be `destroyOrgCmd`? (not plural)

sanderpick: I been using the naming pattern `blahRootCmd`, where the root is the main command as plural. So, `invalidateKeysCmd`, etc., since with the exception of buckets everything is aliased such that plural is the same as singular.

sanderpick:
Great, thanks for the eyes 🙇

PR #200, opened 2020-05-12
Make subdomain handling configurable

sanderpick:
- Using subdomains on localhost it cumbersome and some setups may not want to deal with the extra cert overhead needed to properly support subdomains
- Handle redirects in a middleware (now optional)
- Paths and CIDs are always shown in base32 
- Cleans up some camel-casing where it should be all upper


sanderpick:
Cheers 🙏

PR #202, opened 2020-05-12
Buckets: Transform roots to Cid V1

### LINKED ISSUE #133: Error when updating an existing bucket
Sometimes fails with `Error! 'From' cid was not recursively pinned already`

Once a bucket fails with this error, it doesn't seem possible to recover. 

https://github.com/textileio/gatsby-ipfs-blog/runs/515379641#step:6:20 

short term solution is to start a new bucket

### LINKED ISSUE #196: Impossible to register with a slow email setup
It can easily take several minutes for me to receive an email. This makes it impossible to confirm my email to register an account.

Ideally, the initialization process would:

1. Be resumable.
2. Re-poll the textile hub when it times out.

### LINKED ISSUE #201: Use Cid v1 as bucket roots

sanderpick:
Closes #201 
Closes #196 
Closes #133

### DIFF
--- a/api/buckets/service.go
+++ b/api/buckets/service.go
@@ -64,7 +64,7 @@ func (s *Service) Init(ctx context.Context, req *pb.InitRequest) (*pb.InitReply,
 		Root: &pb.Root{
 			Key:       buck.Key,
 			Name:      buck.Name,
-			Path:      util.Base32Path(buck.Path),
+			Path:      buck.Path,
 			CreatedAt: buck.CreatedAt,
 			UpdatedAt: buck.UpdatedAt,
 		},
@@ -82,8 +82,12 @@ func (s *Service) createBucket(ctx context.Context, dbID thread.ID, dbToken thre
 		ipfsfiles.NewMapDirectory(map[string]ipfsfiles.Node{
 			seedName: ipfsfiles.NewBytesFile(seed),
 		}),
-		options.Unixfs.Pin(true))
+		options.Unixfs.Pin(false))
 	if err != nil {
+		return nil, err
+	}
+	pthv1 := path.New(util.ToCidV1Path(pth.String()))
+	if err = s.IPFSClient.Pin().Add(ctx, pthv1); err != nil {
 		return nil, err
 	}
 
@@ -91,7 +95,7 @@ func (s *Service) createBucket(ctx context.Context, dbID thread.ID, dbToken thre
 	if err != nil {
 		return nil, err
 	}
-	buck, err := s.Buckets.Create(ctx, dbID, bkey, name, pth, WithToken(dbToken))
+	buck, err := s.Buckets.Create(ctx, dbID, bkey, name, pthv1, WithToken(dbToken))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


api/buckets/service.go
```diff
@@ -82,16 +82,20 @@ func (s *Service) createBucket(ctx context.Context, dbID thread.ID, dbToken thre
 		ipfsfiles.NewMapDirectory(map[string]ipfsfiles.Node{
 			seedName: ipfsfiles.NewBytesFile(seed),
 		}),
-		options.Unixfs.Pin(true))
+		options.Unixfs.Pin(false))
```
jsign, line 85: Any words on this change?
Seems like as is no Cids are pinned. I'm not sure how the Pin update behaves if its moving a pin which wasn't `.Pin(true)`. Also, if the IPFS node runs GC before a further PushPath is done, we'll lose the bucket.

api/buckets/service.go
```diff
@@ -398,13 +402,18 @@ func (s *Service) PushPath(server pb.API_PushPathServer) error {
...
+	dirpth, err := s.IPFSClient.Object().AddLink(server.Context(), buckPath, filePath, pth, options.Object.Create(true))
 	if err != nil {
 		return err
 	}
 	if err = s.IPFSClient.Pin().Update(server.Context(), buckPath, dirpth); err != nil {
-		return err
+		if err.Error() == "'from' cid was not recursively pinned already" {
+			if err = s.IPFSClient.Pin().Add(server.Context(), dirpth); err != nil {
+				return err
+			}
```
jsign, line 413: Uhm, I see. This answers my last question. A bit worried about Error string changes and breaking this logic.
Maybe your wording about why the bucket now isn't pinned from start will light this.

sanderpick: Yeah, there are not sentinels from the HTTP client :( Maybe an issue over there would be a good idea.

api/buckets/service.go
sanderpick: The problem with pinning here is that the Cid V0 is pinned, and then below when we update it, it will error w/ "'from' cid was not recursively pinned already", since we'll be asking about the Cid V1. Instead, we add it, then on the next line, manually (recursively) pin the Cid in V1 form.

api/buckets/service.go
sanderpick: Catching this error here just in case something weird happened with the bucket (as reported by some users... though that was an old report). Confusingly, this is not related to the logic in `createBucket`... but pinning the wrong version of the cid would also pop this error.

sanderpick: Issue: https://github.com/ipfs/go-ipfs-http-client/issues/100

api/buckets/service.go
jsign: Ah, I see. Right, the ipfs-node is talking about V0, and you want to pin V1.
Uhm, how unfortunate. It would be good that the ipfs-node can see that both Cids are referring to the same sha256 data. I missed the `.Pin().Add()` below, so looks good.

api/buckets/service.go
jsign: Yep, sounds good.

api/buckets/service.go
sanderpick: Oh, actually looks like there a cid version option with add... I'll try that.

jsign: Sounds nice!

sanderpick: Well damn, looks like I can ditch that util method all together after specifying the cid version with add.

jsign: Nice! :)

PR #220, opened 2020-05-30
Handle collection updates

sanderpick:
Small one here... catches invalid schema errors and force updates the collection.

### DIFF
--- a/api/buckets/service.go
+++ b/api/buckets/service.go
@@ -553,7 +553,13 @@ func (s *Service) RemovePath(ctx context.Context, req *pb.RemovePathRequest) (*p
 		return nil, err
 	}
 	if err = s.IPFSClient.Pin().Update(ctx, buckPath, dirpth); err != nil {
-		return nil, err
+		if err.Error() == "'from' cid was not recursively pinned already" {
+			if err = s.IPFSClient.Pin().Add(ctx, dirpth); err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
 	}
 	buck.Path = dirpth.String()
 	buck.UpdatedAt = time.Now().UnixNano()
--- a/buckets/buckets.go
+++ b/buckets/buckets.go
@@ -131,7 +131,7 @@ func (b *Buckets) Create(ctx context.Context, dbID thread.ID, key, name string, 
 	}
 	ids, err := b.threads.Create(ctx, dbID, CollectionName, dbc.Instances{bucket}, db.WithTxnToken(args.Token))
 	if err != nil {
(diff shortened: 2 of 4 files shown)


buckets/buckets.go
```diff
@@ -212,7 +212,24 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
+				return err
+			}
+			return b.Save(ctx, dbID, bucket, opts...)
+		}
+		return err
+	}
+	return nil
+}
+
+func ensureNoNulls(b *Bucket) {
```
sanderpick, line 229: We'll have to continue to stack field migrations here until there's a better way (see https://github.com/textileio/go-threads/issues/363)

api/buckets/service.go
```diff
@@ -553,7 +553,13 @@ func (s *Service) RemovePath(ctx context.Context, req *pb.RemovePathRequest) (*p
 		return nil, err
 	}
 	if err = s.IPFSClient.Pin().Update(ctx, buckPath, dirpth); err != nil {
-		return nil, err
+		if err.Error() == "'from' cid was not recursively pinned already" {
```
jsign, line 556: Ouch! :(

buckets/buckets.go
```diff
@@ -212,7 +212,24 @@ func (b *Buckets) Save(ctx context.Context, dbID thread.ID, bucket *Bucket, opts
...
-	return b.threads.Save(ctx, dbID, CollectionName, dbc.Instances{bucket}, db.WithTxnToken(args.Token))
+	ensureNoNulls(bucket)
+	err := b.threads.Save(ctx, dbID, CollectionName, dbc.Instances{bucket}, db.WithTxnToken(args.Token))
+	if err != nil {
+		if isInvalidSchemaErr(err) {
+			if err := b.updateCollection(ctx, dbID, opts...); err != nil {
+				return err
+			}
+			return b.Save(ctx, dbID, bucket, opts...)
+		}
```
jsign, line 223: Maybe make `isInvalidSchemaErr` do the check `err != nil` and avoid some nesting?
Like:
```go
if isInvalidSchemaErr(err) {
   ...
}
if err != nil {
   return fmt.Errorf("saving bucket in thread: %s", err)
}
return nil

sanderpick: Sure, can do

api/buckets/service.go
sanderpick: No big deal, this was only reachable by encountering bucket save errors from a schema mismatch... so, shouldn't have happened to anyone here.

PR #227, opened 2020-06-02
Support cid v0 local buckets

sanderpick:
I ran into the issue where I could not get to an "up-to-date" state locally since the local bucket was constructing a cid v1 tree.
- Allows a local bucket to be either cid v0 or cid v1
- Detects remote cid version when needed locally
- Handles bucket schema updates when creating a new bucket in an older thread

### DIFF
--- a/api/buckets/client/client.go
+++ b/api/buckets/client/client.go
@@ -51,6 +51,13 @@ func (c *Client) Init(ctx context.Context, name string) (*pb.InitReply, error) {
 }
 
 // Links returns a list of links that can be used to view the bucket
+func (c *Client) Root(ctx context.Context, key string) (*pb.RootReply, error) {
+	return c.c.Root(ctx, &pb.RootRequest{
+		Key: key,
+	})
+}
+
+// Links returns a list of links that can be used to view the bucket
 func (c *Client) Links(ctx context.Context, key string) (*pb.LinksReply, error) {
 	return c.c.Links(ctx, &pb.LinksRequest{
 		Key: key,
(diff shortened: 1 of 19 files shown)


cmd/buck/cli/cli.go
```diff
@@ -97,3 +100,19 @@ func printLinks(reply *pb.LinksReply) {
 		cmd.Message("%s Bucket website", aurora.White(reply.WWW).Bold())
 	}
 }
+
+func setCidVersion(buck *local.Bucket, key string) {
```
sanderpick, line 104: This does nothing if the local bucket root is defined. Otherwise, it will get the root and cid type from remote.

buckets/buckets.go
```diff
@@ -412,13 +418,9 @@ Loop:
...
 
 func (b *Buckets) saveDealsInArchive(ctx context.Context, key string, dbID thread.ID, ffsToken string, c cid.Cid, opts ...Option) error {
-	args := &Options{}
-	for _, opt := range opts {
-		opt(args)
-	}
-	buck := &Bucket{}
-	if err := b.threads.FindByID(ctx, dbID, CollectionName, key, buck, db.WithTxnToken(args.Token)); err != nil {
-		return fmt.Errorf("finding bucket: %s", err)
+	buck, err := b.Get(ctx, dbID, key, opts...)
```
sanderpick, line 421: Using bucket methods here which are wired to handle adding / updating the collection schema.

api/buckets/pb/buckets.proto
```diff
@@ -166,6 +174,7 @@ message ArchiveWatchReply {
 
 service API {
     rpc Init(InitRequest) returns (InitReply) {}
+    rpc Root(RootRequest) returns (RootReply) {}
```
sanderpick, line 177: New `Root` method which just returns the root.

api/buckets/client/client.go
```diff
@@ -50,6 +50,13 @@ func (c *Client) Init(ctx context.Context, name string) (*pb.InitReply, error) {
 	})
 }
 
+// Links returns a list of links that can be used to view the bucket
+func (c *Client) Root(ctx context.Context, key string) (*pb.RootReply, error) {
```
jsign, line 54: Uhm, some comment mismatch?

buckets/buckets.go
```diff
@@ -136,6 +136,12 @@ func (b *Buckets) Create(ctx context.Context, dbID thread.ID, key, name string,
 		}
 		return b.Create(ctx, dbID, key, name, pth, opts...)
 	}
+	if isInvalidSchemaErr(err) {
+		if err := b.updateCollection(ctx, dbID, opts...); err != nil {
+			return nil, err
+		}
+		return b.Create(ctx, dbID, key, name, pth, opts...)
```
jsign, line 143: Q: So if creation fails because of a schema mismatch, we'll automatically update the collection schema?

buckets/local/bucket.go
```diff
@@ -89,10 +91,17 @@ func (b *Bucket) load() error {
...
+		b.cidver = int(b.root.Version())
 	}
 	return nil
 }
 
+// CidVersion returns the configured cid version (0 or 1).
+// The default is 1.
+func (b *Bucket) SetCidVersion(v int) {
+	b.cidver = v
+}
```
jsign, line 103: So this is used to be able to match the local bucket with the cid version used remotely?

cmd/buck/cli/cli.go
```diff
@@ -97,3 +100,19 @@ func printLinks(reply *pb.LinksReply) {
...
+		defer cancel()
+		root, err := clients.Buckets.Root(ctx, key)
+		if err != nil {
+			cmd.Fatal(err)
+		}
+		rp, err := util.NewResolvedPath(root.Root.Path)
+		if err != nil {
+			cmd.Fatal(err)
+		}
+		buck.SetCidVersion(int(rp.Root().Version()))
```
jsign, line 116: OK, nice. Seems to answer my above question: The local bucket "local DAG" will use the same Cid version as the remote backend is using.

buckets/buckets.go
sanderpick: Yep, too heavy handed? We might have to revisit in a multi-writer world, but here it seems easiest to just not involve the user in the migration process.

api/buckets/client/client.go
sanderpick: Oop will fix

buckets/buckets.go
jsign: Yeah, sounds good. The only risk is having an unwanted/error change this side, and changing the schema to an unwanted thing. But I guess that would be reversible, right?

sanderpick: Yep, totally

PR #245, opened 2020-06-29
configurable Buckets CidConfig & improvements/fixes

### LINKED ISSUE #164: pg: richer configuration for ffs instance
PG allows configuring a bunch of things on how the Cid should be stored. 
Maybe the most important one is the _Replication Factor_ but many more exist.

At least the most important ones should be allowed to be configured per-bucket. Most probably the configuration selected for a bucket will be the default-config of the FFS instance, so each new cid push always uses that config.

### LINKED ISSUE #242: pg: check zero balance before archiving
Considering FFS funding can take ~25s, if the client archives too fast it might fail since hasn't enough funds yet.
Check for zero balances and fail fast.

### LINKED ISSUE #255: archiving: better handling of edge cases
When archiving a bucket, we can find many situations:
0. No previous archive: push cid.
1. Last archive exists with the same Cid:
  a. Last archive Successful
  b. Last archive Executing/Queued.
  c. Last archive Failed/Canceled
2. Archiving on new Cid

Each case should be handled in the most natural way.

jsign:
Closes #164: Includes in the `powd` configuration, a `ffs.DefaultCidConfig` attribute. All new buckets are created and configured with this `ffs.DefaultCidConfig` so further archives using the FFS instance will store in the provided way. If no value is provided, a reasonable default is used (which now is somewhat tuned for Testnet).

---

Closes #242 

---

Closes #255: A review comment on the decision of each case.


cmd/buck/cli/archive.go
```diff
@@ -20,6 +20,14 @@ var bucketArchiveCmd = &cobra.Command{
...
 	},
 	Run: func(c *cobra.Command, args []string) {
+		cmd.Warn("Warning! Archiving is based on an experimental testnet. Your data may be lost at any time.")
+		prompt := promptui.Prompt{
+			Label:     "Proceed",
+			IsConfirm: true,
+		}
+		if _, err := prompt.Run(); err != nil {
+			cmd.End("")
+		}
```
jsign, line 30: Some extra warning. Leveraged suggestion by @sanderpick

cmd/cmd.go
```diff
@@ -11,7 +11,6 @@ import (
 	"github.com/spf13/viper"
 	tc "github.com/textileio/go-threads/api/client"
 	"github.com/textileio/go-threads/core/thread"
-	"github.com/textileio/powergate/cmd/pow/cmd"
```
jsign, line 14: Unnecessary dependency.

collections/ffs.go
```diff
@@ -8,9 +8,10 @@ import (
 )
 
 type FFSInstance struct {
-	BucketKey string   `bson:"_id"`
-	FFSToken  string   `bson:"ffs_token"`
-	Archives  Archives `bson:"archives"`
+	BucketKey  string   `bson:"_id"`
+	FFSToken   string   `bson:"ffs_token"`
+	WalletAddr string   `bson:"ffs_walletaddr"`
```
jsign, line 13: As mentioned before, the Bucket wallet address is saved now too.

buckets/buckets.go
```diff
@@ -270,9 +320,14 @@ func (b *Buckets) Delete(ctx context.Context, dbID thread.ID, key string, opts .
...
-// If it's not the first archive, we leverage the Replace() API to more efficient pinning.
-// The operation is async, so it will return as soon as the archiving is scheduled in Powergate.
-func (b *Buckets) Archive(ctx context.Context, dbID thread.ID, key string, c cid.Cid, opts ...Option) error {
+// The behaviour changes depending on different cases, depending on a previous archive.
+// 0. No previous archive or last one aborted: simply pushes the Cid to the FFS instance.
+// 1. Last archive exists with the same Cid:
+//   a. Last archive Successful: fails, there's nothing to do.
+//   b. Last archive Executing/Queued: fails, that work already starting and is in progress.
+//   c. Last archive Failed/Canceled: work to do, push again with override flag to try again.
+// 2. Archiving on new Cid: work to do, it will always call Replace(,) in the FFS instance.
```
jsign, line 329: Changes in `Archive` are mostly to have well-defined behavior regarding multiple scenarios.
The comment tries to be quite explicit about the scenarios and what the logic decides to do. I think these are reasonable behaviors for each case.

I'm planning to switch each error case in a sentinel error, and permeate them in a reasonable way through the API, so the CLI can show a nicer message in each case. Left that for the next PR since it involves looking at a reasonable way to check controlled errors in gRPC in this API. (We somewhat chatted about that before)

buckets/buckets.go
```diff
@@ -39,10 +40,34 @@ var (
 	// can be retrieved, since the bucket was never archived.
 	ErrNoCurrentArchive = fmt.Errorf("the bucket was never archived")
 
+	// ErrBalance is returned when archiving a bucket which
+	// underlying FFS instance balance is zero.
+	ErrZeroBalance = errors.New("bucket FIL balance is zero, if recently created wait 30s")
```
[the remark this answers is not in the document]
sanderpick: Nit: r/ErrBalance/ErrZeroBalance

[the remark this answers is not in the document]
sanderpick: Nope, still comparing error strings at the CLI level

buckets/buckets.go
```diff
@@ -281,37 +336,70 @@ func (b *Buckets) Archive(ctx context.Context, dbID thread.ID, key string, c cid
...
 	ctxFFS := context.WithValue(ctx, powc.AuthKey, ffsi.FFSToken)
+
+	// Check that FFS wallet addr balance is > 0, if not, fail fast.
+	bal, err := b.pgClient.Wallet.WalletBalance(ctx, ffsi.WalletAddr)
+	if err != nil {
+		return fmt.Errorf("getting ffs wallet address balance: %s", err)
+	}
+	if bal == 0 {
+		return ErrZeroBalance
+	}
```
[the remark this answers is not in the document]
sanderpick: Cool, +1 for the conclusion of that thread: _fail fast_

buckets/buckets.go
sanderpick: Sweet

buckets/buckets.go
```diff
@@ -171,18 +205,34 @@ func (b *Buckets) IsArchivingEnabled() bool {
...
 	}
+	defaultBucketCidConfig := ffs.DefaultConfig{
+		Cold:       b.buckCidConfig.Cold,
+		Hot:        b.buckCidConfig.Hot,
+		Repairable: b.buckCidConfig.Repairable,
+	}
+	defaultBucketCidConfig.Cold.Filecoin.Addr = waddr
+	if err := b.pgClient.FFS.SetDefaultConfig(ctxFFS, defaultBucketCidConfig); err != nil {
+		return fmt.Errorf("setting default bucket FFS cidconfig: %s", err)
+	}
```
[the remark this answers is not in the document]
sanderpick: Sounds good! I suppose the wallet address could be written to the actual bucket model, which would make the fact that a bucket is backed by FIL more transparent, but not sure if there are any security risks there.

cmd/buck/cli/archive.go
sanderpick: Possibly more explicit: _Warning! Archives are currently saved on an experimental test network. They may be lost at any time._

cmd/buck/cli/archive.go
```diff
@@ -68,15 +76,15 @@ var bucketArchiveStatusCmd = &cobra.Command{
 			fmt.Printf("\n")
 			cmd.Message("Cid logs:")
 			ch := make(chan string)
-			wCtx, cancel := context.WithCancel(context.Background())
+			wCtx, cancel := clients.Ctx.Auth(cmd.TimeoutArchiveWatch)
```
sanderpick, line 79: 👍

buckets/buckets.go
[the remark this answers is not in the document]
jsign: Well, a priori shouldn't be a big risk regarding if somebody know only your wallet address, can't do anything _per se_.
On some more paranoid mindset, maybe someone wouldn't like to expose that since having the wallet address means you can check how much balance is in there... so maybe that's preferred to be avoided?

We can think it a bit more, or see if people ask for this and add it easily.

buckets/buckets.go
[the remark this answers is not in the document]
jsign: > Nit: r/ErrBalance/ErrZeroBalance

Oops, will change.

cmd/buck/cli/archive.go
jsign: Changing.

buckets/buckets.go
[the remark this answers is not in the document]
sanderpick: Cool, sounds good. They could always use the `--private` flag to hide the address if it was on the bucket itself.

PR #253, opened 2020-06-30
Iteratively encrypt existing dags

### LINKED ISSUE #246: Parallelize existing DAG enryption

sanderpick:
### Remote Buckets
- Migrates existing dag encryption to an iterative solution
- Migrates existing dad encryption to handle leaf nodes in parallel

Closes #246

### DIFF
--- a/api/buckets/service.go
+++ b/api/buckets/service.go
@@ -32,6 +32,7 @@ import (
 	"github.com/textileio/textile/dns"
 	"github.com/textileio/textile/ipns"
 	"github.com/textileio/textile/util"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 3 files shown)


api/buckets/service.go
```diff
@@ -290,126 +289,194 @@ func (s *Service) newDirFromExistingPath(ctx context.Context, pth path.Path, key
...
-				if err := dir.AddNodeLink(l.name, l.node); err != nil {
-					return nil, err
+// This method returns a map of all nodes keyed by their _original_ plaintext cid.
+func (s *Service) encryptDag(ctx context.Context, ds ipld.DAGService, root ipld.Node, key []byte, add *namedNode) (map[cid.Cid]*namedNode, error) {
+	// Step 1: Create a preordered list of joint and leaf nodes
+	var stack, joints, leaves []*namedNode
+	stack = append(stack, &namedNode{node: root})
+	var cur *namedNode
+
+	for len(stack) > 0 {
```
jsign, line 353: Cool, some good'old iterative walk.

api/buckets/service.go
```diff
@@ -290,126 +289,194 @@ func (s *Service) newDirFromExistingPath(ctx context.Context, pth path.Path, key
...
+	go func() {
+		wg.Wait()
+		close(wgDone)
+	}()
+	select {
+	case <-wgDone:
+		break
+	case err := <-errs:
+		close(errs)
+		return nil, err
```
jsign, line 421: I see some possible problems here, so speaking loud to double-check:

In a weird scenario, `wgDone` could be closed and still someone used to `errs`. Maybe most probably if the last encryption might fail since both `<-wgDone` and `<-errs` are _ok_ for the select case. So possibly, the select might enter to `wgDone` when some encryption failed. This might blow up later in reconstructing the DAG anyway since it won't find the encrypted node that failed. So maybe, can be ignored... but something tricky.

Another, possibly panicking, problem is that `close(errs)` is done. That means if another spun goroutine in the background also pushes to `errs`, it will panic.

Furthermore, since this returns early (`return nil, err`), the other spun goroutines will still be doing work (even if they don't error, which is a bit _useless_).

A solution might be allowing the goroutines to receive a ctx and just check if `ctx.Err() != nil -> return ` to fail fast. And use [errgroup](https://godoc.org/golang.org/x/sync/errgroup), which basically runs a set of goroutines and auto-cancels the ctx as soon as the first errors and returns that error. Which is pretty similar to the intention here.

api/buckets/service.go
```diff
@@ -290,126 +289,194 @@ func (s *Service) newDirFromExistingPath(ctx context.Context, pth path.Path, key
...
-		defer fn.Close()
-		file := ipfsfiles.ToFile(fn)
-		if file == nil {
-			return nil, fmt.Errorf("node is a directory")
-		}
-		r, err := dcrypto.NewEncrypter(file, key)
-		if err != nil {
-			return nil, err
+	// Step 3: Encrypt joint nodes in reverse, walking up to root
+	for i := len(joints) - 1; i >= 0; i-- {
```
jsign, line 425: So this works since the iterative process is still depth-first, so encrypted nodes should have executed in the right order to always find the encrypted links, and so on.

api/buckets/service.go
sanderpick: Nice, `errgroup` sounds much cleaner. Changed.

api/buckets/service.go
sanderpick: Yep, exactly.

api/buckets/service.go
jsign: Nice!

PR #256, opened 2020-07-01
local: fix local diffing with file nodes

sanderpick:
Fixes nodes comparison where one of the nodes is a _file_ (proto) node, which contain only unnamed links to raw blocks.

### DIFF
--- a/buckets/local/bucket.go
+++ b/buckets/local/bucket.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


buckets/local/bucket.go
```diff
@@ -447,3 +448,107 @@ func addFile(dag ipld.DAGService, layout options.Layout, prefix cid.Prefix, r io
 	}
 	return n, err
 }
+
+// Diff returns a set of changes that transform node 'a' into node 'b'.
+// Modified from https://github.com/ipfs/go-merkledag/blob/master/dagutils/diff.go#L104
+// It only traverses links in the following cases:
+// 1. two node's links number are greater than 0.
+// 2. both of two nodes are ProtoNode.
+// 3. neither of the nodes is a file node, which contains only unnamed raw blocks
```
sanderpick, line 457: This is an added condition.

buckets/local/bucket.go
```diff
@@ -447,3 +448,107 @@ func addFile(dag ipld.DAGService, layout options.Layout, prefix cid.Prefix, r io
...
+	}
+
+	var out []*du.Change
+	cleanA, okA := a.Copy().(*md.ProtoNode)
+	cleanB, okB := b.Copy().(*md.ProtoNode)
+	if !okA || !okB {
+		return getChange(a, b)
+	}
+
+	if isFileNode(cleanA) || isFileNode(cleanB) {
```
sanderpick, line 473: If either is a file node, just mark as a change.

jsign: So if we're here, that's because the link Cids of these files were different(some step before line 487 was `false`). Correct?

sanderpick: Yep yep!

PR #262, opened 2020-07-07
hub: allow limiting resources creation and growth

### LINKED ISSUE #235: Bucket storage limits
All this features (plus initing a bucket with a Cid), puts a bit more pressure on having some kind of limits. 
Targeting remote Cids opens the door for big resource usage (network, disk, etc).

This already happened with user pushed data anyway, they could push TBs of data, so not a new problem. Just adding more pressure and making the problem "bigger". Just to keep in mind.

_Originally posted by @jsign in https://github.com/textileio/textile/diffs_

jsign:
Closes #235.

This PR adds for features to limit the size and amount of resources for an account in the Hub:
- Bucket max size: A bucket must be smaller or equal than the size.
- Buckets total max size: The sum of the sizes of all buckets from the account must be smaller or equal than the limit.
- Maximum number of buckets per Thread.
- Maximum number of threads per Owner.

In all cases, a value of 0 indicates no limits.

All these numbers are tunable by `hubd` `Config` (flags/envs).

The CLI for `push`, controls the folder size limit before sending data to the Hub, as a quick safety net before wasting network resources. Even if this limit is _hacked_, the Hub will do its own controls. (But, of course, has no other option than consuming enough network data to realize about the size it's receiving.)


api/buckets/service.go
```diff
@@ -1166,24 +1203,63 @@ func getLink(lnks []*ipld.Link, name string) *ipld.Link {
...
+	}
+
 	return nil
 }
 
+// dagSize returns the cummulative size of root. If root is nil, it returns 0.
+func (s *Service) dagSize(ctx context.Context, root path.Path) (int64, error) {
+	if root == nil {
+		return 0, nil
+	}
```
jsign, line 1254: Clarified the `root==nil` case in comments. This was to make smoother the `updateOrAddPin` if the `to` was `nil`.

cmd/buck/cli/push.go
```diff
@@ -20,6 +20,11 @@ import (
 	"github.com/textileio/textile/cmd"
 )
 
+var (
+	buckMaxSizeMiB = int64(1024)
```
jsign, line 24: So this is 1Gib. Which is the same as the default total max size of buckets in an account.
That's to say, in the CLI we're covering to worst of the worst case: a single bucket eating the whole *total* size quota of the account.

cmd/hubd/docker-compose.yml
```diff
@@ -14,6 +14,10 @@ services:
       - HUB_ADDR_MONGO_URI
       - HUB_ADDR_IPFS_API=/dns4/ipfs/tcp/5001
       - HUB_ADDR_POWERGATE_API
+      - HUB_BUCKET_MAX_SIZE=1073741824
```
jsign, line 17: 1GiB

cmd/hubd/docker-compose.yml
```diff
@@ -14,6 +14,10 @@ services:
...
+      - HUB_BUCKET_MAX_NUMBER_PER_THREAD=10000
+      - HUB_BUCKET_TOTAL_MAX_SIZE=1073741824
+      - HUB_THREAD_MAX_NUMBER_PER_OWNER=100
```
jsign: Tunable. Discussed a bit with @sanderpick .
Since `HUB_BUCKET_TOTAL_MAX_SIZE` (account-wide limit) came after `HUB_BUCKET_MAX_SIZE` (individual bucket limit) maybe only the first limit makes more sense. Leave the other since doesn't hurt. (Could be 0 to be ignored).

api/buckets/service.go
```diff
@@ -1166,24 +1203,71 @@ func getLink(lnks []*ipld.Link, name string) *ipld.Link {
 // updateOrAddPin moves the pin at from to to.
 // If from is nil, a new pin as placed at to.
 func (s *Service) updateOrAddPin(ctx context.Context, from, to path.Path) error {
```
jsign, line 1205: So a bit of a refactor in `updateOrAddPin`.
Basically, it does the same as it did before (with some if-nesting improvement).
Additionally, it does:
- Get a cached total size of buckets attribute at the Account level.
- Get size of the root which will be unpined (or 0 if there isn't)
- Get size of the root to be pinned.
- Calculate the "future" size of this bucket (before moving any pin)
- Check the limit. If not ok, fail. (No pin was moved)
- If ok, move on and update the total size of the account.

Quite an important method since is used in anything that updates the bucket root, so might worth extra review attention just in case.

api/buckets/service.go
```diff
@@ -1166,24 +1203,71 @@ func getLink(lnks []*ipld.Link, name string) *ipld.Link {
...
+	key, isUser := ownerFromContext(ctx)
+
+	var futureSize int64
+	if !isUser {
```
jsign, line 1209: So, it seems that if the context is from a User, then there's no `Account` in the collection, thus we can't track the size of the total buckets. Is that correct @sanderpick ?  I realized this from some tests failing.
Sounds a bit unfortunate, but maybe we can solve this case in the future. Not completely sure what would be the best solution.

api/buckets/service.go
sanderpick: Just combing though it all myself to see if all cases are handled.
- What about deleting a bucket? There should probably be some handling in `Remove`.
- Most likely this is not handling the encrypted dags properly, where cumulative size is unable to walk the encrypted links. I think we'd need some custom version that uses `getNodesToPath`. 
- There are some cases where nodes are unpinned when handling encrypted dags, i.e., `unpinNodeAndBranch` is used in `SetPath` and indirectly in `RemovePath`
- There are some cases where nodes are pinned directly outside of `updateOrAddPin`, i.e., in `insertNodeAtPath` and in `SetPath` where the encryption key is not nil and we're not replacing the entire root path (`addAndPinNodes`)
- `createBootstrappedPath` could lead to an explosion

That's from a first pass, there may be other edge cases. However, handling all these directly sounds like a bit of a nightmare. So, for these bucket sizes, maybe we can only consider actual files (`RawNode` or `ProtoNode`s that wrap `RawNodes`, which are larger files), and ignore tracking the sizes of _all_ directory nodes, encrypted or not. I think that would just mean tracking a size _in the Bucket schema_, (which might be nice to have anyway), and using that number to reject new incoming files. Thoughts?

api/buckets/service.go
```diff
@@ -1166,24 +1203,71 @@ func getLink(lnks []*ipld.Link, name string) *ipld.Link {
...
+		a, err := s.Collections.Accounts.Get(ctx, key)
```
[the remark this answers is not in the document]
jsign: Yes, I think I'll be adding some extra helper functions to make the "counting" points more straightforward. Maybe something like: `sumBytesPInned(ctx, delta)`, so there can always be the logic of handling the org/dev or user case, and updating the correct collection.
That `delta` can be positive or negative, so just how much the total should be changed.

api/users/client/client_test.go
```diff
@@ -326,6 +327,7 @@ func TestUserBuckets(t *testing.T) {
 	// Initialize a new bucket in the db
 	ctx = common.NewThreadIDContext(ctx, dbID)
 	buck, err := buckets.Init(ctx)
+	fmt.Printf("AHHAHHAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA: %s\n", err)
```
sanderpick, line 330: Nit, remove?

api/buckets/service.go
```diff
@@ -1687,6 +1681,88 @@ func (s *Service) getGatewayHost() (host string, ok bool) {
...
+		return fmt.Errorf("getting user from context: %s", err)
+	}
+	newTotalSize := u.BucketsTotalSize + delta
+	if err := s.Collections.Users.SetBucketsTotalSize(ctx, key, newTotalSize); err != nil {
+		return fmt.Errorf("updating new users buckets total size: %s", err)
+	}
+	return nil
+}
+
+// sumBytesPinned returns the current buckets total size usage of the account/user
```
sanderpick, line 1748: Docstring mismatch

api/buckets/service.go
```diff
@@ -1687,6 +1681,88 @@ func (s *Service) getGatewayHost() (host string, ok bool) {
...
+// logged in the context.
+func (s *Service) getBucketsTotalSize(ctx context.Context) (int64, error) {
+	key, isUser := ownerFromContext(ctx)
+	if !isUser {
+		a, err := s.Collections.Accounts.Get(ctx, key)
```
sanderpick, line 1753: Still seems odd to get the account twice... it's already on the context.

api/users/client/client_test.go
jsign: Haha, good catch!

api/buckets/service.go
jsign: Will change. 💯

api/buckets/service.go
```diff
@@ -1204,11 +1222,6 @@ func getLink(lnks []*ipld.Link, name string) *ipld.Link {
 // updateOrAddPin moves the pin at from to to.
 // If from is nil, a new pin as placed at to.
 func (s *Service) updateOrAddPin(ctx context.Context, from, to path.Path) error {
-	currentBucketsSize, err := s.getBucketsTotalSize(ctx)
-	if err != nil {
-		return fmt.Errorf("getting current buckets total size: %s", err)
-	}
-
```
jsign: Only moved below closer to first use.

api/buckets/service.go
```diff
@@ -1685,15 +1702,11 @@ func (s *Service) unpinPath(ctx context.Context, path path.Path) error {
...
-	if err != nil {
-		return fmt.Errorf("getting node from path: %s", err)
-	}
-	size, err := node.Size()
+	stat, err := s.IPFSClient.Object().Stat(ctx, path)
 	if err != nil {
 		return fmt.Errorf("getting size of removed node: %s", err)
 	}
-	if err := s.sumBytesPinned(ctx, -size); err != nil {
+	if err := s.sumBytesPinned(ctx, int64(-stat.CumulativeSize)); err != nil {
```
jsign: Realized cumulative size is the safest thing, since the size may not account for links.
This was applied in other places.

api/buckets/service.go
```diff
@@ -1702,78 +1715,82 @@ func (s *Service) unpinPath(ctx context.Context, path path.Path) error {
...
+		totalAddedSize += int64(s.CumulativeSize)
+	}
+	currentBucketsSize, err := s.getBucketsTotalSize(ctx)
+	if err != nil {
+		return fmt.Errorf("getting current buckets total size: %s", err)
+	}
+
+	if s.BucketsTotalMaxSize > 0 && currentBucketsSize+totalAddedSize > s.BucketsTotalMaxSize {
+		return ErrBucketsTotalSizeExceedsMaxSize
+	}
```
jsign, line 1733: Checking the size limit on `pinBlocks` was missing!

api/buckets/service.go
```diff
@@ -1702,78 +1715,82 @@ func (s *Service) unpinPath(ctx context.Context, path path.Path) error {
...
-	if !isUser {
-		a, err := s.Collections.Accounts.Get(ctx, key)
-		if err != nil {
-			return fmt.Errorf("getting account from context: %s", err)
-		}
-		newTotalSize := a.BucketsTotalSize + delta
-		if err := s.Collections.Accounts.SetBucketsTotalSize(ctx, key, newTotalSize); err != nil {
+	a := accountFromContext(ctx)
+	if a != nil {
+		a.BucketsTotalSize = a.BucketsTotalSize + delta
```
jsign, line 1750: Update the account struct that lives in the `ctx` (not only send the update to Mongo)

api/buckets/service.go
```diff
@@ -1702,78 +1715,82 @@ func (s *Service) unpinPath(ctx context.Context, path path.Path) error {
...
-	u, err := s.Collections.Users.Get(ctx, key)
-	if err != nil {
-		return fmt.Errorf("getting user from context: %s", err)
+	u := userFromContext(ctx)
+	if u == nil {
+		return fmt.Errorf("context must be from an org/dev/user")
 	}
-	newTotalSize := u.BucketsTotalSize + delta
-	if err := s.Collections.Users.SetBucketsTotalSize(ctx, key, newTotalSize); err != nil {
+	u.BucketsTotalSize = u.BucketsTotalSize + delta
```
jsign, line 1760: Same for the `User` in ctx.

sanderpick:
LGTM! Really nice 💯. I'll merge it this morning as you suggested in Slack.

api/buckets/service.go
```diff
@@ -28,6 +28,7 @@ import (
 	pb "github.com/textileio/textile/api/buckets/pb"
 	"github.com/textileio/textile/api/common"
 	bucks "github.com/textileio/textile/buckets"
+	"github.com/textileio/textile/collections"
```
sanderpick, line 31: Duplicate import... can remove later.

PR #263, opened 2020-07-08
enable keepalive pinging

asutula:
Signed-off-by: Aaron Sutula <hi@asutula.com>

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -65,8 +65,8 @@ var (
 		"/threads.pb.API/ListDBs",
 	}
 
-	// wsPingInterval controls the WebSocket keepalive pinging interval. Must be >= 1s.
-	wsPingInterval = time.Second * 5
+	// WSPingInterval controls the WebSocket keepalive pinging interval. Must be >= 1s.
+	WSPingInterval = time.Second * 5
 )
 
 type Textile struct {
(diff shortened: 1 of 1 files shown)


core/core.go
```diff
@@ -282,6 +283,7 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 		}),
 		grpcweb.WithAllowedRequestHeaders([]string{"Origin"}),
 		grpcweb.WithWebsockets(true),
+		grpcweb.WithWebsocketPingInterval(time.Second*5),
```
sanderpick, line 286: How about a package constant that users can tweak?

carsonfarmer:
I hope you don't mind @asutula, but I added a configurable package constant to tweak this.

core/core.go
carsonfarmer: Added

core/core.go
```diff
@@ -63,6 +64,9 @@ var (
 	blockMethods = []string{
 		"/threads.pb.API/ListDBs",
 	}
+
+	// wsPingInterval controls the WebSocket keepalive pinging interval. Must be >= 1s.
+	wsPingInterval = time.Second * 5
```
sanderpick, line 69: If you make this exported, users can tweak it... an easy win

asutula:
So, how does that variable get set if someone wants to change it @carsonfarmer? I guess there would some flags at the cli cmd level, that would filter down to set that var?

core/core.go
carsonfarmer: 🤦 of course! good call

carsonfarmer:
Based on my understanding, we could tweak it (now that its exported) from `bucks/main.go` and/or `hubs/main.go`, probs via some flags. Seems like we can leave it to the default for now, as I expect we're unlikely to want to change it? But I defs could be wrong and we could defs just add it as flags in there?

asutula:
Ok, I'll merge it and we can address the flags thing potentially in another PR if we decide it's needed.

PR #273, opened 2020-07-15
Add code 200 to no route handler

### LINKED ISSUE #272: Accessing bucket content returns data with 404 response
When I access any content on my bucket (which was uploaded from Textile hub command line) through a browser, I am able to see the contents but the HTTP response code is 404.  This seems to be causing an issue with the tools that I am using since the HTTP response code is not OK. 

For example, if you go to this URL in Brave, you see the XML content but Brave shows a 404 error at the top: https://bafkt5edpkmbejifojfarn2wawtc62t36gy4te7bftd64cgjsy7gv2ui.thread.hub.textile.io/buckets/bafzbeidxwtnl7mwj62s35b75cezp2lzsbbecwcv3gt4kmz5qsc2zlqjhsu/com/github/ericglau/project-a/1.0.0-SNAPSHOT/maven-metadata.xml

If you paste that URL in https://websniffer.cc/ , it also shows that the HTTP response header is 404.

sanderpick:
The subdomain handler is called from the `NoRoute` catchall, which adds code 404 :/

Closes #272


api/buckets/service.go
```diff
@@ -67,15 +67,15 @@ const (
 
 // Service is a gRPC service for buckets.
 type Service struct {
-	Collections              *c.Collections
```
sanderpick, line 70: Ignore these cleanup changes

buckets/local/bucket_test.go
```diff
@@ -27,81 +27,96 @@ func TestBucket(t *testing.T) {
 	buck, err := buckets.NewBucket(context.Background(), conf, WithName("bucky"))
 	require.Nil(t, err)
 
-	// Test Key()
-	assert.NotEmpty(t, buck.Key())
+	t.Run("Key", func(t *testing.T) {
```
sanderpick, line 30: Better test organization... can ignore

cmd/buckd/main.go
```diff
@@ -62,8 +60,8 @@ var (
 				Key:      "addr.ipfs.api",
 				DefValue: "/ip4/127.0.0.1/tcp/5001",
 			},
-			"addrFilecoinApi": {
-				Key:      "addr.filecoin.api",
+			"addrPowergateApi": {
```
sanderpick, line 63: Oop

cmd/hubd/docker-compose-dev.yml
```diff
@@ -12,12 +12,12 @@ services:
       - HUB_ADDR_MONGO_URI=mongodb://mongo:27017
       - HUB_ADDR_IPFS_API=/dns4/ipfs/tcp/5001
       - HUB_ADDR_POWERGATE_API
-      - HUB_BUCKET_MAX_SIZE
-      - HUB_BUCKET_MAX_NUMBER_PER_THREAD
-      - HUB_BUCKET_TOTAL_MAX_SIZE
-      - HUB_THREAD_MAX_NUMBER_PER_OWNER
       - HUB_GATEWAY_SUBDOMAINS
       - HUB_EMAIL_SESSION_SECRET=hubsession
+      - HUB_BUCKETS_MAX_SIZE
```
sanderpick, line 17: Note here @jsign: I changes these vars to be plural (buckets/threads vs bucket/thread) by changing the config key names to match the existing pattern (e.g., `addr.threads`). An annoying change :) but we haven't released the limits stuff yet so wanted to change before then.

gateway/gateway.go
```diff
@@ -346,12 +346,14 @@ func formatError(err error) string {
 
 // subdomainHandler handles requests by parsing the request subdomain.
 func (g *Gateway) subdomainHandler(c *gin.Context) {
+	c.Status(200)
```
sanderpick, line 349: This is the actual gateway fix 😆

cmd/hubd/docker-compose-dev.yml
jsign: Np. Actually I was already using `latest` for the K8 deployment, so already used singular-names, but easy change.

Issue #269, opened 2020-07-14
Textile.Close() does not release badger lock on the `RepoPath`


perfectmak:
### Description
After closing a bootstrapped textile instance gotten from `core.NewTextile()`. An attempt to create a new instance at the same `RepoPath` as the previous one failed with an error similar to the stacktrace:

```
Failed to start textile a second time: resource temporarily unavailable
Cannot acquire directory lock on ".buckd/repo/eventstore".  Another process is using this Badger database.
github.com/dgraph-io/badger.acquireDirectoryLock
	/Users/perfect/go/pkg/mod/github.com/dgraph-io/badger@v1.6.1/dir_unix.go:66
github.com/dgraph-io/badger.Open
	/Users/perfect/go/pkg/mod/github.com/dgraph-io/badger@v1.6.1/db.go:230
github.com/ipfs/go-ds-badger.NewDatastore
	/Users/perfect/go/pkg/mod/github.com/ipfs/go-ds-badger@v0.2.4/datastore.go:137
github.com/textileio/go-threads/db.newDefaultDatastore
	/Users/perfect/go/pkg/mod/github.com/textileio/go-threads@v0.1.21/db/options.go:32
github.com/textileio/go-threads/db.NewManager
	/Users/perfect/go/pkg/mod/github.com/textileio/go-threads@v0.1.21/db/manager.go:46
github.com/textileio/go-threads/api.NewService
	/Users/perfect/go/pkg/mod/github.com/textileio/go-threads@v0.1.21/api/service.go:52
github.com/textileio/textile/core.NewTextile
	/Users/perfect/go/pkg/mod/github.com/textileio/textile@v1.0.13-0.20200707162859-d8131f5afba4/core/core.go:197
main.StartTextile
	/Users/perfect/Terminal/space-poc/examples/bucketIssues/stop_start_issue/main.go:15
main.main
	/Users/perfect/Terminal/space-poc/examples/bucketIssues/stop_start_issue/main.go:48
runtime.main
	/Users/perfect/go/go1.14.3/src/runtime/proc.go:203
runtime.goexit
	/Users/perfect/go/go1.14.3/src/runtime/asm_amd64.s:1373
exit status 1
```

### How to reproduce
Here is a sample code of how to reproduce this issue. Of course, you would need to have mongodb running on the url `mongodb://127.0.0.1:27027`.

```go
import (
	"context"
	"log"
	"time"

	"github.com/textileio/textile/cmd"

[...]

PR #278, opened 2020-07-17
Offline-first bucket watch

### LINKED ISSUE #275: buck.Watch should reconnect when the network dies
This would make it really offline-first.

sanderpick:
Closes #275 

This is the missing piece of the _offline-first_ bucket sync w/ diffing functionality. From the `Watch` docstring: 

```
// Watch watches for and auto-pushes local bucket changes at an interval,
// and listens for and auto-pulls remote changes as they arrive.
// This will survive connectivity interruptions.
// Returns a channel of watch connectivity states. Cancel context to stop watching.
```

Demo: https://bafk7wyub5byqrryei4wbb7smmwqhiiuyhikyxu36o5rq4vixnwkwhia.thread.hub.textile.io/buckets/bafzbeib5zrflczb5scupkw2xd3277yxxycpdsre2uiwzzth3s2t2duljbm/watch_offline.mp4

### DIFF
--- a/buckets/local/bucket_test.go
+++ b/buckets/local/bucket_test.go
@@ -365,10 +365,10 @@ func TestBucket_Watch(t *testing.T) {
 	go ec.collect(events)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	var wg sync.WaitGroup
 	var onlineStateCount, offlineStateCount int
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
-		wg.Add(1)
 		defer wg.Done()
 		state := buck1.Watch(ctx, WithWatchEvents(events))
 		for s := range state {
--- a/buckets/local/bucket_test.go
+++ b/buckets/local/bucket_test.go
@@ -370,7 +370,7 @@ func TestBucket_Watch(t *testing.T) {
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		state, err := buck1.Watch(ctx, WithWatchEvents(events))
+		state, err := buck1.Watch(ctx, WithWatchEvents(events), WithOffline(true))
 		require.Nil(t, err)
 		for s := range state {
 			t.Logf("received watch state: %s", s.State)
--- a/buckets/local/bucket_test.go
+++ b/buckets/local/bucket_test.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -17,8 +18,11 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/textileio/go-threads/db"
+	tutil "github.com/textileio/go-threads/util"
+	"github.com/textileio/textile/api/apitest"
 	bucks "github.com/textileio/textile/buckets"
 	. "github.com/textileio/textile/buckets/local"
+	"github.com/textileio/textile/cmd"
 )
 
 func TestBucket(t *testing.T) {
@@ -342,7 +346,15 @@ func TestBucket_DiffLocal(t *testing.T) {
 }
 
 func TestBucket_Watch(t *testing.T) {
-	buckets1 := setup(t)
+	tconf := apitest.DefaultTextileConfig(t)
+	tconf.Hub = false
+	stopTextile1 := apitest.MakeTextileWithConfig(t, tconf, false)
+	target, err := tutil.TCPAddrFromMultiAddr(tconf.AddrAPI)
+	require.Nil(t, err)
+	clients := cmd.NewClients(target, false)
+	buckets1 := NewBuckets(clients, DefaultConfConfig())
+	defer clients.Close()
+
 	conf := getConf(t, buckets1)
 	buck1, err := buckets1.NewBucket(context.Background(), conf)
 	require.Nil(t, err)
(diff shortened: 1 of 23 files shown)


core/core.go
```diff
@@ -294,6 +294,9 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 		if err := t.server.Serve(listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 			log.Fatalf("serve error: %v", err)
 		}
+		if err := ts.Close(); err != nil {
```
sanderpick, line 297: Bug fix https://github.com/textileio/textile/issues/269

core/core.go
```diff
@@ -363,7 +366,11 @@ func (t *Textile) Close() error {
 	if err := t.proxy.Shutdown(ctx); err != nil {
 		return err
 	}
-	t.server.GracefulStop()
+	if force {
+		t.server.Stop()
```
sanderpick, line 370: This is useful for when you actually don't want to wait for client connections to close. I'm just using in tests right now, but maybe it should be the default here and in go-threads where listeners might cause a very slow or never finishing close.

carsonfarmer: Ah interesting. Good idea. I've never had this issue in testing, but I could see this coming up in the case of a ton of listeners, or even some orphan listeners hanging around?

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/AndreasBriese/bbloom v0.0.0-20190825152654-46b345b51c96 // indirect
 	github.com/alecthomas/jsonschema v0.0.0-20191017121752-4bb6e3fae4f2
 	github.com/caarlos0/spin v1.1.0
+	github.com/cenkalti/backoff v2.2.1+incompatible
```
carsonfarmer, line 9: Is this +incompatible something to worry about?

buckets/local/bucket_test.go
```diff
@@ -353,12 +365,24 @@ func TestBucket_Watch(t *testing.T) {
 	go ec.collect(events)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	var wg sync.WaitGroup
+	var onlineStateCount, offlineStateCount int
 	go func() {
-		err = buck1.Watch(ctx, WithWatchEvents(events))
-		require.Nil(t, err)
+		wg.Add(1)
```
jsign, line 371: Nit: I'd move this add out of `go func()`, since `wg.Wait()` could happen before the addition. Being realistic, with the current logic that's impossible without things failing... but just a conceptual nit.

buckets/local/options.go
```diff
@@ -128,22 +126,12 @@ func WithAddEvents(ch chan<- PathEvent) AddOption {
...
+	events chan<- PathEvent
 }
 
 // WatchOption is used when watching a bucket for changes.
 type WatchOption func(*watchOptions)
 
-// WatchInterval sets the interval at which local changes are synced remotely.
-// In other words, this is the interval at which the local file watcher checks
-// for filesystem changes.
-func WithInterval(local time.Duration) WatchOption {
```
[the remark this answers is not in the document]
jsign: Agree, but I liked it since was an optional thing. :)
No strong opinion here.

buckets/local/watch.go
```diff
@@ -5,92 +5,175 @@ import (
...
 		}
 	}()
+	return outerState
+}
 
-	// Start listening for remote changes
+// WatchWhileConnected watches for and auto-pushes local bucket changes at an interval,
+// and listens for and auto-pulls remote changes as they arrive.
+// This will watch until context is canceled or an error occurs.
+func (b *Bucket) WatchWhileConnected(ctx context.Context, opts ...WatchOption) (<-chan WatchState, error) {
```
jsign, line 90: Nit: Maybe consolidate both `Watch*` APIs in a single one with an option? (e.g: `WithRetry(bool)`).

go.mod
jsign: I don't think so.
@sanderpick, btw, maybe would be better to use `	github.com/cenkalti/backoff/v4`

buckets/local/watch.go
sanderpick: Nice, like it

core/core.go
sanderpick: Looks like the way here: https://github.com/grpc/grpc-go/issues/2448#issuecomment-437531965

```
stopped := make(chan, struct{})
go func () {
  s.GracefulStop()
  close(stopped)
}

t := time.NewTimer(10 * time.Second)
select {
case <- t.C:
  s.Stop()
case <- stopped:
  t.Stop()
}
```

go.mod
sanderpick: Huh, that is the one I used with `go get`, maybe it downgraded due to some incompatibility. `go.sum` has:

```
github.com/cenkalti/backoff v2.2.1+incompatible h1:tNowT99t7UNflLxfYYSlKYsBpXdEet03Pg2g16Swow4=
github.com/cenkalti/backoff v2.2.1+incompatible/go.mod h1:90ReRw6GdpyfrHakVjL/QHaoyV4aDUVVkXQJJJ3NXXM=
github.com/cenkalti/backoff/v3 v3.0.0 h1:ske+9nBpD9qZsTBoF41nW5L+AIuFBKMeze18XQ3eG1c=
github.com/cenkalti/backoff/v3 v3.0.0/go.mod h1:cIeZDE3IrqwwJl6VUwCN6trj1oXrTS4rc0ij+ULvLYs=
```

sanderpick: Oh, maybe IDE picked up the wrong import path from local packages. Will look

jsign: @sanderpick, in buckets/local/watch.go import section should say `"github.com/cenkalti/backoff/v4"`

sanderpick: Hmm, no, `mod tidy` seems to revert to the older version, weird

PR #281, opened 2020-07-19
Update Readme

sanderpick:
Adding a bunch more content here.


api/buckets/service.go
```diff
@@ -412,23 +412,37 @@ func (nn *namedNodes) Store(c cid.Cid, n *namedNode) {
...
+	lmap := make(map[cid.Cid]*namedNode)
 	stack = append(stack, &namedNode{node: root})
 	var cur *namedNode
 
 	for len(stack) > 0 {
 		n := len(stack) - 1
 		cur = stack[n]
 		stack = stack[:n]
+
+		if _, ok := jmap[cur.node.Cid()]; ok {
```
sanderpick, line 426: Don't traverse the node if we already have traversed the exact same on elsewhere.

cmd/buck/cli/init.go
```diff
@@ -33,23 +33,27 @@ Use the '--cid' flag to initialize from an existing UnixFS DAG.
 		conf, err := bucks.NewConfigFromCmd(c, ".")
 		cmd.ErrCheck(err)
 
+		existing := conf.Thread.Defined() && conf.Key != ""
```
sanderpick, line 36: Init from existing fixes

README.md
```diff
@@ -7,45 +7,681 @@
...
+> 	 Deal with miner t0117803 changed state to StorageDealWaitingForDataRequest
+> 	 Deal with miner t0101180 changed state to StorageDealWaitingForDataRequest
+> 	 Deal with miner t0119390 changed state to StorageDealWaitingForDataRequest
+>    Deal with miner t01459 changed state to StorageDealProposalAccepted
+> 	 Deal with miner t01459 changed state to StorageDealSealing
+```
+
+The output will looks something like the above. With a little luck, you will start seeing some successful storage deals.
+
+To-do: Add section on recovering a bucket from an archive using the Lotus client.
```
jsign, line 645: N: In the meantime, it might worth to paste a link to the blog post where there was a video about this?

README.md
```diff
@@ -7,45 +7,681 @@
...
+
+### Multi-writer buckets
+
+Multi-writer buckets leverage the distributed nature of ThreadDB by allowing multiple identities to write to the same bucket hosted by different Libp2p hosts. Since buckets are ThreadDB collection _instances_, this is no different than normal ThreadDB peer collaboration.
+
+To-do: Demonstrate joining a bucket from a ThreadDB invite.
```
jsign, line 651: N: Maybe adding some small comment that mulitple-writers should be responsible for overriding unknown concurrent writes? Might not worth to comment that since is a general problem... but multi-writers open the door to a big set of possible problems if used incorrectly.

andrewxhill:
Yep, lots of good stuff here! I'll def just copy some of it in docs, but great to have here already.

README.md
sanderpick: Sure, good idea

sanderpick:
Cool, feel free to grab sections for the docs, but I'd prefer to keep this as is here for now.

README.md
sanderpick: I'll circle back to this section

PR #288, opened 2020-07-28
fix(tests): add size to rootreply and total size to listreply

andrewxhill:


api/buckets/pb/buckets.proto
```diff
@@ -19,6 +19,7 @@ message ListRequest {}
 
 message ListReply {
     repeated Root roots = 1;
+    int64 size = 2;
```
andrewxhill, line 22: the total size of all buckets was useful during tests but didn't really make sense on the rootreply object. moved it here to accompany a request for all buckets. let me know if that seems okay

api/buckets/pb/buckets.proto
```diff
@@ -40,6 +41,7 @@ message RootRequest {
 
 message RootReply {
     Root root = 1;
+    int64 size = 2;
```
andrewxhill, line 44: new value in response

api/buckets/client/client_test.go
```diff
@@ -654,13 +657,145 @@ func removePath(t *testing.T, ctx context.Context, client *c.Client, private boo
...
 	require.Nil(t, err)
+
 	_, err = client.ListPath(ctx, buck.Root.Key, "again")
 	require.NotNil(t, err)
 	rep, err = client.ListPath(ctx, buck.Root.Key, "")
 	require.Nil(t, err)
 	assert.Equal(t, 2, len(rep.Item.Items))
 }
 
+func TestClient_OverlappingBuckets(t *testing.T) {
```
sanderpick, line 668: Sweet, nice test!

api/buckets/pb/buckets.proto
sanderpick: Nice, makes sense to me

api/buckets/service.go
```diff
@@ -564,6 +570,13 @@ func (s *Service) Root(ctx context.Context, req *pb.RootRequest) (*pb.RootReply,
 	if err != nil {
 		return nil, err
 	}
+
+	buckPath := path.New(buck.Path)
+	stat, err := s.IPFSClient.Object().Stat(ctx, buckPath)
```
jsign, line 575: Uhm, I think this doesn't work correctly for encrypted buckets?
cc @sanderpick

sanderpick: Hrm, yeah... moving to fast the AM. Doing this properly could lead to an explosion... maybe we should avoid returning total sizes after all. Thoughts?

jsign: Yeah, it would take a lot of computation to resolve in the encrypted case at least on the fly.
A solution to avoid that is doing in `sumBytesPinned` a counting also per bucket, so to have that pre-calculated.
In short, instead of only adding size changes account/user-wise, also take advantage and do that bucket wise.

Not sure what to do now though... I liked this test. 
Maybe freeze PR and create an issue if the above makes sense? (not sure is correct, was first idea on my mind)

andrewxhill: yeah, if a hassle, no problem. just thought it was a freebie. removing.

sanderpick: Sounds good. I like the idea of tracking bucket size as well as total buckets size. Let's track an issue for now.

andrewxhill: okay, removed. test is still possible because the cumulative size is still reported in the list api

PR #290, opened 2020-07-28
Hub User Mailboxes

### LINKED ISSUE #233: PKI Inboxing Service
Allow verified PKIs to leave messages for other PKIs. This would still require API access via an API key, perhaps throttled / limited to some limits later on.

sanderpick:
Closes #233

Depends on https://github.com/textileio/go-threads/pull/411

### API
- Adds a user mailbox service and client

### Misc
- Adds a local mailbox library for building the mailbox functionality into existing projects


api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
 func (c *Client) ListThreads(ctx context.Context) (*pb.ListThreadsReply, error) {
 	return c.c.ListThreads(ctx, &pb.ListThreadsRequest{})
 }
+
+// SetupMailbox creates inbox and sentbox threads needed user mail.
+func (c *Client) SetupMail(ctx context.Context) (inbox, sentbox thread.ID, err error) {
```
sanderpick, line 53: Mailbox concept: _A user must opt in by setting up their mailboxes_. In return they will have two threads. One for inbox and one for a sentbox.

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	}
+	sentbox, err = thread.Cast(res.SentboxID)
+	if err != nil {
+		return
+	}
+	return
+}
+
+// Message is the client side representation of a mailbox message.
+type Message struct {
```
sanderpick, line 70: This is all client-side sugar

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	m.Body = body
+	m.Signature = sig
+	m.CreatedAt = time.Unix(0, tm.CreatedAt)
+	m.ReadAt = readAt
+	return nil
+}
+
+// SendMessage sends the message body to a recipient.
+func (c *Client) SendMessage(ctx context.Context, from thread.Identity, to thread.PubKey, body []byte) (msg Message, err error) {
+	fromBody, err := from.GetPublic().Encrypt(body)
```
sanderpick, line 135: To send a message, we create two copies of the message body:
- One encrypted for the recipient that will live in their inbox
- One encrypted for the sender that will live in their sentbox

api/users/client/client_test.go
```diff
@@ -160,99 +160,334 @@ func TestClient_ListThreads(t *testing.T) {
...
+	from, fctx := setupUserMail(t, client, threads, key.Key)
+	to, _ := setupUserMail(t, client, threads, key.Key)
+
+	res, err := client.SendMessage(fctx, from, to.GetPublic(), []byte("howdy"))
+	require.NoError(t, err)
+	assert.NotEmpty(t, res.ID)
+	assert.NotEmpty(t, res.CreatedAt)
+}
+
+func TestClient_ListInboxMessages(t *testing.T) {
```
sanderpick, line 299: Here's a nice list message test.
Message are by default listed in descending order _by ID_. Since ULIDs are based on time, we get time ordering for free, plus efficient pagination by telling the datastore to seek keys.
Depends on https://github.com/textileio/go-threads/pull/411.

api/users/service.go
```diff
@@ -64,3 +75,309 @@ func (s *Service) ListThreads(ctx context.Context, _ *pb.ListThreadsRequest) (*p
...
+
+const (
+	inboxName   = "messages-inbox"
+	sentboxName = "messages-sent"
+
+	defaultMessagePageSize = 100
+	maxMessagePageSize     = 10000
+)
+
+func (s *Service) SetupMail(ctx context.Context, _ *pb.SetupMailRequest) (*pb.SetupMailReply, error) {
```
sanderpick, line 94: Start of new service handlers.

buckets/local/buckets.go
```diff
@@ -24,23 +23,16 @@ var (
 	ErrBucketExists = errors.New("bucket is already initialized")
 	// ErrThreadRequired indicates the operation requires a thread ID but none was given.
 	ErrThreadRequired = errors.New("thread ID is required")
-)
 
-// ConfConfig is used to generate new bucket configs.
```
sanderpick, line 29: Ripping things out of this file and into `cmd`, where they can be used by other local libs.

buckets/local/watch.go
```diff
@@ -3,60 +3,25 @@ package local
...
-	"google.golang.org/grpc/status"
+	"github.com/textileio/textile/cmd"
 )
 
 const (
 	fileSystemWatchInterval = time.Millisecond * 100
 	reconnectInterval       = time.Second * 5
 )
 
-// WatchState is used to inform Watch callers about the connection state.
```
sanderpick, line 22: All of this is also moved to `cmd` where other watchers can use it.

cmd/watch.go
```diff
@@ -0,0 +1,88 @@
...
+type WatchFunc func(context.Context) (<-chan WatchState, error)
+
+// Watch calls watchFunc until it returns an error.
+// Normally, watchFunc would block while doing work that can fail,
+// e.g., the local network goes offline.
+// If watchFunc return an error, it will be called
+// again at the given interval so long as the returned error is non-fatal.
+// Returns a channel of watch connectivity states.
+// Cancel context to stop watching.
+func Watch(ctx context.Context, watchFunc WatchFunc, reconnectInterval time.Duration) (<-chan WatchState, error) {
```
sanderpick, line 55: Here's this thing... can be used to "watch" other functions that might stop blocking due to network interruption.

mail/local/mail.go
```diff
@@ -0,0 +1,183 @@
+package local
```
sanderpick, line 1: Another _local_ library... this time for making mailboxes. Eventually I'll add a repo here for an offline outbox.

mail/local/mail.go
```diff
@@ -0,0 +1,183 @@
...
+	return &Mail{clients: clients, config: config}
+}
+
+// Clients returns the underlying clients object.
+func (m *Mail) Clients() *cmd.Clients {
+	return m.clients
+}
+
+// Config contains details for a new local mailbox.
+type Config struct {
```
sanderpick, line 59: A mailbox is created with an api key + secret and a `thread.Identity`

mail/local/mail.go
```diff
@@ -0,0 +1,183 @@
...
+	APIKey string
+	// APISecret is hub API key secret (optional).
+	// It's value may be inflated from a --api-secret flag or {EnvPrefix}_API_SECRET env variable.
+	APISecret string
+}
+
+// NewConfigFromCmd returns a config by inflating values from the given cobra command and path.
+// First, flags for "identity", "api_key", and "api_secret" are used if they exist.
+// If still unset, the env vars {EnvPrefix}_IDENTITY, {EnvPrefix}_API_KEY, and {EnvPrefix}_API_SECRET are used.
+func (m *Mail) NewConfigFromCmd(c *cobra.Command, pth string) (conf Config, err error) {
```
sanderpick, line 76: Pulled from the buckets lib, this enables a CLI to be built around this lib

mail/local/mailbox.go
```diff
@@ -0,0 +1,256 @@
...
+	MessageRead
+	// MessageDeleted indicates a message was deleted from the mailbox.
+	MessageDeleted
+)
+
+// WatchInbox watches the inbox for new mailbox events.
+// If offline is true, this will keep watching during network interruptions.
+// Returns a channel of watch connectivity states.
+// Cancel context to stop watching.
+func (m *Mailbox) WatchInbox(ctx context.Context, mevents chan<- MailboxEvent, offline bool) (<-chan cmd.WatchState, error) {
```
sanderpick, line 90: You can watch your inbox (or sentbox below) for new messages, read messages, and deleted messages.

mail/local/mailbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+func TestMailbox_Identity(t *testing.T) {
+	mail, key, secret := setup(t)
+
+	conf := getConf(t, key, secret)
+	box, err := mail.NewMailbox(context.Background(), conf)
+	require.NoError(t, err)
+	assert.Equal(t, conf.Identity.GetPublic(), box.Identity().GetPublic())
+}
+
+func TestMailbox_SendMessage(t *testing.T) {
```
sanderpick, line 24: This tests shows the mailbox basic usage... send a message between two users.

mongodb/invites.go
```diff
@@ -1,4 +1,4 @@
-package collections
+package mongodb
```
sanderpick, line 1: There's now a couple threaddb collections. So, I moved all the mongo stuff to `mongodb` and the threads stuff now lives in `threaddb`: `buckets` and `mail` are the two built-in thread collections.

threaddb/mail.go
```diff
@@ -0,0 +1,74 @@
...
+	}, {
+		Path: "created_at",
+	}, {
+		Path: "read_at",
+	}}
+	mailConfig db.CollectionConfig
+)
+
+// Message represents the mail threaddb collection schema.
+type Message struct {
```
sanderpick, line 29: Here's the mail collection schema

threaddb/threaddb.go
```diff
@@ -0,0 +1,146 @@
+package threaddb
```
sanderpick, line 1: New package here that both `buckets` and `mail` can use for creating, saving, deleting, etc., threaddb instances.

mail/local/mail.go
```diff
@@ -0,0 +1,183 @@
...
+	// ErrNotAMailbox indicates the given path is not within a mailbox.
+	ErrNotAMailbox = errors.New("not a mailbox (or any of the parent directories): .textile")
+	// ErrMailboxExists is used during initialization to indicate the path already contains a mailbox.
+	ErrMailboxExists = errors.New("mailbox is already initialized")
+	// ErrIdentityRequired indicates the operation requires a thread Identity but none was given.
+	ErrIdentityRequired = errors.New("thread Identity is required")
+	// ErrAPIKeyRequired indicates the operation requires am API keybut none was given.
+	ErrAPIKeyRequired = errors.New("api key is required")
+
+	flags = map[string]cmd.Flag{
```
sanderpick, line 25: Like `buck` this could be made into a CLI pretty easily... these are the config flags.

mail/local/mailbox.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return nil
+}
+
+func (m *Mailbox) context(ctx context.Context) (context.Context, error) {
+	ctx = common.NewAPIKeyContext(ctx, m.conf.Viper.GetString("api_key"))
+	secret := m.conf.Viper.GetString("api_secret")
+	if secret != "" {
+		var err error
+		ctx, err = common.CreateAPISigContext(ctx, time.Now().Add(time.Hour), secret)
```
jsign, line 242: Q: So if someone is watching (in/sent)boxes for more than an hour, the watching logic might fail by an expired signature, right? 
I think that deadline is OK, mostly to validate understanding in that situation. We can always imagine handling that error in the future and auto-refresh sign.

mail/local/mailbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.NoError(t, err)
+	err = box1.DeleteInboxMessage(context.Background(), sent[1].ID)
+	require.NoError(t, err)
+	err = box2.DeleteSentMessage(context.Background(), sent[2].ID)
+	require.NoError(t, err)
+
+	cancel() // Stop listening
+	wg.Wait()
+	ec1.check(t, len(sent), 1, 1)
+	ec2.check(t, len(sent), 0, 1)
```
jsign, line 116: Nice!

threaddb/mail.go
```diff
@@ -0,0 +1,74 @@
...
+
+// NewMailbox creates a new threaddb mail box.
+func (m *Mail) NewMailbox(ctx context.Context, name string, opts ...Option) (thread.ID, error) {
+	args := &Options{}
+	for _, opt := range opts {
+		opt(args)
+	}
+	id := thread.NewIDV1(thread.Raw, 32)
+	ctx = common.NewThreadNameContext(ctx, name)
+	err := m.c.NewDB(ctx, id, db.WithNewManagedName(name), db.WithNewManagedCollections(mailConfig), db.WithNewManagedToken(args.Token))
```
jsign, line 72: Q since my memory on this is old:
This `NewMailbox` was used to have idempotency behavior. A priori before the call a `NewIDV1` is generated.
Why `NewDB` call would fail if called more than once? Can't you always have an identical collection in a new thread?

api/users/client/client_test.go
sanderpick: Ah yeah. Good point. For now, this is a pure-hub thing since it's connecting hub users.

api/users/client/client.go
sanderpick: The main reason though was to be able to watch only a particular mailbox. I figured ACL _could_ be a factor too, though I'm not totally sure how... so just opted for more flex here. I could be convinced to consolidate.

mail/local/mailbox.go
sanderpick: Ah, good catch! I'll create an issue.

mail/local/mailbox.go
```diff
@@ -0,0 +1,256 @@
...
+	go func() {
+		defer close(state)
+
+		// Start listening for remote changes
+		events, err := m.clients.Threads.Listen(ctx, boxID, []tc.ListenOption{{
+			Type:       tc.ListenAll,
+			Collection: mail.CollectionName,
+		}})
+		if err != nil {
+			state <- cmd.WatchState{Err: err, Fatal: !cmd.IsConnectionError(err)}
```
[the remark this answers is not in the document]
sanderpick: Yep, _fatal_ as in fatal for that method. That is confusing though. I'll change it to `Aborted` or something

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	ReadAt    time.Time     `json:"read_at,omitempty"`
+}
+
+// Open decrypts the message body with identity.
+func (m Message) Open(ctx context.Context, id thread.Identity) ([]byte, error) {
+	return id.Decrypt(ctx, m.Body)
+}
+
+// Read returns whether or not the message has been read.
+func (m Message) Read() bool {
```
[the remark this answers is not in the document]
sanderpick: Will do

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	return !m.ReadAt.IsZero()
+}
+
+func (m Message) Marshal() ([]byte, error) {
+	return json.Marshal(m)
+}
+
+func (m Message) Unmarshal(data []byte) error {
```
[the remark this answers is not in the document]
sanderpick: Yep, adding comment to explain that. Probably good idea to just change the name of the method to it's not assumed to be the normal JSON unmarshal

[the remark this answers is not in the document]
jsign: Sounds good!

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+func (m Message) Open(ctx context.Context, id thread.Identity) ([]byte, error) {
+	return id.Decrypt(ctx, m.Body)
+}
+
+// Read returns whether or not the message has been read.
+func (m Message) Read() bool {
+	return !m.ReadAt.IsZero()
+}
+
+func (m Message) Marshal() ([]byte, error) {
```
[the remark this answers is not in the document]
sanderpick: I'll just remove this actually... I only added to match `Unmarshal`

threaddb/mail.go
sanderpick: Oh, this is more interceptor magic... making this call here results in a thread being tracked (along with the name) for the owner. Subsequent calls will error with a unique constraint violation. The unique constraint on `mongodb.Thread` is fired before the DB is created in `go-threads`.

jsign: Ah, so uniqueness is done by Hub, not a threadsd.

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+		To:        to,
+		Body:      fromBody,
+		Signature: fromSig,
+		CreatedAt: time.Unix(0, res.CreatedAt),
+	}, nil
+}
+
+// ListInboxMessages lists messages from the inbox.
+// Use options to paginate with seek and limit,
+// and filter by read status.
```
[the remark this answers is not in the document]
sanderpick: Unwrapped :P

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	m.To = to
+	m.Body = body
+	m.Signature = sig
+	m.CreatedAt = time.Unix(0, tm.CreatedAt)
+	m.ReadAt = readAt
+	return nil
+}
+
+// SendMessage sends the message body to a recipient.
+func (c *Client) SendMessage(ctx context.Context, from thread.Identity, to thread.PubKey, body []byte) (msg Message, err error) {
```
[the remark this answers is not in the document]
sanderpick: Nice, I'll create an issue... agree that the client side check would be helpful.

threaddb/mail.go
sanderpick: Yep, uniqueness by name is done by the Hub

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+		readAt = time.Unix(0, tm.ReadAt)
+	}
+	m.ID = tm.ID
+	m.From = from
+	m.To = to
+	m.Body = body
+	m.Signature = sig
+	m.CreatedAt = time.Unix(0, tm.CreatedAt)
+	m.ReadAt = readAt
+	return nil
```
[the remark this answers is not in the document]
sanderpick: Done

api/users/client/client_test.go
```diff
@@ -160,99 +160,334 @@ func TestClient_ListThreads(t *testing.T) {
...
+	ctx = thread.NewTokenContext(ctx, tok)
+
+	inbox, sentbox, err := client.SetupMail(ctx)
+	require.NoError(t, err)
+	assert.NotEmpty(t, inbox)
+	assert.NotEmpty(t, sentbox)
+
+	// should be idempotent
+	_, _, err = client.SetupMail(ctx)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
sanderpick: Nice, done

api/users/client/client_test.go
```diff
@@ -160,99 +160,334 @@ func TestClient_ListThreads(t *testing.T) {
...
+		var all []c.Message
+		var seek string
+		pageSize := 10
+		for { // 10 pages
+			list, err := client.ListInboxMessages(tctx, c.WithLimit(pageSize+1), c.WithSeek(seek))
+			require.NoError(t, err)
+			all = append(all, list[:pageSize]...)
+			if len(list) < pageSize+1 {
+				break
+			}
```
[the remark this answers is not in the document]
sanderpick: Yep, actually I could not figure out a clean way to _not_ include the seek key in the badger results. I had something working where,
- When listing in descending order, find the lexicographically next smallest ulid by altering the last rune to `0`. This only works if the last rune is not `0`!. I'm sure some more complex logic could be devised, but I stopped there and left it to the client to be smart. Basically, we want keys that are _less than_ seek key, not _less than or greater than_. 
- The same but opposite for listing by ascending order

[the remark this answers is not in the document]
sanderpick: [This section](https://github.com/dgraph-io/badger#reverse-iteration-doesnt-give-me-the-right-results) of the badger docs doesn't jive with what I'm seeing... which is that in either order, the seek key is included in the results.

go.mod
```diff
@@ -2,11 +2,15 @@ module github.com/textileio/textile
 
 go 1.14
 
+replace github.com/ipfs/go-datastore v0.4.4 => github.com/textileio/go-datastore v0.4.5-0.20200728205504-ffeb3591b248
+
+replace github.com/ipfs/go-ds-badger v0.2.4 => github.com/textileio/go-ds-badger v0.2.5-0.20200728212847-1ec9ac5e644c
```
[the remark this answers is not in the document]
sanderpick: I will poke at them again and see... if not, can move it! (I already made a comment on a similar (state) PR, no response yet.

api/users/service.go
```diff
@@ -64,3 +74,309 @@ func (s *Service) ListThreads(ctx context.Context, _ *pb.ListThreadsRequest) (*p
...
+		limit = maxMessagePageSize
+	}
+	q.LimitTo(int(limit))
+	switch stat {
+	case 0:
+		break
+	case 1:
+		q.And("read_at").Gt(float64(0))
+	case 2:
+		q.And("read_at").Eq(float64(0))
```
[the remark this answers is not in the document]
sanderpick: Constant instead of `0`?

api/users/client/client_test.go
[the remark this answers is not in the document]
jsign: Ah, I see. Tricky.

go.mod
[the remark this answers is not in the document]
jsign: Yeah, I'd expect for a long time not having much movement in any request. Maybe I can be surprised.

api/users/service.go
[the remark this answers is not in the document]
jsign: Yep, some name.

api/users/service.go
```diff
@@ -64,3 +74,309 @@ func (s *Service) ListThreads(ctx context.Context, _ *pb.ListThreadsRequest) (*p
...
+	dbToken, _ := thread.TokenFromContext(ctx)
+
+	ok, err := user.Key.Verify(req.ToBody, req.ToSignature)
+	if !ok || err != nil {
+		return nil, status.Error(codes.Unauthenticated, "Bad message signature")
+	}
+	ok, err = user.Key.Verify(req.FromBody, req.FromSignature)
+	if !ok || err != nil {
+		return nil, status.Error(codes.Unauthenticated, "Bad message signature")
+	}
```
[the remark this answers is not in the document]
sanderpick: Wouldn't that mean both bodies would be required in each message? The way it is now, the inbox message just has the body for the recipient, and the sentbox message just has the body for the sender. So, verification can happen on each message independent of the other.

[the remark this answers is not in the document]
jsign: Oh, OK yeah makes sense. I was maybe too focused on this particular piece of code validating multiple signatures.

api/users/service.go
```diff
@@ -64,3 +74,309 @@ func (s *Service) ListThreads(ctx context.Context, _ *pb.ListThreadsRequest) (*p
...
+
+	to := &thread.Libp2pPubKey{}
+	if err := to.UnmarshalString(req.To); err != nil {
```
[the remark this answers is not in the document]
sanderpick: Good point, maybe the sigs are bit overkill. I'd opt to leave as is for now... sometimes these things are just "peace of mind" for users, even if a bit redundant... unless it becomes an efficiency bottleneck.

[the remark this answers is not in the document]
jsign: Similar to above comment, yeah, is a lower layer concern.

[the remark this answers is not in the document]
jsign: Agree, shouldn't hurt :)

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	return
+}
+
+// Message is the client side representation of a mailbox message.
+type Message struct {
+	ID        string        `json:"_id"`
+	From      thread.PubKey `json:"from"`
+	To        thread.PubKey `json:"to"`
+	Body      []byte        `json:"body"`
+	Signature []byte        `json:"signature"`
```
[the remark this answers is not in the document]
sanderpick: Done

api/users/client/client.go
```diff
@@ -42,3 +48,227 @@ func (c *Client) GetThread(ctx context.Context, name string) (*pb.GetThreadReply
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return handleMessageList(res)
+}
+
+// ListSentMessages lists messages from the sentbox.
+// Use options to paginate with seek and limit.
+func (c *Client) ListSentMessages(ctx context.Context, opts ...ListOption) ([]Message, error) {
```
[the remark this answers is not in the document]
sanderpick: Yep, can do!

api/users/service.go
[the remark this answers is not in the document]
sanderpick: Cool, done 👍

api/users/service.go
```diff
@@ -64,3 +74,309 @@ func (s *Service) ListThreads(ctx context.Context, _ *pb.ListThreadsRequest) (*p
...
+			return thread.Undef, status.Error(codes.FailedPrecondition, ErrSentboxNotFound.Error())
+		}
+		return thread.Undef, err
+	}
+	return thrd.ID, nil
+}
+
+func (s *Service) getOrCreateMailbox(ctx context.Context, key crypto.PubKey, name string, opts ...tdb.Option) (thread.ID, error) {
+	id, err := s.Mail.NewMailbox(ctx, name, opts...)
+	if err != nil && strings.Contains(err.Error(), mdb.DuplicateErrMsg) {
```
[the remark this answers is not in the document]
sanderpick: Nice, good call... done

api/users/client/client.go
sanderpick: Yep, but they both use the same collection schema. I guess we could do sent and received messages

sanderpick: Let's give this a shot, it's the two threads are cumbersome, we could just add a different collection schema to the inbox and start putting sent messages there

andrewxhill: coming in late here. sorry, just some questions to understand. 

sounds like an inbox cross-application then?

sanderpick: Yep, but sorta indirectly. It's more that each user (which may be part of multiple apps using the same API key) gets one mail setup. If a user is using the same identity across multiple dev's apps, this current setup means that the first dev is the custodian of the mail threads. Not sure if that'll cause problems or not. If so, we could namespace the mail threads by the account key or something

andrewxhill: okay cool, sounds good. let's see!

PR #291, opened 2020-07-29
Account/User FFS Integration

asutula:
Ended up using this branch to focus on integrating FFS instances into the `Account` and `User` models, and updating Bucket archives to use those FFS instances. More work on exposing Powergate functionality, either direct API access or wrapped by Buckets APIs, will come in a later PR.

This PR does a few main things:

- Creates a FFS instance for each new `Account` and `User`
- Detects when a FFS instance should be created lazily because the FFS token is no longer valid or because the `Account` or `User` was created before Bucket archiving was enabled
- If a FFS instance is created during a call to `Archive`, returns a error telling the caller to try again in 30 sec to wait for wallet funds
- Refactors Bucket archiving to use the `Account` or `User` FFS instance


core/core.go
```diff
@@ -75,11 +81,13 @@ var (
...
-	thn   *netclient.Client
-	bucks *bc.Buckets
-	powc  *powc.Client
+	ts             tc.NetBoostrapper
+	th             *threads.Client
+	thn            *netclient.Client
+	bucks          *bc.Buckets
+	powc           *powc.Client
+	powStub        grpcdynamic.Stub
+	ffsServiceDesc *desc.ServiceDescriptor
```
asutula, line 90: The `powStub` and `ffsServiceDesc` are dynamic gRPC representations... they are created at runtime by reading in the descriptor data from the `powergate/ffs` package that is a dependency.

core/core.go
```diff
@@ -154,10 +162,28 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
 		}
+		t.powc, err = powc.NewClient(target)
+		if err != nil {
+			return nil, err
+		}
+		pgConn, err := grpc.Dial(target, grpc.WithInsecure())
+		if err != nil {
+			return nil, err
+		}
+		t.powStub = grpcdynamic.NewStub(pgConn)
```
asutula, line 177: We create the dynamic stub just by passing in the powergate connection.

core/core.go
```diff
@@ -154,10 +162,28 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
+		ffsDesc, err := desc.LoadFileDescriptor("ffs/rpc/rpc.proto")
+		if err != nil {
+			return nil, err
+		}
+		ffsServiceDesc := ffsDesc.FindService("ffs.rpc.RPCService")
+		if ffsServiceDesc == nil {
+			return nil, fmt.Errorf("no ffs service description found")
+		}
+		t.ffsServiceDesc = ffsServiceDesc
```
asutula, line 186: We create the dynamic FFS service description by loading the file descriptor data from the runtime (`ffs/rpc/rpc.proto` is not a path to a file, but a key that the ffs file descriptor is registered under), then looking up the service we want (`ffs.rpc.RPCService`)

asutula, line 186: We create the dynamic FFS service description by loading the file descriptor data from the runtime (`ffs/rpc/rpc.proto` is not a path to a file, but a key that the ffs file descriptor is registered under), then looking up the service we want (`ffs.rpc.RPCService`)

core/core.go
```diff
@@ -261,15 +287,17 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
-	ptarget, err := tutil.TCPAddrFromMultiAddr(conf.AddrAPIProxy)
+	ptarget, err := TCPAddrFromMultiAddr(conf.AddrAPIProxy)
 	if err != nil {
 		return nil, err
 	}
 	var opts []grpc.ServerOption
 	if conf.Hub {
 		opts = []grpc.ServerOption{
-			grpcm.WithUnaryServerChain(auth.UnaryServerInterceptor(t.authFunc), t.threadInterceptor()),
+			grpcm.WithUnaryServerChain(auth.UnaryServerInterceptor(t.authFunc), t.powergateInterceptor, t.threadInterceptor()),
```
asutula, line 300: Add another interceptor here to potentially handle Powergate calls.

core/core.go
```diff
@@ -291,6 +319,7 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 			upb.RegisterAPIServer(t.server, us)
 		}
 		bpb.RegisterAPIServer(t.server, bs)
+		ffsRpc.RegisterRPCServiceServer(t.server, &ffsService)
```
asutula, line 322: There is some logic in gRPC that runs before the interceptors that determines if the request is supported by the server. You have to register _some_ FFS service if you want to be able to deal with FFS requests in the interceptor layer, so I just register the generated `UnimplementedRPCServiceServer` and we don't have to create our own stub in any way

core/core.go
```diff
@@ -538,6 +567,22 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
+func (t *Textile) powergateInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	if !strings.Contains(info.FullMethod, "ffs.rpc.RPCService") {
```
asutula, line 571: Ignore non-FFS calls.

core/core.go
```diff
@@ -538,6 +567,22 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+		return handler(ctx, req)
+	}
+	methodParts := strings.Split(info.FullMethod, "/")
+	if len(methodParts) < 2 {
+		return nil, fmt.Errorf("error parsing method string %s", info.FullMethod)
+	}
+	methodName := methodParts[len(methodParts)-1]
+	methodDesc := t.ffsServiceDesc.FindMethodByName(methodName)
```
asutula, line 579: Get the method name from the incoming call info and look up the method descriptor in our FFS service descriptor.

core/core.go
```diff
@@ -538,6 +567,22 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+	if methodDesc == nil {
+		return nil, fmt.Errorf("no method found for %s", methodName)
+	}
+	return t.powStub.InvokeRpc(ctx, methodDesc, req.(proto.Message))
```
asutula, line 583: Use our stub to invoke the method on the remote powergate server. Boom, dynamic gRPC.

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+	var isAccount bool
+	if acct := accountFromContext(ctx); acct != nil {
+		ffsInfo = acct.FFSInfo
+		owner = acct.Key
+		isAccount = true
+	} else if user := userFromContext(ctx); user != nil {
+		ffsInfo = user.FFSInfo
+		owner = user.Key
+		isAccount = false
+	}
```
asutula, line 1621: Bucket archiving now get the FFS instance via the `Account` or `User`. Building up the information we need here.

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+
+	createNewFFS := func() error {
```
asutula, line 1623: Little reusable func that is called to create a new FFS instance and update the `Account` or `User` database record.

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+		} else {
+			_, err = s.Collections.Users.UpdateFFSInfo(ctx, owner, &mdb.FFSInfo{ID: id, Token: token})
+		}
+		if err != nil {
+			return fmt.Errorf("updating user/account with new ffs information: %v", err)
+		}
+		return nil
+	}
+
+	tryAgain := fmt.Errorf("new ffs instance created, please try again in 30 seconds to allow time for wallet funding")
```
asutula, line 1639: Reusable error that is returned is a new FFS instance is created.

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+
+	// case where account/user was created before bucket archives were enabled.
+	// create a ffs instance for them.
+	if ffsInfo == nil {
+		if err := createNewFFS(); err != nil {
+			return nil, err
+		}
+		return nil, tryAgain
```
asutula, line 1647: If we don't find any FFS info for the caller, this means their record was created before this code was deployed, so we should lazily create the FFS instance for them.

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+	if err != nil {
+		if err.Error() != "auth token not found" {
+			return nil, fmt.Errorf("getting ffs default StorageConfig: %v", err)
+		} else {
+			// case where the ffs token is no longer valid because powergate was reset.
+			// create a new ffs instance for them.
+			if err := createNewFFS(); err != nil {
+				return nil, err
+			}
+			return nil, tryAgain
```
asutula, line 1662: If we did find FFS info for the caller, but our call to FFS returned "no auth token found", that implies that Powergate was reset, so we should create a new FFS instance for the caller.

api/buckets/service.go
```diff
@@ -1671,19 +1726,19 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
 		JobID:     jid.String(),
 		JobStatus: int(ffs.Queued),
 	}
-	if err := s.Collections.FFSInstances.Replace(ctx, ffsi); err != nil {
+	if err := s.Collections.BucketArchives.Replace(ctx, ba); err != nil {
 		return nil, fmt.Errorf("updating ffs instance data: %s", err)
 	}
 
-	if err := s.ArchiveTracker.Track(ctx, dbID, dbToken, req.GetKey(), jid, p.Cid()); err != nil {
+	if err := s.ArchiveTracker.Track(ctx, dbID, dbToken, req.GetKey(), jid, p.Cid(), owner); err != nil {
```
asutula, line 1741: Need to pass the owner PK in as part of the saved archive info so that we call pull the owner and ffs instance info out later when we are monitoring the archive

api/buckets/service.go
```diff
@@ -1698,12 +1753,23 @@ func (s *Service) ArchiveWatch(req *pb.ArchiveWatchRequest, server pb.API_Archiv
...
+	var ffsInfo *mdb.FFSInfo
+	if acct := accountFromContext(server.Context()); acct != nil {
+		ffsInfo = acct.FFSInfo
+	} else if user := userFromContext(server.Context()); user != nil {
+		ffsInfo = user.FFSInfo
+	}
+
+	if ffsInfo == nil {
+		return fmt.Errorf("no account or no FFS info associated with account")
+	}
```
asutula, line 1765: In `ArchiveWatch` we get the ffs info from the `Account` or `User` now.

asutula: If there is no FFS info, just return that error. FFS instance is only lazily created on a call to `Archive`

api/buckets/service.go
```diff
@@ -1698,12 +1753,23 @@ func (s *Service) ArchiveWatch(req *pb.ArchiveWatchRequest, server pb.API_Archiv
...
+
 	var err error
 	ctx, cancel := context.WithCancel(server.Context())
 	defer cancel()
 	ch := make(chan string)
 	go func() {
-		err = s.Buckets.ArchiveWatch(ctx, req.GetKey(), ch)
+		err = s.Buckets.ArchiveWatch(ctx, req.GetKey(), ffsInfo.Token, ch)
```
asutula, line 1772: Making sure this is correct... in this case, if the ffs token is no longer valid, we will get the "auth token not found" error immediately and just return it a few lines below. That is fine, we don't want to do any FFS creation here.

api/hub/service.go
```diff
@@ -70,7 +72,16 @@ func (s *Service) Signup(ctx context.Context, req *pb.SignupRequest) (*pb.Signup
...
+	var ffsInfo *mdb.FFSInfo
+	if s.Pow != nil {
+		ffsId, ffsToken, err := s.Pow.FFS.Create(ctx)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Unable to create FFS instance: %v", err)
+		}
+		ffsInfo = &mdb.FFSInfo{ID: ffsId, Token: ffsToken}
+	}
+
+	dev, err := s.Collections.Accounts.CreateDev(ctx, req.Username, req.Email, ffsInfo)
```
asutula, line 84: If Powergate integration is enabled, create a FFS instance for the dev

api/hub/service.go
```diff
@@ -283,11 +294,20 @@ func (s *Service) CreateOrg(ctx context.Context, req *pb.CreateOrgRequest) (*pb.
...
+			return nil, status.Errorf(codes.Internal, "Unable to create FFS instance: %v", err)
+		}
+		ffsInfo = &mdb.FFSInfo{ID: ffsId, Token: ffsToken}
+	}
 	org, err := s.Collections.Accounts.CreateOrg(ctx, req.Name, []mdb.Member{{
 		Key:      dev.Key,
 		Username: dev.Username,
 		Role:     mdb.OrgOwner,
-	}})
+	}}, ffsInfo)
```
asutula, line 310: If Powergate integration is enabled, create a FFS instance for the org

buckets/archive/tracker.go
```diff
@@ -88,7 +89,20 @@ func (t *Tracker) run() {
...
 						ctx, cancel := context.WithTimeout(t.ctx, time.Second*5)
 						defer cancel()
-						reschedule, cause, err := t.trackArchiveProgress(ctx, a.BucketKey, a.DbID, a.DbToken, a.JID, a.BucketRoot)
+
+						var ffsInfo *mdb.FFSInfo
+						if account, err := t.colls.Accounts.Get(ctx, a.Owner); err == nil {
+							ffsInfo = account.FFSInfo
+						} else if user, err := t.colls.Users.Get(ctx, a.Owner); err == nil {
+							ffsInfo = user.FFSInfo
+						}
```
asutula, line 98: Here we are in the tracker where we pull out the newly persisted `Owner` so we can look up the `Account` or `User` to get their FFS info

buckets/archive/tracker.go
```diff
@@ -123,20 +137,21 @@ func (t *Tracker) Track(ctx context.Context, dbID thread.ID, dbToken thread.Toke
...
 	defer cancel()
-	ctx = context.WithValue(ctx, powc.AuthKey, ffsi.FFSToken)
+	ctx = context.WithValue(ctx, powc.AuthKey, ffsInfo.Token)
 	ch := make(chan powc.JobEvent, 1)
 	if err := t.pgClient.FFS.WatchJobs(ctx, ch, jid); err != nil {
+		// if error specifies that the auth token isn't found, powergate must have been reset.
+		// return the error as fatal so the archive will be untracked
+		if err.Error() == "auth token not found" {
+			return false, "", err
+		}
```
asutula, line 154: Important case here

core/core.go
```diff
@@ -159,8 +156,7 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 		return nil, err
 	}
 	if conf.AddrPowergateAPI != "" {
-		t.powc, err = powc.NewClient(conf.AddrPowergateAPI, grpc.WithInsecure(), grpc.WithPerRPCCredentials(powc.TokenAuth{}))
```
asutula, line 162: These options are the defaults now

core/core.go
```diff
@@ -653,7 +650,15 @@ func (t *Textile) threadInterceptor() grpc.UnaryServerInterceptor {
...
-			if err := t.collections.Users.Create(ctx, owner); err != nil {
+			var ffsInfo *mdb.FFSInfo
+			if t.powc != nil {
+				ffsId, ffsToken, err := t.powc.FFS.Create(ctx)
+				if err != nil {
+					return nil, err
+				}
+				ffsInfo = &mdb.FFSInfo{ID: ffsId, Token: ffsToken}
+			}
+			if err := t.collections.Users.Create(ctx, owner, ffsInfo); err != nil {
```
asutula, line 661: Create an FFS instance for new `User`s

core/core.go
```diff
@@ -210,7 +206,7 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 	if err != nil {
 		return nil, err
 	}
-	t.bucks, err = tdb.NewBuckets(t.th, t.powc, t.collections.FFSInstances, conf.FFSDefaultConfig)
```
asutula, line 213: Bucket archives just use the FFS instance default `StorageConfig`. We may want to provide an api to control that later

mongodb/accounts.go
```diff
@@ -112,7 +113,7 @@ func NewAccounts(ctx context.Context, db *mongo.Database) (*Accounts, error) {
 	return a, err
 }
 
-func (a *Accounts) CreateDev(ctx context.Context, username, email string) (*Account, error) {
+func (a *Accounts) CreateDev(ctx context.Context, username, email string, ffsInfo *FFSInfo) (*Account, error) {
```
asutula, line 116: Updated to persist `FFSInfo`

mongodb/accounts.go
```diff
@@ -136,21 +138,23 @@ func (a *Accounts) CreateDev(ctx context.Context, username, email string) (*Acco
...
+	}
+	encodeFFSInfo(data, doc.FFSInfo)
+	if _, err := a.col.InsertOne(ctx, data); err != nil {
 		return nil, err
 	}
 	return doc, nil
 }
 
-func (a *Accounts) CreateOrg(ctx context.Context, name string, members []Member) (*Account, error) {
+func (a *Accounts) CreateOrg(ctx context.Context, name string, members []Member, ffsInfo *FFSInfo) (*Account, error) {
```
asutula, line 157: Updated to persist `FFSInfo`

mongodb/accounts.go
```diff
@@ -198,20 +203,43 @@ func (a *Accounts) CreateOrg(ctx context.Context, name string, members []Member)
...
-	}); err != nil {
+	}
+	encodeFFSInfo(data, doc.FFSInfo)
+	if _, err = a.col.InsertOne(ctx, data); err != nil {
 		return nil, err
 	}
 	return doc, nil
 }
 
+func (a *Accounts) UpdateFFSInfo(ctx context.Context, key crypto.PubKey, ffsInfo *FFSInfo) (*Account, error) {
```
asutula, line 222: New method to update the `FFSInfo` of an `Account`

mongodb/archivetracking.go
```diff
@@ -32,6 +34,7 @@ type trackedArchive struct {
 	DbToken    thread.Token `bson:"db_token"`
 	BucketKey  string       `bson:"bucket_key"`
 	BucketRoot []byte       `bson:"bucket_root"`
+	Owner      []byte       `bson:"owner"`
```
asutula, line 37: New property to track the archive owner so we can get their ffs info while running the tracker

mongodb/bucketarchives.go
```diff
@@ -7,11 +7,9 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 )
 
-type FFSInstance struct {
-	BucketKey  string   `bson:"_id"`
-	FFSToken   string   `bson:"ffs_token"`
-	WalletAddr string   `bson:"ffs_walletaddr"`
-	Archives   Archives `bson:"archives"`
+type BucketArchive struct {
```
asutula, line 10: Renamed this since it has nothing to do with FFS instances now.

mongodb/common.go
```diff
@@ -0,0 +1,41 @@
...
+		raw := v.(bson.M)
+		if v, ok := raw[idKey]; ok {
+			ffsInfo.ID = v.(string)
+		}
+		if v, ok := raw[tokenKey]; ok {
+			ffsInfo.Token = v.(string)
+		}
+	}
+	return ffsInfo
+}
```
asutula, line 41: Some utilities for dealing with `FFSInfo`

mongodb/users.go
```diff
@@ -34,20 +35,23 @@ func NewUsers(_ context.Context, db *mongo.Database) (*Users, error) {
 	return &Users{col: db.Collection("users")}, nil
 }
 
-func (u *Users) Create(ctx context.Context, key crypto.PubKey) error {
+func (u *Users) Create(ctx context.Context, key crypto.PubKey, ffsInfo *FFSInfo) error {
```
asutula, line 38: Takes in `FFSInfo` now

mongodb/users.go
```diff
@@ -56,6 +60,27 @@ func (u *Users) Create(ctx context.Context, key crypto.PubKey) error {
 	return nil
 }
 
+func (u *Users) UpdateFFSInfo(ctx context.Context, key crypto.PubKey, ffsInfo *FFSInfo) (*User, error) {
```
asutula, line 63: New method to update `User` `FFSInfo`

threaddb/buckets.go
```diff
@@ -204,39 +175,6 @@ func (b *Buckets) IsArchivingEnabled() bool {
 	return b.pgClient != nil
 }
 
-func (b *Buckets) createFFSInstance(ctx context.Context, bucketKey string) error {
```
asutula, line 207: No longer creating FFS instances with new Buckets

api/buckets/service.go
jsign: Nit: Shouldn't happen, but maybe would be safer to add an `else` and fail fast?

api/buckets/service.go
```diff
@@ -1606,33 +1607,87 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+		}
+		return nil, tryAgain
 	}
 
-	ctxFFS := context.WithValue(ctx, powc.AuthKey, ffsi.FFSToken)
+	ctxFFS := context.WithValue(ctx, powc.AuthKey, ffsInfo.Token)
+
+	defConf, err := s.PGClient.FFS.DefaultStorageConfig(ctxFFS)
+	if err != nil {
+		if err.Error() != "auth token not found" {
```
jsign, line 1654: Future idea: Maybe create an `ErrTokenNotFound` so at least can be stringified using the client library of powergate? At least if we change the msg **and** client is kept up to date with powd, we shouldn't need to remember changing this string. 

(General idea though, not really specific to this case)

api/buckets/service.go
```diff
@@ -1698,12 +1753,23 @@ func (s *Service) ArchiveWatch(req *pb.ArchiveWatchRequest, server pb.API_Archiv
...
 
+	var ffsInfo *mdb.FFSInfo
+	if acct := accountFromContext(server.Context()); acct != nil {
+		ffsInfo = acct.FFSInfo
+	} else if user := userFromContext(server.Context()); user != nil {
+		ffsInfo = user.FFSInfo
+	}
+
+	if ffsInfo == nil {
+		return fmt.Errorf("no account or no FFS info associated with account")
```
jsign, line 1764: Nit: Maybe `no user/account or no FFS info associated with user/account`.

buckets/archive/tracker.go
```diff
@@ -88,7 +89,20 @@ func (t *Tracker) run() {
...
+						if account, err := t.colls.Accounts.Get(ctx, a.Owner); err == nil {
+							ffsInfo = account.FFSInfo
+						} else if user, err := t.colls.Users.Get(ctx, a.Owner); err == nil {
+							ffsInfo = user.FFSInfo
+						}
+
+						if ffsInfo == nil {
+							log.Errorf("no FFSInfo, could not find an Account or User: %s", err)
+							return
+						}
```
jsign, line 103: I think here we should call `.Finalize` since looks like this will fail forever.
e.g:
```
if err := t.colls.ArchiveTracking.Finalize(ctx, a.JID, "huh? no FFS info, weird"); err != nil {
        log.Errorf("finalizing errored/rescheduled archive tracking: %s", err)
}
```

core/core.go
jsign: Uhm, OK! I think the default is `repFactor: 1` which is just what we need for new nerpa since there aren't many miners. Before was 10 because `testnet` had a bunch.

A sensible default is a pain... if too high will fail to find enough miners. If too low, might choose "the highest reputation miner" but maybe isn't the best, so would fail.

Definitely allowing an API to customize which config sounds good. At least to say: "hey, the default isn't really helping now. Set a storageconfig that forces miner X".

Sharing thoughts.

core/core.go
jsign: Always forget which case was covering. It's the first gRPC call that some new user does, right?

integrationtest/pg/pg_test.go
```diff
@@ -28,7 +27,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-var powMultiaddr = "127.0.0.1:5002"
+var powAddr = "127.0.0.1:5002"
```
jsign, line 30: 👍🏼

api/buckets/service.go
asutula: Think you probably realized this below, but we don't want to fail. In the case of no ffs info, we want to create a ffs instance.

asutula: ah, wait. I see. the else case would be "no user found". that would be an error. will add.

api/buckets/service.go
asutula: Yea I was thinking we need similar improvements in general. Been looking at gRPC's support for custom error metadata. Handy package [here](https://godoc.org/google.golang.org/genproto/googleapis/rpc/errdetails)

core/core.go
asutula: Agreed. I see plenty of FFS-related api enhancements to the Hub/Buckets API in our future, so we'll definitely have time to address this.

core/core.go
asutula: That is my understanding, yes. This interceptor always checks for an existing user. So if that user isn't found, it's the first time we've seen them, so we create the FFS instance and user record.

mongodb/common.go
```diff
@@ -0,0 +1,41 @@
...
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+type FFSInfo struct {
+	ID    string
+	Token string
+}
+
+const (
+	key      = "ffs_info"
```
sanderpick, line 14: This would require a migration if we wanted to move away from `FFS` terminology

asutula: good call, forgot about that. will change

PR #301, opened 2020-07-30
pg: archive tracker component

### LINKED ISSUE #282: pg: make archive tracking a first-class citizen
Currently, after an archive is fired, a go-routine will track the status change of its corresponding Powergate job.

This isn't very convenient for some reasons. The main one is that the goroutine lifecycle will forcibly end if the Hub is shutdown, and it will switch the tracking to a defined `aborted` status, which basically means: _I'm not sure how this job ended. Doesn't mean it fails nor succeeds._. The same happens if the connection with Powergate fails; will be finished with `aborted`.

This can lead to the archive succeeding on the Powergate side, but left "untracked" in the Hub. This means that making `hub buck archive info` keep saying that "The bucket was never archived.

This was originally by design as a first version, but we should have a better way of managing this tracking. Most probably some extra component should be responsible for tracking all archives and have some persistent information about which ones were stopped, resume them, etc.

jsign:
Closes #282, which already has an explanation of the intention of the change.

The gateway token was promoted to a more general concept. It is (and will be) used as an internal session to bypass auth interceptor in any internal logic of the Hub. This avoids the need of having to serialize auth information from dev/org/user and recover it out-of-request to perform actions on threads on its behalf.

### DIFF
--- a/buckets/archive/tracker.go
+++ b/buckets/archive/tracker.go
@@ -149,9 +149,7 @@ func (t *Tracker) trackArchiveProgress(ctx context.Context, buckKey string, dbID
 		return true, "watching cancelled", nil
 	case s, ok := <-ch:
 		if !ok {
-			log.Errorf("getting job %s status stream closed", jid)
-			aborted = true
-			abortMsg = "server closed communication channel"
+			return true, "powergate closed communication chan", nil
 		}
 		if s.Err != nil {
 			log.Errorf("job %s update: %s", jid, s.Err)
@@ -176,7 +174,12 @@ func (t *Tracker) trackArchiveProgress(ctx context.Context, buckKey string, dbID
 		return true, fmt.Sprintf("updating archive status: %s", err), nil
 	}
 
-	return false, "reached final status", nil
+	msg := "reached final status"
+	if aborted {
+		msg = "aborted with reason " + abortMsg
+	}
+
+	return false, msg, nil
 }
 
 // updateArchiveStatus save the last known job status. It also receives an
--- a/buckets/archive/tracker.go
+++ b/buckets/archive/tracker.go
@@ -1,0 +1,263 @@
+package archive
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	logger "github.com/ipfs/go-log"
+	"github.com/textileio/go-threads/core/thread"
+	powc "github.com/textileio/powergate/api/client"
+	"github.com/textileio/powergate/ffs"
+	"github.com/textileio/textile/api/common"
+	mdb "github.com/textileio/textile/mongodb"
+	tdb "github.com/textileio/textile/threaddb"
+)
+
+const (
+	maxConcurrent = 20
+)
+
+var (
+	CheckInterval         = time.Second * 15
+	JobStatusPollInterval = time.Second * 30
+
+	log = logger.Logger("pow-archive")
+)
+
+type Tracker struct {
+	lock   sync.Mutex
+	ctx    context.Context
+	cancel context.CancelFunc
+	closed chan (struct{})
+
+	internalSession string
+	colls           *mdb.Collections
+	buckets         *tdb.Buckets
+	pgClient        *powc.Client
+}
+
+func New(colls *mdb.Collections, buckets *tdb.Buckets, pgClient *powc.Client, internalSession string) (*Tracker, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	t := &Tracker{
+		ctx:    ctx,
+		cancel: cancel,
+		closed: make(chan struct{}),
+
+		internalSession: internalSession,
+		colls:           colls,
+		buckets:         buckets,
+		pgClient:        pgClient,
+	}
+	go t.run()
+	return t, nil
+}
+
+func (t *Tracker) Close() error {
+	t.cancel()
+	<-t.closed
+	return nil
+}
+
+func (t *Tracker) run() {
+	defer close(t.closed)
+	for {
+		select {
+		case <-t.ctx.Done():
+			log.Info("shutting down archive tracker daemon")
+			return
+		case <-time.After(CheckInterval):
+			for {
+				archives, err := t.colls.ArchiveTracking.GetReadyToCheck(t.ctx, maxConcurrent)
+				if err != nil {
+					log.Errorf("getting tracked archives: %s", err)
+					break
+				}
+				log.Infof("get %d ready archive tracking to be processed", len(archives))
+				if len(archives) == 0 {
+					break
+				}
+				var wg sync.WaitGroup
+				wg.Add(len(archives))
+				for _, a := range archives {
+					go func(a *mdb.TrackedArchive) {
+						defer wg.Done()
+
+						ctx, cancel := context.WithTimeout(t.ctx, time.Second*5)
+						defer cancel()
+						reschedule, cause, err := t.trackArchiveProgress(ctx, a.BucketKey, a.DbID, a.DbToken, a.JID, a.BucketRoot)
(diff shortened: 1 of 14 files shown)


buckets/archive/tracker.go
```diff
@@ -0,0 +1,265 @@
...
+	var aborted bool
+	var abortMsg string
+	var job ffs.Job
+	select {
+	case <-ctx.Done():
+		log.Infof("job %s status watching canceled", jid)
+		return true, "watching cancelled", nil
+	case s, ok := <-ch:
+		if !ok {
+			return true, "powergate closed communication chan", nil
```
jsign, line 152: Unexpected channel close? Better retry.

buckets/archive/tracker.go
```diff
@@ -0,0 +1,265 @@
...
+	if err := t.updateArchiveStatus(ctx, ffsi, job, aborted, abortMsg); err != nil {
+		return true, fmt.Sprintf("updating archive status: %s", err), nil
+	}
+
+	msg := "reached final status"
+	if aborted {
+		msg = "aborted with reason " + abortMsg
+	}
+
+	return false, msg, nil
```
jsign, line 182: Say to _not_ reschedule because reached final status (being good or bad, nothing will change that by retrying).

buckets/archive/tracker.go
```diff
@@ -0,0 +1,265 @@
...
+// updateArchiveStatus save the last known job status. It also receives an
+// _abort_ flag which indicates that the provided newJobStatus might not be
+// final. For example, if tracking the Job status changes errored by some network
+// condition, we have only the last known Job status. Powergate would continue doing
+// its work until a final status is reached; it only means we aren't sure how this
+// archive really finished.
+// To track for this situation, we use the _aborted_ and _abortMsg_ parameters.
+// An archive with _aborted_ true should eventually be re-queried to understand
+// how it finished (if wanted).
+func (t *Tracker) updateArchiveStatus(ctx context.Context, ffsi *mdb.FFSInstance, job ffs.Job, aborted bool, abortMsg string) error {
```
jsign, line 194: Nothing new in this method.

buckets/archive/tracker.go
```diff
@@ -0,0 +1,265 @@
...
+	lastArchive.FailureMsg = prepareFailureMsg(job)
+	if err := t.colls.FFSInstances.Replace(ctx, ffsi); err != nil {
+		return fmt.Errorf("updating ffs status update instance data: %s", err)
+	}
+	return nil
+}
+
+func (t *Tracker) saveDealsInArchive(ctx context.Context, key string, dbID thread.ID, dbToken thread.Token, ffsToken string, c cid.Cid) error {
+	opts := tdb.WithToken(dbToken)
+	ctx = common.NewSessionContext(ctx, t.internalSession)
```
jsign, line 218: Using the renamed `gatewaySession` (now `internalSession`) to bypass auth whenever the Threads API call gets re-intercepted again.

Fields `dbID` and `dbToken` were persisted when the archive was registered with `.Track(..)`.

buckets/local/options.go
```diff
@@ -2,14 +2,14 @@ package local
...
 )
 
 type newOptions struct {
 	name    string
 	private bool
-	thread  thread.ID
-	key     string
+	// (jsign): unused?
+	//thread  thread.ID
+	//key     string
```
jsign, line 12: @sanderpick , got some lint about unused fields. 
What's your opinion to do here? (left this code now to remember to comment in the review)

buckets/local/watch.go
```diff
@@ -125,27 +125,25 @@ func (b *Bucket) watchWhileConnected(ctx context.Context, pevents chan<- PathEve
 }
 
 func (b *Bucket) watchPush(ctx context.Context, events chan<- PathEvent) error {
-	select {
```
jsign, line 128: Nit: `select` isn't needed.

core/core.go
```diff
@@ -573,7 +589,6 @@ func (t *Textile) threadInterceptor() grpc.UnaryServerInterceptor {
 		} else if user, ok := mdb.UserFromContext(ctx); ok {
 			owner = user.Key
 		}
-		key, _ := mdb.APIKeyFromContext(ctx)
```
jsign, line 576: Moved closer to usage since only makes sense in a very narrow branch.

integrationtest/pg/pg_test.go
```diff
@@ -22,15 +23,24 @@ import (
...
+	"github.com/textileio/textile/core"
 	"github.com/textileio/textile/util"
 	"google.golang.org/grpc"
 )
 
 var powMultiaddr = multiaddr.StringCast("/ip4/127.0.0.1/tcp/5002")
 
+func TestMain(m *testing.M) {
+	archive.CheckInterval = time.Second * 5
+	archive.JobStatusPollInterval = time.Second * 5
```
jsign, line 36: Make the tracker act faster in the test environment.

integrationtest/pg/pg_test.go
```diff
@@ -44,9 +54,60 @@ func TestCreateBucket(t *testing.T) {
 	require.Equal(t, 1, len(lst))
 }
 
+func TestArchiveTracker(t *testing.T) {
```
jsign, line 57: The idea here is simple.

Spin things as usual, fire an archive, stop the Hub completely, re-spin up again, and continue to watch if the original fired archive is resumed and finalizes correctly.

This simulates the Hub being closed/crashing, and all being good after it starts again resuming things.

mongodb/archivetracking.go
```diff
@@ -0,0 +1,165 @@
+package mongodb
```
jsign, line 1: Mongo collection used by the tracker to know whats pending or finalizing things.

mongodb/archivetracking.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return fmt.Errorf("inserting in collection: %s", err)
+	}
+	return nil
+}
+
+func (at *ArchiveTracking) GetReadyToCheck(ctx context.Context, n int64) ([]*TrackedArchive, error) {
+	opts := options.Find()
+	opts.SetLimit(n)
+	filter := bson.M{"ready_at": bson.M{"$lte": time.Now()}, "active": true}
```
jsign, line 72: We might discover that eventually some index might worth the cost.

mongodb/archivetracking.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, fmt.Errorf("getting tracked archive: %s", res.Err())
+	}
+	var ta trackedArchive
+	if err := res.Decode(&ta); err != nil {
+		return nil, err
+	}
+	return cast(&ta)
+}
+
+func (at *ArchiveTracking) Finalize(ctx context.Context, jid ffs.JobID, cause string) error {
```
sanderpick, line 105: Do we ever delete the archives tracking objects in mongo? Maybe they should be deleted when the bucket gets deleted?

buckets/local/options.go
jsign: Cool, will delete then. 👍

mongodb/archivetracking.go
jsign: Oh, yeah forgot to mention that. Yeah, now all history gets preserved mostly to keep as much info as possible since all these features are quite new. 
 
What I had in mind was some background job that prunes all `active = false` older than X.  I didn't think about deleting them on bucket deletion, most prob because I'm tilted towards saving as much info as possible in the initial phase.

For sure we won't let this be a problem since is an easy thing to delete. I created #303 to keep track with some of the discussed ideas.

jsign: Oops, cliked too fast. What do you think @sanderpick ?

sanderpick: Nice, SGTM!

PR #304, opened 2020-07-30
Mail: use a single thread per user

sanderpick:
This changes the internal mailbox handling to use a single thread per user instead of one for an inbox and one for a sentbox.

### DIFF
--- a/threaddb/mail.go
+++ b/threaddb/mail.go
@@ -5,19 +5,18 @@ import (
 	"errors"
 	"strings"
 
-	mdb "github.com/textileio/textile/mongodb"
-
 	"github.com/alecthomas/jsonschema"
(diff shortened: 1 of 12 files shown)


threaddb/mail.go
```diff
@@ -43,40 +51,68 @@ type Message struct {
 	ReadAt    int64  `json:"read_at"`
 }
 
+// SentboxMessage represents the sentbox threaddb collection schema.
+type SentboxMessage struct {
```
sanderpick, line 55: Only diff here is that there's no `ReadAt`, but I'd imagine these could diverge more in the future, so opted to go the more verbose path and have two. The client-side still just uses a single model.

sanderpick:
Yep! Thanks for the quick 👀

PR #310, opened 2020-08-11
updates to latest Powergate

jsign:
Updates to the latest Powergate version, which had breaking APIs and uses the latest calibration network code for localnet. 

I also included the `FlakyT`, and already used it for PG integration testing which are heavy and might be flaky. So if somehow we identify other ones to leverage, is ready to use.


integrationtest/pg/pg_test.go
```diff
@@ -55,167 +54,175 @@ func TestCreateBucket(t *testing.T) {
 }
 
 func TestArchiveTracker(t *testing.T) {
-	_ = spinup(t)
-	ctx, conf, client, shutdown := setup(t)
+	util.RunFlaky(t, func(t *util.FlakyT) {
```
sanderpick, line 57: Indents from copy+paste? There's a bunch in this file.

jsign: Oh, diff looks terrible. Nothing about the test changed. Yep, the indentation bumped since test logic is wrapped in `util.RunFlaky`.

sanderpick: Ahhhh, I see

PR #316, opened 2020-08-12
File-level bucket access control

### LINKED ISSUE #313: Add a buckets validator that implements a file-level ACL
Add bucket structure to the buckets model to allow file-level access and leveraging the `go-threads` [collection validation functions](https://github.com/textileio/go-threads/issues/418).

### LINKED ISSUE #314: Allow users to access other users' threads
Update hub to let threads handle thread access (users should be able to access others’ threads)

### LINKED ISSUE #318: Local random key encryption
Local (client-side) encryption is possible using a password, but it would be nice to also expose random key encryption for cases where a human-readable password is not needed. Password-based encryption is slower since the keys need to be derived from the password.

sanderpick:
Closes #313 
Closes #314 
Closes #318 

#### Versioning
Bucket now have versions. All original buckets are version `0`. New buckets are version `1`.

#### Bucket file collaboration
File-level access control is available to `v1` buckets, which is enabled by the new `WriteValidator` and `ReadFilter` JavaScript [collection functions](https://github.com/textileio/go-threads/pull/416).

The buckets API has two new methods for manipulating file access roles:
- `PushPathAccessRoles` (available on the gRPC client and `local` library)
- `PullPathAccessRoles` (available on the gRPC client and `local` library)

`local` gets a helper method for sending path invites, `SendPathInvite`.

### Hub
The gRPC API and client have a new method, `GetIdentity`. This can be used by accounts that do not know their identity (private key), but wish to use the Mailbox APIs.


buckets/local/access.go
```diff
@@ -0,0 +1,59 @@
...
+
+// PathInvite wraps information needed to collaborate on a bucket path.
+type PathInvite struct {
+	Key  string `json:"key"`
+	Path string `json:"path"`
+}
+
+// SendPathInvite sends a message containing a bucket key and path,
+// which can be used to access a shared file / folder.
+func (b *Bucket) SendPathInvite(ctx context.Context, from thread.Identity, to thread.PubKey, pth string) error {
```
jsonsivar, line 46: For this we were thinking of sending 1 notification per recipient so it would have a list of paths `pths []string`. We also want to customize the body to include some metadata so wondering if we have 1 of the 2 options it would work:
1. Ability to override the body for this message being sent
2. Turn of messaging during the sharing request then we can send the mailbox item directly

sanderpick: This is really just a helper function. I decided to not make a single method that edits the access roles and then sends an "invite" to all the new identities because it felt too app specific. For your use case, you can just not use this method and send whatever body message you need, possible with something like this `PathInvite` object. In the end it will be up to your app to handle the invite anyway.

sanderpick:
Tests passing after a bit of whack a mole. Definitely need more tests here, but the basics are in place finally.

core/core.go
```diff
@@ -108,6 +107,26 @@ var (
...
+		"/threads.pb.API/WriteTransaction",
+		"/threads.pb.API/Listen",
+		"/api.buckets.pb.APIService/Root",
+		"/api.buckets.pb.APIService/Links",
+		"/api.buckets.pb.APIService/ListPath",
+		"/api.buckets.pb.APIService/PushPath",
+		"/api.buckets.pb.APIService/PullPath",
+		"/api.buckets.pb.APIService/SetPath",
+		"/api.buckets.pb.APIService/RemovePath",
+		"/api.buckets.pb.APIService/GetPathAccessRoles",
```
jsonsivar, line 126: Should this and the one below be `PullPathAccessRoles` and `PushPathAccessRoles` based on the client code above?

sanderpick: Yep, good catch

threaddb/buckets.go
```diff
@@ -105,6 +300,138 @@ func init() {
 		Name:    buckets.CollectionName,
 		Schema:  bucketsSchema,
 		Indexes: bucketsIndexes,
+		WriteValidator: `
+			if (instance && !instance.version) {
```
sanderpick, line 304: Here is the spaghetti code `WriteValidator` for a bucket. The logic is not terribly complex, but the code ends up being pretty long winded. 

Note the incline comments!

threaddb/buckets.go
```diff
@@ -105,6 +300,138 @@ func init() {
...
+			  case "delete":
+			    if (instance.owner !== "" && writer !== instance.owner) {
+			      return "permission denied" // no owner access to delete instance
+			    }
+			    break
+			}
+			return true
+		`,
+		ReadFilter: `
+			if (!instance.version) {
```
sanderpick, line 402: For old instances, return the the entire thing (don't filter).

threaddb/buckets.go
```diff
@@ -105,6 +300,138 @@ func init() {
...
+			  return instance
+			}
+			if (instance.owner === "") {
```
sanderpick, line 405: Also, don't filter buckets that don't have an owner (buckets created without an identity using `buckd`)

threaddb/buckets.go
```diff
@@ -105,6 +300,138 @@ func init() {
...
+			    var x = instance.metadata[path]
+			    if (x && (x.roles[reader] > 0 || x.roles["*"] > 0)) {
+			      filtered[keys[i]] = m
+			      continue outer
+			    }
+			  }
+			}
+			instance.metadata = filtered
+			if (Object.keys(instance.metadata).length === 0) {
+			  delete instance.key
```
sanderpick, line 431: Delete the link encryption key if the identity has no access to any paths.

threaddb/buckets.go
```diff
@@ -105,6 +300,138 @@ func init() {
...
+		ReadFilter: `
+			if (!instance.version) {
+			  return instance
+			}
+			if (instance.owner === "") {
+			  return instance
+			}
+			var filtered = {}
+			var keys = Object.keys(instance.metadata)
+			outer: for (i = 0; i < keys.length; i++) {
```
sanderpick, line 410: The idea here is to walk each metadata object, collecting ones that the identity has access to, considering the tree inheritance pattern mentioned earlier.

buckets/buckets.go
```diff
@@ -12,6 +15,16 @@ const (
 	SeedName = ".textileseed"
 )
 
+// Role describes an access role for a bucket item.
+type Role int
```
sanderpick, line 19: Here are the actual access roles. I went with this pattern as opposed to _read_, _write_, _delete_, etc., because those don't imply any relationship between the roles, i.e., _write_ doesn't really imply _read_. However, a writer sounds more clearly like they are also able to read... an admin can do it all, etc.

api/buckets/service.go
```diff
@@ -132,86 +177,119 @@ func (s *Service) Create(ctx context.Context, req *pb.CreateRequest) (*pb.Create
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("creating bucket: invalid token public key")
+		}
+	} else {
+		owner = getOwnerFromContext(ctx)
+	}
+
+	// Create bucket keys if private
+	var linkKey, fileKey []byte
+	if private {
```
sanderpick, line 231: This is creating the single link key and the initial top-level file key. More file keys would be added if paths were opened to other users.

api/buckets/service.go
```diff
@@ -403,13 +492,14 @@ func (nn *namedNodes) Store(c cid.Cid, n *namedNode) {
 // Leaf nodes are encrypted and linked to parents, which are then encrypted and
 // linked to their parents, and so on up to root.
 // add will be added to the encrypted root node if not nil.
-// This method returns a map of all nodes keyed by their _original_ plaintext cid.
-func (s *Service) encryptDag(ctx context.Context, ds ipld.DAGService, root ipld.Node, key []byte, add *namedNode) (map[cid.Cid]*namedNode, error) {
+// This method returns a map of all nodes keyed by their _original_ plaintext cid,
+// and a list of the root's direct links.
+func (s *Service) encryptDag(ctx context.Context, ds ipld.DAGService, root ipld.Node, destPath string, linkKey []byte, currentFileKeys, newFileKeys map[string][]byte, newFileKey []byte, add *namedNode) (map[cid.Cid]*namedNode, error) {
```
sanderpick, line 497: This method can now decrypt the dag before re-encrypting with the given key map. The key map holds the correct file key for each path, unless a single key is given for all paths (the case when creating a bucket from an existing Cid).

api/buckets/service.go
```diff
@@ -616,36 +739,74 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (*pb.SetP
...
+		UpdatedAt: buck.UpdatedAt,
+	})
+	buck.UnsetMetadataWithPrefix(destPath + "/")
+
+	txn, err := s.Buckets.WriteTxn(ctx, dbID, tdb.WithToken(dbToken))
 	if err != nil {
-		return nil, fmt.Errorf("invalid remote cid: %s", err)
+		return nil, err
+	}
+	if err = txn.Verify(ctx, buck); err != nil {
```
sanderpick, line 763: Here's this new `Verify` method in use. The idea is that we should not change the underlying UnixFs bucket directory if the final bucket save is going to fail. Instead,
- Create a transaction
- Check if the change will success (if the token has permission)
- Perform the UnixFs changes
- Finally, save the bucket and end the transaction (commit the changes)

Still, this is not perfect... any non-access related errors during save will leave the relationship between the UnixFs directory and the bucket state broken. Though, this is not different than the current setup.

api/buckets/service.go
```diff
@@ -616,36 +739,74 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (*pb.SetP
...
+	buckPath := path.New(buck.Path)
+	dirPath, err := s.setPathFromExistingCid(ctx, buck, buckPath, destPath, bootCid, linkKey, fileKey)
+	if err != nil {
+		return nil, err
+	}
+	buck.Path = dirPath.String()
+	if err = txn.Save(ctx, buck); err != nil {
+		return nil, err
+	}
+	if err = txn.End(); err != nil {
```
sanderpick, line 785: Here's the final commit.

api/buckets/service.go
```diff
@@ -817,30 +997,38 @@ func (s *Service) nodeToItem(ctx context.Context, node ipld.Node, pth string, ke
...
+		Cid:      node.Cid().String(),
+		Name:     gopath.Base(pth),
+		Path:     pth,
+		Size:     int64(stat.CumulativeSize),
+		Metadata: pmd,
 	}
-	for _, l := range node.Links() {
-		if l.Name == "" {
-			break
+	if pn, ok := node.(*dag.ProtoNode); ok {
```
sanderpick, line 1007: This is a better way to determine if this node is a directory. The prior method could not detect empty dirs (they were marked as non a directory).

api/buckets/service.go
```diff
@@ -917,15 +1104,35 @@ func (s *Service) PushPath(server pb.APIService_PushPathServer) error {
...
 	}
 	if root != "" && root != buck.Path {
 		return status.Error(codes.FailedPrecondition, buckets.ErrNonFastForward.Error())
 	}
 
+	buck.UpdatedAt = time.Now().UnixNano()
+	buck.SetMetadataAtPath(filePath, tdb.Metadata{
+		UpdatedAt: buck.UpdatedAt,
+	})
+	buck.UnsetMetadataWithPrefix(filePath + "/")
```
sanderpick, line 1121: In the case where this push is overwriting a directory, we want to wipe out the metadata for those killed paths.

api/buckets/service.go
```diff
@@ -1584,6 +1809,155 @@ func (s *Service) removeNodeAtPath(ctx context.Context, pth path.Path, key []byt
...
+			continue
+		}
+		if r > 0 {
+			target.Roles[k] = r
+		} else {
+			delete(target.Roles, k)
+		}
+		changed = true
+	}
+	if changed {
```
sanderpick, line 1876: This method edits the access roles for the requested path. The idea here is to determine if the request, which contains a map of identities and roles, actually amounts to a change.

api/buckets/service.go
```diff
@@ -1584,6 +1809,155 @@ func (s *Service) removeNodeAtPath(ctx context.Context, pth path.Path, key []byt
...
+
+		if buck.IsPrivate() {
+			if err := buck.RotateFileEncryptionKeysForPrefix(reqPath); err != nil {
+				return nil, err
+			}
+			newFileKeys, err := buck.GetFileEncryptionKeysForPrefix(reqPath)
+			if err != nil {
+				return nil, err
+			}
+			nmap, err := s.encryptDag(ctx, s.IPFSClient.Dag(), pathNode, reqPath, linkKey, currentFileKeys, newFileKeys, nil, nil)
```
sanderpick, line 1897: Re-encrypt the subtree if the roles were altered.

api/buckets/service.go
```diff
@@ -1913,15 +2287,38 @@ func (s *Service) pinBlocks(ctx context.Context, nodes []ipld.Node) error {
 // sumBytesPinned adds the provided delta to the buckets total size from
 // the account/user.
 func (s *Service) sumBytesPinned(ctx context.Context, delta int64) error {
-	a := accountFromContext(ctx)
+	var a *mdb.Account
+	var u *mdb.User
+	if owner, ok := ctx.Value(ctxKey("owner")).(string); ok && owner != "" {
```
sanderpick, line 2292: This is a bit of a mess, but the idea is to target the storage for the bucket owner, not necessarily the account/user that is making the change.

threaddb/buckets.go
```diff
@@ -66,6 +100,167 @@ type Deal struct {
...
+// Version 0 buckets use the link key for all files and folders.
+// Version 1 buckets only use the link for folders.
+func (b *Bucket) GetLinkEncryptionKey() []byte {
+	return keyBytes(b.LinkKey)
+}
+
+// GetFileEncryptionKeyForPath returns the encryption key for path.
+// Version 0 buckets use the link key for all paths.
+// Version 1 buckets use a different key defined in path metadata.
+func (b *Bucket) GetFileEncryptionKeyForPath(pth string) ([]byte, error) {
```
sanderpick, line 118: Note: The key returned may not be from metadata _that is at the path_, i.e., it could be from a parent.

threaddb/buckets.go
```diff
@@ -66,6 +100,167 @@ type Deal struct {
...
+func keyBytes(k string) []byte {
+	if k == "" {
+		return nil
+	}
+	b, _ := base64.StdEncoding.DecodeString(k)
+	return b
+}
+
+// GetFileEncryptionKeysForPrefix returns a map of keys for every path under prefix.
+func (b *Bucket) GetFileEncryptionKeysForPrefix(pre string) (map[string][]byte, error) {
```
sanderpick, line 139: This is used to build a map of keys for _every_ metadata path under the given prefix. As noted above, any given key might be from a parent.

threaddb/buckets.go
```diff
@@ -66,6 +100,167 @@ type Deal struct {
...
+		return md, at, true
+	}
+
+	// Check for an exact match
+	if md, ok = b.Metadata[pth]; ok {
+		if !b.IsPrivate() || !requireKey || md.Key != "" {
+			return md, pth, true
+		}
+	}
+	// Check if we can see this path via a parent
```
sanderpick, line 192: Walk up the tree, looking for metadata. Note, the read filter will only include metadata which the identity has access to... so, if this method doesn't find metadata or inherited metadata for the requested path, it's not visible.

threaddb/buckets.go
```diff
@@ -66,6 +100,167 @@ type Deal struct {
...
+				return nil, fmt.Errorf("could not resolve path: %s", p)
+			}
+			keys[p] = keyBytes(md.Key)
+		}
+	}
+	return keys, nil
+}
+
+// RotateFileEncryptionKeysForPrefix re-generates existing metadata keys for every path under prefix.
+func (b *Bucket) RotateFileEncryptionKeysForPrefix(pre string) error {
```
sanderpick, line 158: This method walk the metadata subtree beginning with `pre`, and re-generates each file key.

sanderpick (commented):
Many comments 😅

api/buckets/service.go
```diff
@@ -1584,6 +1809,155 @@ func (s *Service) removeNodeAtPath(ctx context.Context, pth path.Path, key []byt
...
+	newRoles, err := buckets.RolesFromPb(req.Roles)
+	if err != nil {
+		return nil, err
+	}
+	var changed bool
+	for k, r := range newRoles {
+		if x, ok := target.Roles[k]; ok && x == r {
+			continue
+		}
+		if r > 0 {
```
jsign, line 1869: Nit: Use `Role.None`

api/buckets/service.go
```diff
@@ -1584,6 +1809,155 @@ func (s *Service) removeNodeAtPath(ctx context.Context, pth path.Path, key []byt
...
+			pn := nmap[pathNode.Cid()].node
+			dirPath, err := s.insertNodeAtPath(ctx, pn, path.Join(path.New(buck.Path), reqPath), linkKey)
+			if err != nil {
+				return nil, fmt.Errorf("updating pinned root: %s", err)
+			}
+			if err := s.addAndPinNodes(ctx, nodes); err != nil {
+				return nil, err
+			}
+			buck.Path = dirPath.String()
+		}
```
jsign, line 1916: Q: So the old subtree was unpined as a result of above methods, right?

api/buckets/service.go
jsign: Q: So the owner of the bucket is always loaded in context in the interceptor?

Should the below `getOwnerFromContext` be renamed? Doesn't look like returning doing the same logic or using "owner" in another sense?

buckets/archive/tracker.go
```diff
@@ -33,7 +32,7 @@ type Tracker struct {
 	lock   sync.Mutex
 	ctx    context.Context
 	cancel context.CancelFunc
-	closed chan (struct{})
+	closed chan struct{}
```
jsign, line 35: Huh, that `(struct{})` was weird.

buckets/buckets.go
jsign: Sounds good.

threaddb/buckets.go
```diff
@@ -29,23 +34,52 @@ var (
...
-}
-
-// GetEncKey returns the encryption key as bytes if present.
-func (b *Bucket) GetEncKey() []byte {
-	if b.EncKey == "" {
-		return nil
+	Key       string              `json:"_id"`
+	Owner     string              `json:"owner"`
+	Name      string              `json:"name"`
+	Version   int                 `json:"version"`
```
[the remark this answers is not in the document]
jsign: Nice!

threaddb/buckets.go
jsign: 🤯

util/flakyt.go
```diff
@@ -46,7 +46,7 @@ func (ft *FlakyT) Cleanup(cls func()) {
 	ft.cls = append([]func(){cls}, ft.cls...)
 }
 
-var numRetries = 5
+var numRetries = 50
```
jsign, line 49: Uhm, I'd lower this to 5 again... or maybe 10. But 50 is too much and will be running forever on a real error.
With the new local-devnet update things should be less flaky.

jsign (approved):
LGTM! I had to believe you in many parts!

api/buckets/service.go
```diff
@@ -457,21 +557,26 @@ func (s *Service) encryptDag(ctx context.Context, ds ipld.DAGService, root ipld.
 	eg, gctx := errgroup.WithContext(ctx)
 	for _, l := range lmap {
 		l := l
+		cfk := getFileKey(nil, currentFileKeys, l.path)
+		nfk := getFileKey(newFileKey, newFileKeys, l.path)
```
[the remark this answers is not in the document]
sanderpick: Yep. I kinda hate that API... but it's the cleanest way I could think of to handle that edge case. Originally, I had a `"*"` entry in the map for that, but really any wildcard like string _could_ conflict with a valid path.

api/buckets/service.go
```diff
@@ -457,21 +557,26 @@ func (s *Service) encryptDag(ctx context.Context, ds ipld.DAGService, root ipld.
...
+		if nfk == nil {
+			return nil, fmt.Errorf("new file key not found for path %s", l.path)
+		}
```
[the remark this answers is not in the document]
sanderpick: It shouldn't happen... I'll make an inline comment there.

api/buckets/client/client.go
```diff
@@ -315,6 +316,35 @@ func (c *Client) RemovePath(ctx context.Context, key, pth string, opts ...Option
 	return util.NewResolvedPath(res.Root.Path)
 }
 
+// PushPathAccessRoles updates path access roles by merging the pushed roles with existing roles.
+// roles is a map of string marshaled public keys to path roles. A non-nil error is returned
+// if the map keys are not unmarshalable to public keys.
+// To delete a role for a public key, set its value to buckets.None.
+func (c *Client) PushPathAccessRoles(ctx context.Context, key, pth string, roles map[string]buckets.Role) error {
```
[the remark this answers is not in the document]
sanderpick: Nice, good call... I wonder, is it worth just adding an initial `Metadata.Version`? I'll make a note below where that would be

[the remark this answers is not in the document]
jsign: Yeah, something like that should make the trick. So the backend can check if the current version is the one sent with the wanted change, and if not: fail.

api/buckets/service.go
sanderpick: Yep, `insertNodeAtPath` handles unpinning old branches either directly or by a pin replace.

util/flakyt.go
sanderpick: Cool, done... also reverted the `require.Eventually` timeouts to the lower values.

api/buckets/service.go
sanderpick: Any API method that can effect bucket size attaches the "owner" directly (not in the interceptor), like
```
ctx = context.WithValue(ctx, ctxKey("owner"), buck.Owner)
```
I went that route instead of passing around the bucket owner string to all these pinning, unpinning, node adding, etc. methods, which looked pretty nasty.

jsign: OK! Do you have any comments about `getOwnerFromContext` possible renaming?
Shouldn't be `getUAccountOrUserFomContext` instead of using the "owner" concept? 
Maybe I'm confused!

sanderpick: Good call, I'll rename `getOwnerFromContext` to be more clear, maybe `getAccountOrUserFromContext`. It will be nice to remove the `User` abstraction and make it a special type of `Account`. I think that alone would help clean this up.

jsign: Oop, we posted at the same time. SGTM!

api/buckets/service.go
sanderpick: Oh yeah, good catch. The error from ending the transaction is important... I went with this style,
```
txn, err := s.Buckets.WriteTxn(ctx, dbID, tdb.WithToken(dbToken))
if err != nil {
	return nil, err
}
defer func() {
	if e := txn.End(); err == nil {
		err = e
	}
}()
```
Where `err` is named param in the surrounding function.

jsign: But what indicates that End will commit or rollback?
On an early return or in a success case both will call `End` without any extra thing.

jsign:
```
	if e := txn.End(); err == nil {
		err = e
	}
```
Is that `== nil` intended, or `!= nil`?

api/buckets/client/client.go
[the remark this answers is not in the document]
sanderpick: On second thought, I'm not sure this is needed or not. What you're suggesting is like an etag, in which case at some point you have fetched that thing that you're changing. Here, we're just requesting that a change be merged into the existing state. As in, _add/remove this role_... it's not a _put_ style, replace the roles with this new set. Make sense?

api/buckets/service.go
sanderpick: `End` will commit if there are things to commit, which in this case only happens if `Save` below succeeds. 

The `err` check is saying, if the error isn't already set (from outside the defer), return the one that happened here.

jsign: Ah, was `err == nil` not `e == nil`. :)

Regarding `End`: but what happens if you have two `Save(...)`, first succeeds, and then there's an early return before the second `Save(..)` call. You don't want the first `Save(...)` to be committed right?

sanderpick: Ah yeah, that would be bad. Just to be clear about the correct usage, I could add a manual `Discard` to the error catch when using save.

jsign: Cool, yeah. So the client should call `Discard` in any kind of early return while the transaction is open.
Might worth adding to the "WriteTxn(..)" comments 👍🏼

sanderpick: Hrm, looks like I need to add `Discard` to the DB API.

PR #320, opened 2020-08-20
Add bingo and buf

asutula:
[Bingo](https://github.com/bwplotka/bingo) manages dev tooling packages and versioning. I'm using it to install and run [Buf](https://github.com/bufbuild/buf)which is used to lint `.proto` files and check for breaking API changes.

This PR integrates those tools, uses them in CI workflows, fixes all the lint problems, re-generates from the fixed protos, and updates the code to accommodate those changes. 

There are some breaking API changes, so keep an eye on those to make sure they are acceptable. This is the one painful time since we're going from no linting to linting, so breaking changes should be non existent or rare in the future.

The other thing to keep an eye on... I don't think we're persisting any protobuf data, but if so, there may be problems since I made some breaking changes to the proto definitions. Let me know if you know of any case where this might apply.


.bingo/Variables.mk
```diff
@@ -0,0 +1,36 @@
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.2.3. DO NOT EDIT.
```
asutula, line 1: Bingo is great, you can use these Makefile variables in any other Makefile, and it will assure that the correct version of the tool is installed and used. Check out `Makefile` to see it in action.

.bingo/variables.env
```diff
@@ -0,0 +1,16 @@
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.2.3. DO NOT EDIT.
```
asutula, line 1: It also creates these shell script variables that can be used. Not as magic as the Makefile ones.

.github/workflows/buf.yml
```diff
@@ -0,0 +1,27 @@
+name: Buf
```
asutula, line 1: Run `buf` in CI to do lint checks, and in the future, breaking change detection.

Makefile
```diff
@@ -18,3 +22,25 @@ buck-stop:
 
 buck-clean:
 	docker-compose -f cmd/buckd/docker-compose-dev.yml down -v --remove-orphans
+
+# local is what we run when testing locally.
+# This does breaking change detection against our local git repository.
+.PHONY: buf-local
+buf-local: $(BUF)
```
asutula, line 29: This is now you use the Bingo Makefile vars. Here is make `buf-local` depend on `$(BUF)` which will make sure the correct version of `buf` is downloaded, built, and used.

Makefile
```diff
@@ -18,3 +22,25 @@ buck-stop:
...
+	$(BUF) check lint
```
asutula, line 30: And here is how I call `buf` using the var

buckets/local/archive.go
```diff
@@ -106,6 +106,7 @@ func (b *Bucket) ArchiveStatus(ctx context.Context, watch bool) (<-chan ArchiveS
 				r, err := b.clients.Buckets.ArchiveStatus(sctx, key)
 				if err != nil {
 					msgs <- ArchiveStatusMessage{Type: ArchiveError, Error: err}
+					scancel()
```
asutula, line 109: VSCode caught that this was leaking, so fixed it.

api/buckets/pb/buckets.proto
```diff
@@ -195,27 +191,27 @@ message ArchiveWatchRequest {
...
 }
 
-message ArchiveWatchReply {
+message ArchiveWatchResponse {
     string msg = 1;
 }
 
-service API {
-    rpc List(ListRequest) returns (ListReply) {}
-    rpc Init(InitRequest) returns (InitReply) {}
```
asutula, line 204: `Init` is no good in Obj C and Swift because its a reserved name. Went with `Create`, but let me know if there is something better.

buf.yaml
```diff
@@ -0,0 +1,16 @@
+build:
```
asutula, line 1: `buf` config file

buf.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - .
+  excludes:
+    - api/buckets/pb/javascript
+    - api/hub/pb/javascript
+    - api/users/pb/javascript
+lint:
+  use:
+    - DEFAULT
+  except:
+    - PACKAGE_VERSION_SUFFIX
```
asutula, line 12: I'm ignoring this rule for now, but it's actually cool if we want to use it one day. It enforces that your proto package path include a version folder like `v0` so that, when you choose, you can break API compatibility... a client using `some/path/v0/myservice.proto` simply can't talk to a server serving `some/path/v1/myservice.proto`.

api/hub/service.go
```diff
@@ -204,50 +204,67 @@ func getSessionSecret(secret string) string {
...
 
-func (s *Service) CreateKey(ctx context.Context, req *pb.CreateKeyRequest) (*pb.GetKeyReply, error) {
+func (s *Service) CreateKey(ctx context.Context, req *pb.CreateKeyRequest) (*pb.CreateKeyResponse, error) {
 	log.Debugf("received create key request")
 
 	owner := ownerFromContext(ctx)
-	key, err := s.Collections.APIKeys.Create(ctx, owner, mdb.APIKeyType(req.Type), req.Secure)
+
+	var keyType mdb.APIKeyType
+	switch req.Type {
```
asutula, line 238: Treating all conversion to/from proto enums explicitly is a good practice. I added a "unspecified" value in the 0 position of each of them, so the number doesn't always align with our native Go representation, which isn't ever a good thing to assume anyway.

asutula: Made similar changes elsewhere

PR #323, opened 2020-08-24
Powergate interceptor, client, CLI

asutula:
This PR adds Powergate API support to Hub for Powergate calls that can simply be proxied through to Powergate. Calls to `PushStorageConfig` are excluded because the primary way of saving data to Powergate is via Bucket archiving.

Client and CLI support is included as well.


api/pow/client/client.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// Client provides the client api.
+type Client struct {
+	healthC healthRpc.RPCServiceClient
+	netC    netRpc.RPCServiceClient
+	ffsC    ffsRpc.RPCServiceClient
+	walletC walletRpc.RPCServiceClient
+	conn    *grpc.ClientConn
+}
```
asutula, line 20: Ended up adding all the pow client code as it's own package. Didn't make sense (for `FFS` related code) to add it as a sub package of `users` or `hub` apis because `FFS` related calls are relevant to both `User`s and `Account`s.

asutula: It's just one file/package now, we can always split it up if we add more apis in here later

api/pow/client/client_test.go
```diff
@@ -0,0 +1,150 @@
...
+	hc "github.com/textileio/textile/api/hub/client"
+	"github.com/textileio/textile/core"
+	"github.com/textileio/textile/integrationtest/pg"
+	"github.com/textileio/textile/util"
+	"google.golang.org/grpc"
+)
+
+const powAddr = "127.0.0.1:5002"
+
+func TestPowClient(t *testing.T) {
```
asutula, line 22: Simple integration test to make sure all the wires connect from the client all the way through to Powergate.

cmd/cmd.go
```diff
@@ -80,6 +81,7 @@ type Clients struct {
 	Threads *tc.Client
 	Hub     *hc.Client
 	Users   *uc.Client
+	Pow     *pc.Client
```
asutula, line 84: Use the new pow client in the CLI

cmd/hub/cli/pow.go
```diff
@@ -0,0 +1,249 @@
...
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/spf13/cobra"
+	ffsRpc "github.com/textileio/powergate/ffs/rpc"
+	"github.com/textileio/textile/cmd"
+)
```
asutula, line 11: Simple CLI for Powergate-related calls, all commands live under `hub pow`. Keeping it as simple as possible for now, just calling the methods and printing out the response.

core/core.go
```diff
@@ -68,6 +78,36 @@ var (
 		"/threads.pb.API/ListDBs",
 	}
 
+	healthServiceName = "health.rpc.RPCService"
+	netServiceName    = "net.rpc.RPCService"
+	ffsServiceName    = "ffs.rpc.RPCService"
+	walletServiceName = "wallet.rpc.RPCService"
```
asutula, line 84: Just some constants to be used in this file.

core/core.go
```diff
@@ -68,6 +78,36 @@ var (
...
+			"Info",
+			"Show",
+			"ShowAll",
+			"ListStorageDealRecords",
+			"ListRetrievalDealRecords",
+		},
+		walletServiceName: {
+			"Balance",
+		},
+	}
```
asutula, line 109: All allowed Powergate calls by service name. Easily editable at any time.

core/core.go
```diff
@@ -283,8 +323,37 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
 		opts = []grpc.ServerOption{
-			grpcm.WithUnaryServerChain(auth.UnaryServerInterceptor(t.authFunc), t.threadInterceptor()),
+			grpcm.WithUnaryServerChain(
+				auth.UnaryServerInterceptor(t.authFunc),
+				t.threadInterceptor(),
+				generatePowUnaryInterceptor(healthServiceName, allowedPowMethods[healthServiceName], healthServiceDesc, powStub, t.powc, t.collections),
+				generatePowUnaryInterceptor(netServiceName, allowedPowMethods[netServiceName], netServiceDesc, powStub, t.powc, t.collections),
+				generatePowUnaryInterceptor(ffsServiceName, allowedPowMethods[ffsServiceName], ffsServiceDesc, powStub, t.powc, t.collections),
+				generatePowUnaryInterceptor(walletServiceName, allowedPowMethods[walletServiceName], walletServiceDesc, powStub, t.powc, t.collections),
+			),
```
asutula, line 356: Idea here is to always register the powergate interceptors, even if powergate isn't configured to be running as part of Hub. That way, the interceptors can detect that and return a nice error saying powergate isn't enabled.

core/core.go
```diff
@@ -283,8 +323,37 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 	}
 	var opts []grpc.ServerOption
 	if conf.Hub {
+		var powStub *grpcdynamic.Stub
+		var healthServiceDesc *desc.ServiceDescriptor
+		var netServiceDesc *desc.ServiceDescriptor
+		var ffsServiceDesc *desc.ServiceDescriptor
+		var walletServiceDesc *desc.ServiceDescriptor
+		if conf.AddrPowergateAPI != "" {
+			if powStub, err = createPowStub(conf.AddrPowergateAPI); err != nil {
```
asutula, line 332: Stub is used to dynamically make calls to powergate gRPC service.

core/core.go
```diff
@@ -283,8 +323,37 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
+			if healthServiceDesc, err = createServiceDesciptor("health/rpc/rpc.proto", healthServiceName); err != nil {
+				return nil, err
+			}
+			if netServiceDesc, err = createServiceDesciptor("net/rpc/rpc.proto", netServiceName); err != nil {
+				return nil, err
+			}
+			if ffsServiceDesc, err = createServiceDesciptor("ffs/rpc/rpc.proto", ffsServiceName); err != nil {
+				return nil, err
+			}
+			if walletServiceDesc, err = createServiceDesciptor("wallet/rpc/rpc.proto", walletServiceName); err != nil {
```
asutula, line 344: Each of these service descriptors allows us to look up methods by name to be executed against the stub

core/core.go
```diff
@@ -304,6 +373,10 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
 		if conf.Hub {
 			hpb.RegisterAPIServiceServer(t.server, hs)
 			upb.RegisterAPIServiceServer(t.server, us)
+			healthRpc.RegisterRPCServiceServer(t.server, &healthRpc.UnimplementedRPCServiceServer{})
+			netRpc.RegisterRPCServiceServer(t.server, &netRpc.UnimplementedRPCServiceServer{})
+			ffsRpc.RegisterRPCServiceServer(t.server, &ffsRpc.UnimplementedRPCServiceServer{})
+			walletRpc.RegisterRPCServiceServer(t.server, &walletRpc.UnimplementedRPCServiceServer{})
```
asutula, line 379: For Hub gRPC server to process requests to powergate, we have to register _any_ service implementations, even if they aren't used. So I use the generated "unimplemented" services.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
+func generatePowUnaryInterceptor(serviceName string, allowedMethods []string, serviceDesc *desc.ServiceDescriptor, stub *grpcdynamic.Stub, pc *powc.Client, c *mdb.Collections) grpc.UnaryServerInterceptor {
```
asutula, line 633: Factory method to generate the powergate interceptor functions. All the needed dependencies are injected here.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		methodParts := strings.Split(info.FullMethod, "/")
+		if len(methodParts) != 3 {
+			return nil, status.Errorf(codes.Internal, "error parsing method string %s", info.FullMethod)
+		}
+		methodServiceName := methodParts[1]
+		methodName := methodParts[2]
+
+		if methodServiceName != serviceName {
+			return handler(ctx, req)
```
asutula, line 643: If the request is for a service other than the one this interceptor is handing, just pass it through.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+		}
+
+		if pc == nil {
+			return nil, status.Error(codes.Internal, "Powergate isn't enabled in Hub")
```
asutula, line 647: If we don't have powergate configured to be running, return that error.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+		}
+
+		isAllowedMethod := false
+		for _, method := range allowedMethods {
+			if method == methodName {
+				isAllowedMethod = true
+				break
+			}
+		}
```
asutula, line 656: Now we know we're dealing with a request for the service we're configured to be handling, make sure the request method is allowed.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+
+		if !isAllowedMethod {
+			return nil, status.Errorf(codes.PermissionDenied, "method not allowed: %s", info.FullMethod)
```
asutula, line 659: If not allowed, return that error.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+			isAccount = true
+		} else if dev, ok := mdb.DevFromContext(ctx); ok {
+			ffsInfo = dev.FFSInfo
+			owner = dev.Key
+			isAccount = true
+		} else if user, ok := mdb.UserFromContext(ctx); ok {
+			ffsInfo = user.FFSInfo
+			owner = user.Key
+			isAccount = false
+		}
```
asutula, line 677: Get account/user info from the context

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+				return fmt.Errorf("updating user/account with new ffs information: %v", err)
+			}
+			return nil
+		}
+
+		tryAgain := fmt.Errorf("powergate newly integrated into your account, please try again in 30 seconds to allow time for setup to complete")
+
+		// case where account/user was created before powergate was enabled.
+		// create a ffs instance for them.
+		if ffsInfo == nil {
```
asutula, line 699: Like the comment says, lazily create the ffs instance if needed.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+			if err := createNewFFS(); err != nil {
+				return nil, err
+			}
+			return nil, tryAgain
```
asutula, line 703: Even though not all calls to powergate rely on some time passing for wallet funding to complete, it's just easy/safe to return this error instead of building recursive retry logic.

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+		}
+
+		ffsCtx := context.WithValue(ctx, powc.AuthKey, ffsInfo.Token)
+
+		methodDesc := serviceDesc.FindMethodByName(methodName)
+		if methodDesc == nil {
+			return nil, status.Errorf(codes.Internal, "no method found for %s", methodName)
+		}
+
+		res, err := stub.InvokeRpc(ffsCtx, methodDesc, req.(proto.Message))
```
asutula, line 713: Ok, we know we have our account/user and their ffs information and context, go ahead and pass the call through to powergate using the stub

core/core.go
```diff
@@ -557,6 +630,103 @@ func (t *Textile) noAuthFunc(ctx context.Context) (context.Context, error) {
...
+		if err != nil {
+			if err.Error() != "auth token not found" {
+				return nil, err
+			} else {
+				// case where the ffs token is no longer valid because powergate was reset.
+				// create a new ffs instance for them.
+				if err := createNewFFS(); err != nil {
```
asutula, line 720: Powergate has been reset and the ffs token is no longer valid, create a new ffs instance and return the try again error

integrationtest/pg/pg_test.go
```diff
@@ -303,58 +298,3 @@ func reSetup(t util.TestingTWithCleanup, conf core.Config) *c.Client {
 
 	return client
 }
-
-func spinup(t util.TestingTWithCleanup) *pc.Client {
```
asutula, line 307: Moved this to another file and exported it so it can be used elsewhere.

integrationtest/pg/powergate.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/textileio/powergate/health"
+	"github.com/textileio/textile/util"
+	"google.golang.org/grpc"
+)
+
+var powAddr = "127.0.0.1:5002"
+
+func StartPowergate(t util.TestingTWithCleanup) *pc.Client {
+	_, currentFilePath, _, _ := runtime.Caller(0)
+	dirpath := path.Dir(currentFilePath)
```
asutula, line 22: Little trick to make sure the `docker-compose.yml` file is found when this func is used from elsewhere in the project.

api/pow/client/client.go
asutula: CC @andrewxhill similar additions will be made to the JS Hub client next.

andrewxhill: cool, makes sense

andrewxhill: well... wait, what do you mean "Users and Accounts"? 

there is no Accounts object, class, API used in JS

asutula: Oh I was just talking about the structure of the client API in Go... It has a package `hub` that is used by developers to manage their dev and org `Account`s and a package `users` that is used by end `User`s to interact with threads. Powergate interactions make sense in both of those contexts (because FFS instances are created for both `User`s and `Account`s), so in this case, I just keep the powergate client code in it's own package that can be used however makes sense.

In the case of `js-hub`, there is currently no equivalent of `hub` admin APIs, so we may choose to integrate the powergate apis in differently. If we ever did add the admin apis to the JS lib, we'd be facing the same situation of powergate apis applying to both admin `Account`s and end `User`s.

core/core.go
asutula: Right, if you don't register these, you gRPC will return an error to the client before your interceptors are even called.

PR #325, opened 2020-09-01
GetOrCeate BucketArchive, better error messages

asutula:
This fixes an edge case where a Bucket exists before archiving was implemented so its `BucketArchive` record can't be found. Also removes the "ffs" wording from publicly exposed error messages and a couple places where "bucket archive" as incorrectly referred to as "ffs".

### DIFF
--- a/buckets/archive/tracker.go
+++ b/buckets/archive/tracker.go
@@ -98,7 +98,7 @@ func (t *Tracker) run() {
 						}
 
 						if powInfo == nil {
-							if err := t.colls.ArchiveTracking.Finalize(ctx, a.JID, "huh? no FFS info, weird"); err != nil {
+							if err := t.colls.ArchiveTracking.Finalize(ctx, a.JID, "no powergate info found"); err != nil {
 								log.Errorf("finalizing errored/rescheduled archive tracking: %s", err)
 							}
 							return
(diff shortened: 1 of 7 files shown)


mongodb/bucketarchives.go
```diff
@@ -56,10 +55,14 @@ func (k *BucketArchives) Replace(ctx context.Context, ffs *BucketArchive) error
 	return nil
 }
 
-func (k *BucketArchives) Get(ctx context.Context, bucketKey string) (*BucketArchive, error) {
+func (k *BucketArchives) GetOrCreate(ctx context.Context, bucketKey string) (*BucketArchive, error) {
```
asutula, line 58: This is the main change that fixes the bug that emerged this morning.

buckets/archive/tracker.go
```diff
@@ -98,7 +98,7 @@ func (t *Tracker) run() {
 						}
 
 						if powInfo == nil {
-							if err := t.colls.ArchiveTracking.Finalize(ctx, a.JID, "huh? no FFS info, weird"); err != nil {
```
sanderpick, line 101: 😂

mongodb/bucketarchives.go
```diff
@@ -32,21 +32,20 @@ type BucketArchives struct {
 }
 
 func NewBucketArchives(_ context.Context, db *mongo.Database) (*BucketArchives, error) {
-	// ToDo: Let's rename this collection to bucketarchives when there is a data loss event like a powergate reset
-	s := &BucketArchives{col: db.Collection("ffsinstances")}
+	s := &BucketArchives{col: db.Collection("bucketarchives")}
```
sanderpick, line 35: 👍

PR #327, opened 2020-09-09
feat(cli): add version command to CLI

andrewxhill:
- Adds version command where prod versions are set by ldflags
- [x] Needs to be added to `gox` build command in release action
- [ ] Can be replicated in the `buck` CLI if this is the right way to go, but didn't know if it made sense before any split happens.
- [x] Allow the -upgrade flag to be used from a `development` version => production latest version. This would be handy for any developer (especially on mac) to go from `make textile` back to production version in a single command

Little preview

version-update

* Doesn't choose any release candidates, only production releases
* Could optionally include an override flag to choose to install version you want manually

Only partially done because I wasn't certain if this looks like a good approach.


cmd/cmd.go
```diff
@@ -20,6 +20,9 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
+// Version is the current semvar or "development" if not production
+var Version = "development"
```
andrewxhill, line 24: any `go install` version will just be development, unless setting ldflags manually.

cmd/hub/cli/version.go
```diff
@@ -0,0 +1,75 @@
...
+		version := cmd.Version
+
+		update, err := c.Flags().GetBool("update")
+		cmd.ErrCheck(err)
+
+		current, err := semver.ParseTolerant(version)
+		if err == nil {
+			config := su.Config{
+				Filters: []string{
+					"hub",
```
andrewxhill, line 29: just filtering out the `buck` release assets

asutula:
I'm quite happy with how we did this in Powergate, using `govvv` to set version information during build time, while simultaneously using `gox` for releases. Might be worth checking out, and would be especially cool if the `--upgrade` functionality could be integrated along side `govvv` and `gox`.

andrewxhill:
that looks easy enough if having the dev build with some kind of version is a requirement

cmd/hub/cli/cli.go
```diff
@@ -54,6 +54,8 @@ func Init(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(bucketCmd)
 	buck.Init(bucketCmd)
 
+	versionCmd.Flags().BoolP("update", "u", false, "Update hub cli to latest release")
```
sanderpick, line 57: Feels to me like `update` should be its own command. Thoughts?

cmd/hub/cli/version.go
```diff
@@ -0,0 +1,75 @@
...
+				if current.LT(latest.Version) {
+					if update {
+						// Update cli if requested
+						exe, err := os.Executable()
+						if err != nil {
+							cmd.Fatal(err)
+							return
+						}
+
+						s := spin.New("%s Downloading release")
```
sanderpick, line 54: Slick

cmd/cmd.go
sanderpick: Cool! I'm partial to `govvv` as well here... it's nice to see how far off the latest mater commit you are, and if the branch is "dirty" or not. What you have here could be used in conjunction w/ that, i.e., using `make` would properly setup `govvv`, but just a normal `go install` would use this default `Version` tag. Thoughts?

andrewxhill: yep, that's easy

cmd/hub/cli/cli.go
andrewxhill: i've got the least opinion here. i had originally intended to go that way, but then felt that the nav felt cleaner with just a version command

sanderpick: K, cool. My two cent: We still have relatively few commands... and putting it on the top level will make it more noticeable as well

.github/workflows/release.yml
```diff
@@ -26,14 +26,16 @@ jobs:
...
           go get -v -t -d ./...
-      - name: Install gox
+      - name: Build govvv
         run: |
-          export PATH=${PATH}:`go env GOPATH`/bin
-          go get github.com/mitchellh/gox  
+          go build -modfile .bingo/govvv.mod -o=. github.com/ahmetb/govvv
+      - name: Build gox
+        run: |
+          go build -modfile .bingo/gox.mod -o=. github.com/mitchellh/gox
```
andrewxhill, line 34: need to verify that this wouldn't mess with the `buck` build below that doesn't need govvv yet

andrewxhill: @asutula seems okay right?

Makefile
```diff
@@ -2,8 +2,10 @@ include .bingo/Variables.mk
 
 .DEFAULT_GOAL=textile
 
-textile:
-	go install ./...
+GOVVV_FLAGS=$(shell $(GOVVV) -version git -flags -pkg $(shell go list ./buildinfo))
```
andrewxhill, line 5: okay, 

default version is `development` for `go install ./...`. running `make textile` changes it to `git` so that i can fully display the git info in the version command results easily. screenshots coming

buildinfo/buildinfo.go
```diff
@@ -0,0 +1,31 @@
+package buildinfo
+
+import "fmt"
+
+var (
+	// GitCommit is set by govvv at build time.
+	GitCommit = ""
```
andrewxhill, line 7: went with empty strings over 'n/a' as it was in powergate. i do some logic against these, so empty strings were more helpful. flagging if you did 'n/a' for a critical reason @asutula

cmd/hub/cli/update.go
```diff
@@ -0,0 +1,79 @@
...
+				if err = install(latest.AssetURL); err != nil {
+					cmd.Error(err)
+					cmd.Warn("Error: install failed.")
+				} else {
+					version = latest.Version.String()
+					cmd.Message("Success: hub updated.")
+				}
+			}
+		}
+		if version == "git" {
```
andrewxhill, line 65: if installed with `make textile` it will display the full results as a table. otherwise, just print the version as the result

cmd/hub/cli/version.go
```diff
@@ -0,0 +1,91 @@
...
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Show current version",
+	Long:  `Shows the installed CLI version.`,
+	Args:  cobra.ExactArgs(0),
+	Run: func(c *cobra.Command, args []string) {
+		version := bi.Version
+
+		apiVersion, err := getAPIVersion()
```
andrewxhill, line 54: this will also let you know what the API you are connected to is running. default `development` if you are building with docker-compose-dev, but again, if that's the case it will print the commit info running on the api

cmd/output.go
```diff
@@ -35,7 +35,7 @@ func End(format string, args ...interface{}) {
 	os.Exit(0)
 }
 
-func Fatal(err error, args ...interface{}) {
+func Error(err error, args ...interface{}) {
```
andrewxhill, line 38: I needed non fatal errors and wanted to ensure they used the same formatting that had been done with fatal

andrewxhill:
A common update flow

Screen Shot 2020-09-09 at 4 44 52 PM

What happens when you can't make a connection to the releases page,

Screen Shot 2020-09-09 at 2 58 13 PM

what happens when you're running a locally built version from `make textile`

Screen Shot 2020-09-09 at 4 42 28 PM

what happens when you update any version

Screen Shot 2020-09-09 at 4 41 51 PM

after taking most of those screenshots, i added the API version line to the response. now seen here,

Screen Shot 2020-09-09 at 5 28 56 PM

.github/workflows/release.yml
asutula: Yea, should be fine. Just building the `govvv` executable shouldn't cause any problems.

andrewxhill:
weird: https://github.com/textileio/textile/pull/327/checks?check_run_id=1097315190#step:7:181 i don't see the footer html, it's true. but the panic is new i guess...? maybe an updated version of gin?

andrewxhill:
ooookay, that error went away. new ones emerge

andrewxhill:
phhhhew, bumped the timeouts and we're back in business

sanderpick:
This looks great @andrewxhill - what are you thoughts on having the version command for `buck` as well? Might be a little cryptic to setup... we can just defer for now if that's what you're thinking

andrewxhill:
yeah, maybe we can close this to be sure it is available for release tomorrow. i can try to tackle a buck follow-up, i deferred for that reason and since i didn't know if it would be split out soon anyway. so either way, yes, but let's close this guy for now

sanderpick:
Cool. It is actually split right now... the `cmd` package spits out a `hub` with nested `buck` and `buck` by itself... but maybe that's not what you meant

sanderpick (approved):
LGTM!

andrewxhill:
cool. yeah, i thought i remembered that there was an idea for it's own repo. but could be mixing things. will get it in there soon

PR #332, opened 2020-09-12
Allow to print archive logs on success

jsign:
Improve CLI experience allowing to print all logs on _watch_ mode if archive finished successfully.

### DIFF
--- a/buckets/local/archive.go
+++ b/buckets/local/archive.go
@@ -111,11 +111,9 @@ func (b *Bucket) ArchiveStatus(ctx context.Context, watch bool) (<-chan ArchiveS
 					return
 				}
 				scancel()
-				final, err := isJobStatusFinal(r.GetStatus())
+				_, err = isJobStatusFinal(r.GetStatus())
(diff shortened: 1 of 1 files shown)


buckets/local/archive.go
```diff
@@ -111,12 +111,10 @@ func (b *Bucket) ArchiveStatus(ctx context.Context, watch bool) (<-chan ArchiveS
 					return
 				}
 				scancel()
-				final, err := isJobStatusFinal(r.GetStatus())
+				_, err = isJobStatusFinal(r.GetStatus())
 				if err != nil {
 					msgs <- ArchiveStatusMessage{Type: ArchiveError, Error: err}
 					cancel()
-				} else if final {
-					cancel()
```
jsign, line 119: The problem here was that if they reach a _final_ status in a good way (success), the context of `ArchiveWatch` was indirectly canceled, so the logs were printed partially.

andrewxhill (approved):

PR #333, opened 2020-09-14
Hotfix/buckets-acl: use correct names in cross user method list.

jsonsivar:
Note: I updated our dev instance of the hub to use this version and it's working.  However when I run `go test ./...` it hangs.  I was able to compile `buckd` and `hubd` fine as well. Let me know if there are any steps to get all the tests working and if there are any other steps I should do before making the PR as per your local dev process.


sanderpick:
Awesome, thanks @jsonsivar - I've been adding more tests and discovered some issues today, including one that would cause tests to hang. Give it another try when you get a chance.

Issue #347, opened 2020-09-18
Remove sendFil API and CLI command


andrewxhill:

jimmylee:
Is this because of legal guidance?

andrewxhill:
correct, yeah

andrewxhill:
good to close right @asutula ?

asutula:
Yep

PR #348, opened 2020-09-18
Makefile builds, improves release builds

asutula:
- removes SendFil closing #347 
- moves all build commands to the `Makefile`
- this takes advantage of Bingo's nice Makefile variables that assure that the correct build tool version is installed in `GOPATH` and used
- fixes the `-dirty` problem we had with `govvv` state
- simplifies release workflow a lot
- makes it easy to test building releases locally (or anywhere not in CI)


.bingo/gomplate.mod
```diff
@@ -0,0 +1,5 @@
+module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
+
+go 1.14
+
+require github.com/hairyhenderson/gomplate/v3 v3.8.0 // cmd/gomplate
```
asutula, line 5: Using Gomplate to create the `install` script included in release bundles

.dockerignore
```diff
@@ -1,5 +1,4 @@
 docker-compose.yml
 docker-compose-dev.yml
-Makefile
```
asutula, line 3: Need to include `Makefile` in the Docker build because we use `make` to build `hubd` in Docker.

.github/workflows/release.yml
```diff
@@ -26,51 +26,12 @@ jobs:
...
-                  mv buck_${VERSION}_${OSARCH}.zip ../${OUT}/
-              else
-                  tar -czvf hub_${VERSION}_${OSARCH}.tar.gz LICENSE README.md install hub
-                  tar -czvf buck_${VERSION}_${OSARCH}.tar.gz LICENSE README.md install buck
-                  mv hub_${VERSION}_${OSARCH}.tar.gz ../${OUT}/
-                  mv buck_${VERSION}_${OSARCH}.tar.gz ../${OUT}/
-              fi
-          done
-          cd .. && rm -rf tmp
-          echo $(ls ./release/cli)
```
asutula, line 71: All this logic lives in `Makefile` now and is a more simple and reusable function to release any package as an artifact

Makefile
```diff
@@ -1,11 +1,76 @@
...
+build-hub-release: $(GOX) $(GOVVV) $(GOMPLATE)
+	$(call gen_release_files,./cmd/hub,hub,"linux/amd64 linux/386 linux/arm darwin/amd64 windows/amd64")
+.PHONY: build-hub-release
+
+build-buck-release: $(GOX) $(GOVVV) $(GOMPLATE)
+	$(call gen_release_files,./cmd/buck,buck,"linux/amd64 linux/386 linux/arm darwin/amd64 windows/amd64")
+.PHONY: build-buck-release
+
+build-releases: build-hub-release build-buck-release
+.PHONY: build-releases
```
asutula, line 73: We can add releases for `hubd` and `buckd` as well if we want, it's now trivial. Didn't do that because I figured there is a reason we're not releasing them?

Makefile
```diff
@@ -1,11 +1,76 @@
...
+	$(TXTL_BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/hubd
+.PHONY: install-hubd
+
+install-buck: $(GOVVV)
+	$(TXTL_BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/buck
+.PHONY: install-buck
+
+install-buckd: $(GOVVV)
+	$(TXTL_BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/buckd
+.PHONY: install-buckd
```
asutula, line 47: Above are a bunch of single make commands to build or install all the things.

Makefile
```diff
@@ -1,11 +1,76 @@
...
+	for release in *; do \
+		cp ../../LICENSE ../../README.md $${release}/; \
+		if [[ $${release} != *"windows"* ]]; then \
+  		TXTL_FILE=$(2) $(GOMPLATE) -f ../../dist/install.tmpl -o "$${release}/install"; \
+			tar -czvf ../dist/$${release}.tar.gz $${release}; \
+		else \
+			zip -r ../dist/$${release}.zip $${release}; \
+		fi; \
+	done
+endef
```
asutula, line 62: This is a reusable function that builds a package, copies in all the files we want to bundle in the release, and creates the tar or zip archive.

cmd/hubd/Dockerfile
```diff
@@ -31,13 +31,8 @@ RUN cd $SRC_DIR \
...
 COPY . $SRC_DIR
 
-# Setup the API buildinfo
-RUN cd $SRC_DIR \
-  && go build -modfile .bingo/govvv.mod -o=. github.com/ahmetb/govvv
-
 # Install the daemon
-RUN cd $SRC_DIR \
-  && pkg="$(go list ./buildinfo)" && govvvflags="$(./govvv -flags -pkg $pkg)" && CGO_ENABLED=0 GOOS=linux go build -ldflags="$govvvflags" -o hubd cmd/hubd/main.go
+RUN cd $SRC_DIR && TXTL_BUILD_FLAGS="CGO_ENABLED=0 GOOS=linux" make build-hubd
```
asutula, line 35: Docker uses `make` now too.

dist/install.tmpl
```diff
@@ -8,16 +8,16 @@ set -Eeuo pipefail
 
 INSTALL_DIR="$(dirname "$0")"
 
-hub="$INSTALL_DIR/hub"
+file="$INSTALL_DIR/{{ .Env.TXTL_FILE }}"
```
asutula, line 11: This is now a template file used by Gomplate

Makefile
```diff
@@ -1,11 +1,76 @@
 include .bingo/Variables.mk
 
-.DEFAULT_GOAL=textile
+.DEFAULT_GOAL=install
```
sanderpick, line 3: Should this be `build`? Considering the usual `make` and then `make install` pattern...

Makefile
sanderpick: Ah yeah, I think we decided to not promote the use of `hubd` but doing that by not releasing the artifacts doesn't sound right anymore. I vote to include them.

cmd/hub/cli/version.go
```diff
@@ -78,7 +77,7 @@ var versionCmd = &cobra.Command{
 			}
 		}
 
-		if version == "git" {
+		if version == "none" {
```
sanderpick, line 80: Can we change `version` above to be `version := bi.GitSummary`?

sanderpick: Then, no need to report "custom version". The version is what it is.

dist/install.tmpl
sanderpick: Slick

sanderpick (approved):
Slick! LGTM, with a couple questions.

Makefile
```diff
@@ -1,11 +1,76 @@
...
+install-buckd: $(GOVVV)
+	$(TXTL_BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/buckd
+.PHONY: install-buckd
+
+define gen_release_files
+	$(GOX) -osarch=$(3) -output="build/$(2)/$(2)_${TXTL_VERSION}_{{.OS}}-{{.Arch}}/$(2)" -ldflags="${GOVVV_FLAGS}" $(1)
+	mkdir -p build/dist; \
+	cd build/$(2); \
+	for release in *; do \
+		cp ../../LICENSE ../../README.md $${release}/; \
```
sanderpick, line 54: I think we can just delete this file... it's empty anyway

cmd/hub/cli/version.go
andrewxhill: just account for the 3rd case from go install. you could

* just print the GitSummary but leave the version check here. GitSummary will be "" when `go install ./...`. 
* or change the default in the buildinfo file so gitsummary default string is something other than "". but then will need to do some check still if you want to display more when the git info is available or remove that extra info so go install ./... version isn't weird empties

Makefile
asutula: Ok will do.

Makefile
asutula: Not sure. I preserved the behavior that was there before... the `textile` goal ran `go install`. Also my feeing is the traditional `make`, `make install` pattern exists when there is a dependency between the two: first you have to build, then you install the result of build. That isn't the case with the Go commands we're using. Build and install are two separate things.

asutula: But I really don't have any strong option here. I'll switch it to `build`. Curious though.. when you run `go build ./...`, it seems to just compile the code, but not generate executables anywhere. Is that correct or am I missing something?

Makefile
asutula: Which file do you mean?

cmd/hub/cli/version.go
```diff
@@ -53,7 +52,9 @@ var versionCmd = &cobra.Command{
 	Long:  `Shows the installed CLI version.`,
 	Args:  cobra.ExactArgs(0),
 	Run: func(c *cobra.Command, args []string) {
-		version := bi.Version
+		version := bi.GitSummary
```
asutula, line 55: Let me know if this is what you had in mind @andrewxhill and @sanderpick

andrewxhill: i think going in circles a bit. probably better to start from the other side and answer these three and then the solution is as simple as it looks.

1. What's the expected behavior if someone downloads from the release page, installs, and calls `version`?
2. What's the expected behavior if someone `make build` and calls `version`?
3. What's the expected behavior if someone `go install ./...` and calls `version`?

andrewxhill: sorry, `behavior => result` in those three questions

Makefile
sanderpick: Oops. README.md

cmd/hub/cli/version.go
sanderpick: Yep, that's what I was thinking @asutula. 

@andrewxhill, here's what I think:
1. Should be the result of `GitSummary` (govvv used in CI)
1. Should be the result of `GitSummary` (govvv used in local Makefile)
1. Should be the default value of `GitSummary`, perhaps "unknown"

Makefile
asutula: Oh, I'll delete the `dist/README.md` because it is empty. But this command is actually copying the repo's main `README.md` which is full of good information. Should we leave that?

sanderpick: Ah, I see... yeah nice let's keep that

cmd/hub/cli/version.go
andrewxhill: k, for #3 that's what i was getting at in my comment over the weekend. current default GitSummary is "" so needs an update.

i think #1 or #2 could be broken here

asutula: If we want 1 and 2, that is the way it works in this current revision. For 3, it will currently print `""`, then the semver parsing will fail and the resulting `Running a custom hub build...` will be displayed. So I guess the only thing to change there is to print something better than `""` for 3? We can just change the default in `buildinfo`. I actually don't mind the idea of making all the defaults in there `"git"` instead of `""`.

asutula: @andrewxhill I don't think 1 and 2 are broken, so if you see how they are, please explain.

andrewxhill: kk, is there something new to setup maybe?

Screen Shot 2020-09-21 at 9 56 39 AM

asutula: Another idea to float... In powergate, we have the equivalent of the `version` command just print out all the information from `buildinfo`. Looks like:
```
pow --version
pow build info:
        version:        git
        build date:     2020-09-21T15:01:22Z
        git summary:    v0.6.0-dirty
        git branch:     master
        git commit:     a928c57
        git state:      dirty
```
This seems like a good option to me because we don't have to make any decisions about what buildinfo value to assign to `version`. It provides a lot more context. In this case above, I can infer that the user ran a local `make` build, has some local changes on top of the `v0.6.0` tag, and I have the commit hash and all.

asutula: Oh I see. Yea, the `-dirty` in GitSummary is breaking the semantic version parser I guess. Yes, that is broken indeed. See my above comment. I think this could be a solution.

asutula: Also, I'm a little confused about what `make build` actually does. It runs `go build ./...` but I don't think that installs the binaries in you PATH. It looks like you're assuming it does. Might want to use `make install` instead just to be sure you're installing the binaries.

.github/workflows/release.yml
```diff
@@ -26,51 +26,12 @@ jobs:
...
+          TXTL_VERSION=${GITHUB_REF##*/} make build-releases
```
andrewxhill, line 31: @asutula this isn't really "broken" but we saw a case where if two tags are created for the same commit, the publish version may not match expected. i'm not certain it fixes it, but you may want `${{ github.event.release.tag_name }}` instead

asutula: Nice, that seems better.

cmd/hub/cli/version.go
andrewxhill: oh nice, read that wrong, thought it had changed. better!

andrewxhill: the semantic version parsing was really helpful so you could easily check if the user is behind production. which is only useful if it's not dirty. so if it fails you could assume it's a custom build and skip the remote check perhaps... though it still seems valuable if someone is on 0.6.0-dirty and production is on 0.7.0 to give them the warning

andrewxhill: (or at least print remote still)

Makefile
sanderpick: That's correct, `build` compiles but does not install. Actually, I think you had it right initially. A normal make install would take the result of build and just move the binary... which is not what our install does here (it could move the result of build, but that feels weird given Go has it's own install command)

cmd/hub/cli/version.go
```diff
@@ -63,34 +70,32 @@ var versionCmd = &cobra.Command{
...
+							{"Git Commit", bi.GitCommit},
+							{"Git Summary", bi.GitSummary},
+							{"Build Date", bi.BuildDate},
+						},
+					)
+				}
+			} else if current.LT(latestRelease.Version) {
 				// Display warning if outdated
-				cmd.Warn("There is a new hub release. Run `%s` to install %s.", aurora.White("hub update").Bold(), aurora.Cyan(latest.Version.String()))
+				cmd.Warn("There is a new hub release. Run %s to install %s.", aurora.White("hub update").Bold(), aurora.Cyan(latestRelease.Version.String()))
```
asutula, line 97: Ok see what you think of this @andrewxhill and @sanderpick. I didn't go as far as doing custom parsing of the git summary, but I think this represents everything we talked about in today's call.

- Always print out some version string. This will be the `buildinfo.Version` for CI release builds, `buildinfo.GitSummary` for local makefile builds, or `git` for local `go install ./...` builds.
- Always print out the API version.
- If we can parse `buildinfo.Version` as a valid version, we compare it to the latest release and possibly print out the  warning message that there is a newer version available.
- If we can't parse `buildinfo.Version` as a valid version (this is the case for local builds using `make` or `go install`) we print the "running a custom build" warning.
- If we can't parse `buildinfo.Version` as a valid version AND we have `buildinfo` set because it was a `make` build,  we print out a table showing all the build information as context.

sanderpick: Sounds good to me!

PR #361, opened 2020-09-24
Archive storage config API

### LINKED ISSUE #336: Add default storage config to bucket archive api
This would be the default that applies to all bucket archives per account/user.

### LINKED ISSUE #339: Add per bucket archive storage config option
Add ability to provide a custom storage config for a single call to `Buckets.Archive`

asutula:
Allows getting and setting the default `ArchiveConfig` per `Bucket` and allows you to pass an override `ArchiveConfig` with each call to `Archive`.

Updates everything from mongo, rpc, client, local buckets api, and cli.

closes #336 
closes #339


api/buckets/client/client.go
```diff
@@ -345,11 +345,33 @@ func (c *Client) PullPathAccessRoles(ctx context.Context, key, pth string) (map[
...
+		ArchiveConfig: config,
+	}
+	_, err := c.c.SetDefaultArchiveConfig(ctx, req)
+	return err
+}
+
 // Archive creates a Filecoin bucket archive via Powergate.
-func (c *Client) Archive(ctx context.Context, key string) (*pb.ArchiveResponse, error) {
-	return c.c.Archive(ctx, &pb.ArchiveRequest{
+func (c *Client) Archive(ctx context.Context, key string, opts ...ArchiveOption) error {
```
asutula, line 366: Added options arg to specify a custom `ArchiveConfig`. This is also reflected in the local Buckets API.

api/buckets/pb/buckets.proto
```diff
@@ -179,8 +179,43 @@ message PullPathAccessRolesResponse {
...
+    string key = 1;
+    ArchiveConfig archive_config = 2;
+}
+
+message SetDefaultArchiveConfigResponse {
+}
+
 message ArchiveRequest {
     string key = 1;
+    ArchiveConfig archive_config = 2;
```
asutula, line 218: Added the custom override config here. This can be `nil` when the default should be used.

api/buckets/pb/buckets.proto
```diff
@@ -247,6 +282,8 @@ service APIService {
     rpc PullPathAccessRoles(PullPathAccessRolesRequest) returns (PullPathAccessRolesResponse) {}
 
     // Archive
+    rpc DefaultArchiveConfig(DefaultArchiveConfigRequest) returns (DefaultArchiveConfigResponse) {}
+    rpc SetDefaultArchiveConfig(SetDefaultArchiveConfigRequest) returns (SetDefaultArchiveConfigResponse) {}
```
asutula, line 286: New methods for getting and setting the bucket default archive config.

api/buckets/service.go
```diff
@@ -59,6 +62,25 @@ var (
...
+
+	// baseArchiveStorageConfig is used as
+	baseArchiveStorageConfig = ffs.StorageConfig{
+		Hot: ffs.HotConfig{
+			Enabled: false,
+		},
+		Cold: ffs.ColdConfig{
+			Enabled: true,
+		},
+	}
```
asutula, line 74: This is the base `StorageConfig` we use to store bucket data in Powergate. Biggest thing to note here is that hot storage is disabled. When a bucket is archived, we create a `ffs.FilConfig` from a `ArchiveConfig` and set `Cold.Filecoin` value with it.

api/buckets/service.go
```diff
@@ -59,6 +62,25 @@ var (
...
+
+	// ToDo: Export the default storage config from powergate so we can create this from it.
+	defaultDefaultArchiveConfig = mdb.ArchiveConfig{
+		RepFactor:       5,
+		TrustedMiners:   []string{"t016303", "t016304", "t016305", "t016306", "t016309"},
+		DealMinDuration: powUtil.MinDealDuration * 2,
+		FastRetrieval:   true,
+		DealStartOffset: 72 * 60 * 60 / powUtil.EpochDurationSeconds, // 72hs
+	}
```
asutula, line 83: If the user hasn't set a default `ArchiveConfig` for their Bucket, this one is used. I'd like to export the default `StorageConfig` from Powergate so we can create this dynamically based on whatever is set in Powergate.

api/buckets/service.go
```diff
@@ -2003,6 +2025,47 @@ func (s *Service) PullPathAccessRoles(ctx context.Context, req *pb.PullPathAcces
...
+
+	ba, err := s.Collections.BucketArchives.GetOrCreate(ctx, req.Key)
+	if err != nil {
+		return nil, fmt.Errorf("getting bucket archive data: %s", err)
+	}
+	archiveConfig := ba.DefaultArchiveConfig
+	if archiveConfig == nil {
+		// The queried BucketArchive was created before we added .DefaultArchiveConfig, so just return the default
+		archiveConfig = &defaultDefaultArchiveConfig
```
asutula, line 2042: Just use the default specified at the top of this file. Logic of not persisting a default on the `BucketArchive` struct is that maybe we want to change the default with time. Don't want to be stuck with an old default that was persisted to mongo. We should take this approach in Powergate @jsign .

api/buckets/service.go
```diff
@@ -2003,6 +2025,47 @@ func (s *Service) PullPathAccessRoles(ctx context.Context, req *pb.PullPathAcces
...
+
+	if !s.Buckets.IsArchivingEnabled() {
+		return nil, ErrArchivingFeatureDisabled
+	}
+
+	ba, err := s.Collections.BucketArchives.GetOrCreate(ctx, req.Key)
+	if err != nil {
+		return nil, fmt.Errorf("getting bucket archive data: %s", err)
+	}
+	archiveConfig := ba.DefaultArchiveConfig
```
asutula, line 2039: Use the `ArchiveConfig` that was set by the user if it exists.

api/buckets/service.go
```diff
@@ -2003,6 +2025,47 @@ func (s *Service) PullPathAccessRoles(ctx context.Context, req *pb.PullPathAcces
...
+		return nil, ErrArchivingFeatureDisabled
+	}
+
+	ba, err := s.Collections.BucketArchives.GetOrCreate(ctx, req.Key)
+	if err != nil {
+		return nil, fmt.Errorf("getting bucket archive data: %s", err)
+	}
+
+	ba.DefaultArchiveConfig = fromPbArchiveConfig(req.ArchiveConfig)
+	err = s.Collections.BucketArchives.Replace(ctx, ba)
```
asutula, line 2062: If the user calls `SetDefaultArchiveConfig` explicitly, we do persist the `ArchiveConfig` they provide and it will be used as the default in `DefaultArchiveConfig` and `Archive` from now on.

api/buckets/service.go
```diff
@@ -2084,26 +2147,54 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
 	}
 
+	ba, err := s.Collections.BucketArchives.GetOrCreate(ctx, req.Key)
+	if err != nil {
+		return nil, fmt.Errorf("getting bucket archive data: %s", err)
+	}
+
+	var archiveConfig *mdb.ArchiveConfig
+	if req.ArchiveConfig != nil {
+		archiveConfig = fromPbArchiveConfig(req.ArchiveConfig)
```
asutula, line 2157: Use the override if it's provided in the request.

api/buckets/service.go
```diff
@@ -2084,26 +2147,54 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+	} else if ba.DefaultArchiveConfig != nil {
+		archiveConfig = ba.DefaultArchiveConfig
```
asutula, line 2159: Or use the user default.

api/buckets/service.go
```diff
@@ -2084,26 +2147,54 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+	} else {
+		// The queried BucketArchive was created before we added .DefaultArchiveConfig so just use the default default
+		// and defer saving the default until the user calls DefaultArchiveConfic or SetDefaultArchiveConfig.
+		archiveConfig = &defaultDefaultArchiveConfig
```
asutula, line 2163: Or just use the default default that is specified in code at the top of this file.

api/buckets/service.go
```diff
@@ -2084,26 +2147,54 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+	}
+
+	storageConfig := baseArchiveStorageConfig
+	storageConfig.Cold.Filecoin = toFilConfig(archiveConfig)
```
asutula, line 2167: Build the `StorageConfig` by using the base value plus a `FilConfig` created from the `ArchiveConfig`

api/buckets/service.go
```diff
@@ -2084,26 +2147,54 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+
+	if storageConfig.Cold.Filecoin.Addr == "" {
+		// We don't have an address to use, which is the case for a BucketArchive.DefaultArchiveConfig
+		// that is the default value, get the default address from the FFS instance
+		storageConfig.Cold.Filecoin.Addr = defConf.Cold.Filecoin.Addr
```
asutula, line 2172: The default `ArchiveConfig` doesn't include any address information, and if the user sets their own Bucket default `ArchiveConfig`, it may not have an address. In that case, grab the default address from the FFS instance. This is nice because it can change if the user wants. 

Long term, I'm working on a Powergate branch where default address in managed separately from default config, and address can be provided as an option in `PushConfig`, `Replace`, etc. That will make this cleaner.

api/buckets/service.go
```diff
@@ -2131,18 +2214,46 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+				jid, err = s.PGClient.FFS.PushStorageConfig(ctxFFS, p.Cid(), powc.WithStorageConfig(storageConfig), powc.WithOverride(true))
 				if err != nil {
 					return nil, fmt.Errorf("pushing config: %s", err)
 				}
 			default:
 				return nil, fmt.Errorf("unexpected current archive status: %d", ba.Archives.Current.JobStatus)
 			}
 		} else { // Case 2.
-			jid, err = s.PGClient.FFS.Replace(ctxFFS, oldCid, p.Cid())
+			res, err := s.PGClient.FFS.GetStorageConfig(ctxFFS, oldCid)
```
asutula, line 2225: Logic here, in the case of a second (or more) archive, is to get the current `StorageConfig` and compare it to the new `StorageConfig`. If they are the same, use `Replace`. If not, use `Remove` and `PushConfig`.

api/buckets/service.go
```diff
@@ -2131,18 +2214,46 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+			var oldStorageConfig *ffs.StorageConfig
+			if res.Config != nil {
+				oldStorageConfig = &ffs.StorageConfig{
+					Repairable: res.Config.Repairable,
+					Cold:       fromRPCColdConfig(res.Config.Cold),
+					Hot:        fromRPCHotConfig(res.Config.Hot),
+				}
 			}
+
+			if reflect.DeepEqual(&storageConfig, oldStorageConfig) {
```
asutula, line 2239: Here's that comparison. Cool to use `reflect`?

buckets/local/archive.go
```diff
@@ -7,21 +7,156 @@ import (
...
+
+func WithArchiveConfig(config ArchiveConfig) ArchiveRemoteOption {
+	return func(opts *archiveRemoteOptions) {
+		opts.archiveConfig = &config
+	}
+}
+
 // ArchiveRemote requests an archive of the current remote bucket.
-func (b *Bucket) ArchiveRemote(ctx context.Context) error {
+func (b *Bucket) ArchiveRemote(ctx context.Context, opts ...ArchiveRemoteOption) error {
```
asutula, line 140: Adds option for override `ArchiveConfig`

buckets/local/archive.go
```diff
@@ -117,17 +252,6 @@ func (b *Bucket) ArchiveStatus(ctx context.Context, watch bool) (<-chan ArchiveS
 	return msgs, nil
 }
 
-func isJobStatusFinal(status pb.ArchiveStatusResponse_Status) (bool, error) {
```
asutula, line 120: This wasn't used.

go.mod
```diff
@@ -39,42 +36,32 @@ require (
...
-	github.com/spf13/cast v1.3.1 // indirect
 	github.com/spf13/cobra v1.0.0
-	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.6.1
 	github.com/textileio/dcrypto v0.0.1
 	github.com/textileio/go-assets v0.0.0-20200430191519-b341e634e2b7
 	github.com/textileio/go-threads v1.0.0
-	github.com/textileio/powergate v0.6.0-rc2
+	github.com/textileio/powergate v0.6.1
```
asutula, line 58: Note: updated to newish version I think is being deployed everywhere

mongodb/bucketarchives.go
```diff
@@ -8,8 +8,9 @@ import (
 )
 
 type BucketArchive struct {
-	BucketKey string   `bson:"_id"`
-	Archives  Archives `bson:"archives"`
+	BucketKey            string         `bson:"_id"`
+	Archives             Archives       `bson:"archives"`
+	DefaultArchiveConfig *ArchiveConfig `bson:"default_archive_config"`
```
asutula, line 13: This will be `nil` until the user sets it. We take advantage of that fact to detect `nil` and use the hard coded default instead.

asutula:
Ok getting a legit error in the integration test `TestArchiveBucketWorkflow`. The error is:
```
executing cold-storage config: making deal configs: getting miners from minerselector: not enough miners that satisfy the constraints
```
@jsign just checking if there is something we need to change about the lotus devnet and/or powergate setup in here? Anything come to mind?

asutula:
Updated the compose file to powergate 0.6.6 and lotus-devnet sha-35afaa3 and now the error is:
```
executing cold-storage config: all proposals were rejected
```

jsign:
> Ok getting a legit error in the integration test `TestArchiveBucketWorkflow`. The error is:
> 
> ```
> executing cold-storage config: making deal configs: getting miners from minerselector: not enough miners that satisfy the constraints
> ```
> 
> @jsign just checking if there is something we need to change about the lotus devnet and/or powergate setup in here? Anything come to mind?

Yes, most probably is related to forcing Peter's miners in the version of Powergate you've updated. Now that in v0.6.6 we switched to SR2 miner selector by default, might make sense again to have the original _StorageConfig_ which doesn't have a replication factor of 5, which is my best guess about that error. 
So the idea would be:
- Revert to the original _StorageConfig_ with rep facto 1 and without trusted miners.
- Set explicitly for tests here to use `--ffsminerselector=reputation`, since now the default is `sr2`.

I can confirm this theory on Monday, but makes sense to me.

asutula:
Thanks @jsign. That gives me a path forward. Will dig in more later.

api/buckets/service.go
```diff
@@ -2131,18 +2213,50 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+				// Old storage config is the same as the new so use replace.
+				jid, err = s.PGClient.FFS.Replace(ctxFFS, oldCid, p.Cid())
+				if err != nil {
+					return nil, fmt.Errorf("replacing cid: %s", err)
+				}
+			} else {
+				// New storage config, so remove and push.
+				_, err = s.PGClient.FFS.PushStorageConfig(ctxFFS, oldCid, powc.WithStorageConfig(ffs.StorageConfig{}), powc.WithOverride(true))
```
asutula, line 2246: @jsign I just want to check that this is ok. I have to disable hot and cold storage before calling `Remove`. I do that here, just ignore the jobId, and immediately call `Remove`. This works when I run the tests, but just want to make sure it's not just luck. Do I need to watch the job and wait until it is successful before calling `Remove`?

jsign (approved):
LGTM! I left some questions in the wild.

api/buckets/service.go
asutula: Correct. The `ArchiveConfig` that they can edit maps to `StorageConfig.Cold.FilConfig` so that is the only part of the `StorageConfig` they can effectively edit. They can't edit the `Hot` settings at all. `Hot` settings are exclusively controlled right here.

api/buckets/service.go
```diff
@@ -59,6 +62,24 @@ var (
...
+		Hot: ffs.HotConfig{
+			Enabled: false,
+		},
+		Cold: ffs.ColdConfig{
+			Enabled: true,
+		},
+	}
+
+	// ToDo: Export the default storage config from powergate so we can create this from it.
+	defaultDefaultArchiveConfig = mdb.ArchiveConfig{
```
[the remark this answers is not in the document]
asutula: Yea I understand that. I decided to just build the long term solution here and figured we could discuss if/what to change in the short term. Soooo....

Should we remove `RepFactor`, `ExcludedMiners`, `TrustedMiners`, and `CountryCodes` from `ArchiveConfig` for now?

api/buckets/service.go
asutula: No very strong option here either. It made sense in my mind for `ArchiveConfig` just because it is a new concept and it feels good to not commit too much to some default that is persisted and force us into having to do a data migration soon when things change. We can discuss further for Powergate if it ever seems important.

api/buckets/service.go
asutula: I had read that `cmp.Equal` is meant for testing and panics under some circumstances. Not sure if that is true, didn't dig much deeper.

cmd/buck/cli/archive.go
```diff
@@ -1,18 +1,74 @@
...
+	"io"
+	"os"
 
 	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
 	"github.com/textileio/textile/v2/buckets/local"
 	"github.com/textileio/textile/v2/cmd"
 )
 
+var defaultArchiveConfigCmd = &cobra.Command{
```
sanderpick, line 16: Just noticing now that these archive commands (the older ones as well) as missing the `Args` param, which is handy to restrict args, e.g., `Args: cobra.ExactArgs(0),`, `Args: cobra.RangeArgs(1, 2),`, `Args:  cobra.MaximumNArgs(1),`, etc. It also shows in the `--help` output.

sanderpick (approved):
LGTM! I left a bunch of nits, but otherwise looks solid.

api/buckets/service.go
```diff
@@ -2131,18 +2213,50 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+			}
+
+			if reflect.DeepEqual(&storageConfig, oldStorageConfig) {
+				// Old storage config is the same as the new so use replace.
+				jid, err = s.PGClient.FFS.Replace(ctxFFS, oldCid, p.Cid())
+				if err != nil {
+					return nil, fmt.Errorf("replacing cid: %s", err)
+				}
+			} else {
+				// New storage config, so remove and push.
```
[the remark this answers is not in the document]
asutula: I think that for the end state that you want to exist the answer is: yes, we are on the same page about that. Hot storage will be disabled.

Just to clear some details up, in the scenario you describe, `oldStorageConfig` won't be `nil`, it will be whatever the previous bucket archive used, probably has hot storage enabled, etc. The new `storageConfig` will likely be different because it must have hot storage disabled, and likely has other differences because of the new default settings. So for that reason, we will `Remove` and `Push` the new storage config with hot storage disabled.

api/buckets/service.go
asutula: It is disabled once this PR is merged in. But for previously archived buckets that are archiving again once this PR is merged (the situation we discuss above), hot storage may be enabled and will be disabled by this `PushStorageConfig`

integrationtest/pg/pg_test.go
```diff
@@ -102,62 +102,62 @@ func TestArchiveTracker(t *testing.T) {
 }
 
 func TestArchiveBucketWorkflow(t *testing.T) {
-	util.RunFlaky(t, func(t *util.FlakyT) {
```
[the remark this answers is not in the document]
asutula: Oops, did that to speed up local test debugging, will re-enable.

api/buckets/service.go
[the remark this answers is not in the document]
jsign: I'm not sure, so I'd leave that up to you. :)

api/buckets/service.go
jsign: `cmp.Equal` was created as a better solution than `reflect.DeepEqual` which compares unexported fields, and might not behave correctly on particular types which its equality might not be easily comparable by bytes

I'd say both are meant for a quick way of comparing things. The full way is manual.

api/buckets/service.go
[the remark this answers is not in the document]
jsign: 👍🏼

integrationtest/pg/docker-compose.yml
```diff
@@ -34,13 +35,11 @@ services:
   ipfsbuckets:
     image: ipfs/go-ipfs:v0.6.0
     environment:
-     - IPFS_PROFILE=local-discovery
+      - IPFS_PROFILE=local-discovery
```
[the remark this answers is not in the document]
asutula: I think it was missing a space before so this is a formatting fix

api/buckets/service.go
asutula: Cool, switched to `cmd.Equal`

PR #365, opened 2020-09-26
Remove WS ping and add some requested buck flags

### LINKED ISSUE #350: Add a --quiet option to hub buck push
I'm doing some scripting with the cli and the default output is quite verbose, it'd be great if there was a "quiet" option, similar to in go-ipfs that only outputs the final cid, skipping progress bars etc.

There's probably other commands where this would be useful too but push is the one where I'd find it most useful at the moment.

### LINKED ISSUE #351: -y flag for hub buck archive command
Similar to the `-y/--yes` functionality to skip the confirmation prompt in `hub buck push` it'd be great to be able to pass `-y/--yes` to `hub buck archive` to agree to the prompt for scripting.

### LINKED ISSUE #355: ListPath and ListIPFSPath return response closed without headers
`Response closed without grpc-status (Headers only)`

on some buckets, particularly very large buckets, the response to listpath will never succeed. 

added private details in slack

sanderpick:
Closes #355 
Closes #350
Closes #351

### DIFF
--- a/cmd/buck/cli/archive.go
+++ b/cmd/buck/cli/archive.go
@@ -14,13 +14,17 @@ var archiveCmd = &cobra.Command{
 	Short: "Create a Filecoin archive",
 	Long:  `Creates a Filecoin archive from the remote bucket root.`,
 	Run: func(c *cobra.Command, args []string) {
-		cmd.Warn("Archives are currently saved on an experimental test network. They may be lost at any time.")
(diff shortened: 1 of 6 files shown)


jsign (approved):
Nice, useful for scripting.

andrew:
🙇‍♂️

jimmylee:
🙇‍♂️

Issue #373, opened 2020-10-05
Account Billing


sanderpick:
The model here is metered billing beyond a free quota. Like AWS, I’m imagining an invoice that enumerates usage by product, but where all products pull from the same quota.

#### ThreadDB

The database layer is used by other products, but applications can also use it directly.

Things to charge for:

- Number of reads
- Number of writes
- Stored data
- Network egress

Things that need limiting: 

- Number of threads
- Number of collections
- Number of records per collection
- Number of indexes per collection
- Max instance size
- Number of real-time listeners

#### Buckets

Buckets are built in part on ThreadDB and will accrue usage to the ThreadDB quota. Buckets are also built on IPFS (UnixFS) and will accrue associated data storage and network egress charges.

Things to charge for:

- Stored data
- Network egress
- Number of archives

Things that need limiting: 

- Max bucket size

#### Mail

Mailboxes are built in part on ThreadDB and will also accrue usage to the ThreadDB quota.

Things to charge for:

- Number of sent messages

Things that need limiting: 

- Messages sent per day

### Free Quota

Product | Price
-- | --
Stored data (ThreadDB, Buckets) | 5 GiB
Document reads (ThreadDB) | 10,000 per day
Document writes (ThreadDB) | 5,000 per day
Network egress (ThreadDB, Buckets) | 10 GiB per month

#### Beyond the Free Quota

Product | Price
-- | --
Stored data | $0.2/GiB/month
Document reads | $0.1 per 100,000 documents
Document writes | $0.2 per 100,000 documents
Network egress | $0.1/GiB

Issue #315, opened 2020-08-12
Consolidate Account and User


sanderpick:
`User` can become another `AccountType`

PR #374, opened 2020-10-05
Billing, Take 1

### LINKED ISSUE #276: Re-consider user quotas
From a slack convo:

> Maybe we need to rethink the users max threads limit. Right now, users get 100 threads just like accounts. But any reasonable user limit per account would be large (5-10,000), which means that one account gets a huge threads quota. Maybe the best thing is take user threads out of the account’s quota, and just bump that up to 1000 or so for the free tier.

sanderpick:
Closes #276
Addresses #373 in part (the limits are not all in place)

This PR required a number of big structural changes, which is why the diff is so huge. I've commented on the bits that are worth looking at.

Understanding this requires some knowledge of how Stripe works with recurring metered subscriptions. 
- A subscription is made of _prices_, which define what a _product_ costs at different _tiers_ and at what intervals.
- A price can only refer to one product, but a subscription can have multiple prices.
- There are various setting for how [prices](https://stripe.com/docs/api/prices) are charged.

The Hub has four products:
- _Stored data_: 5 Gib free, this is a static value, i.e, it's carried from day to day.
- _Network egress_: 500 Mib per day, time-based, i.e., it's reset every day
- _Instance reads_: 10,000 per day, time-based, i.e., it's reset every day
- _Instance writes_: 5,000 per day, time-based, i.e., it's reset every day

Stripe only cares about _units_ of a product. I derived the current unit values by considering that we should only notify Stripe about usage at the 1 cent resolution. For example, the smallest unit of storage we have is 5Gib / 100, which costs 1 cent. The first 100 of which are free, amounting to 5Gib free. The same logic is used for the other products.

Note: Currently, this functionality is only available at the developer level. After https://github.com/textileio/textile/issues/315, I'll extend usage billing to _their_ users.

The life-cycle of a customer and the various states they can arrive in are important to understand.
- A new customer is created when an account is created (dev/org)
- That customer is auto-subscribed to the Hub usage subscription
- The account goes about business as usual, acruing usage against their free quota
- Once they hit the ceiling on any product _price_, the CLI will error and let the user know they can run `hub billing portal` to launch the [customer portal](https://stripe.com/docs/billing/subscriptions/integrating-customer-portal) where a payment method can be added.
- If they do that, Stripe hits the billing gateways webhook handler with an event that we use to update our customer object
- This unlocks the account, and they can use however much Hub as they want to pay for
- They can also cancel the subscription in the customer portal, which we also handle via a webhook, which locks the account completely
- The can re-create the subscription by running `hub billing setup`
- In this case, stored data is not reset, but egress, reads, and writes are reset
- A clever user may cancel and re-subscribe over and over to get more egress, reads, writes... something we can work on stopping at some point, but those are reset daily so doesn't feel super important right now
- At any point, a user can run `hub billing status` to see relevant usage and status info, like...
```
$ hub billing status
> Account status: pay-as-you-go
> Account balance: $0.00
> Subscription status: active

                          USAGE  FREE QUOTA         START      END
  Stored data (bytes)     0      5368709100 (100%)  03-Nov-20  04-Nov-20
  Network egress (bytes)  406    429496322 (100%)   03-Nov-20  04-Nov-20
  ThreadDB reads          0      10000 (100%)       03-Nov-20  04-Nov-20
  ThreadDB writes         0      5000 (100%)        03-Nov-20  04-Nov-20
```
- If an account is deleted, the customer object is retained since it may still have a balance with open invoices, etc.

Here's what the customer portal looks like. I added the logo and our color scheme.

Screen Shot 2020-11-04 at 3 22 33 PM
Screen Shot 2020-11-04 at 3 22 57 PM


api/billingd/service/service.go
```diff
@@ -0,0 +1,708 @@
...
+		},
+	})
+	if r.Err() != nil {
+		return nil, r.Err()
+	}
+	var pre Customer
+	if err := r.Decode(&pre); err != nil {
+		return nil, err
+	}
+	if pre.Period.End < req.Period.End {
```
sanderpick, line 430: If the update includes a period change, we reset the time-based usage stats.

api/billingd/service/service.go
```diff
@@ -0,0 +1,708 @@
...
+		usage.Total,
+		usage.Free,
+	)
+	return &pb.IncStoredDataResponse{
+		Period:     periodToPb(doc.Period),
+		StoredData: usage,
+	}, nil
+}
+
+func (s *Service) handleUsage(
```
sanderpick, line 545: This is the logic that handles incrementing a usage value and when to fire off an request to stripe.

api/billingd/service/service.go
```diff
@@ -0,0 +1,708 @@
...
+	total := usage.Total + inc
+	if total < 0 {
+		total = 0
+	}
+	update := bson.M{key + ".total": total}
+	units := int64(math.Round(float64(total) / float64(unitSize)))
+	if units > freeUnits && !billable {
+		return nil, common.ErrExceedsFreeUnits
+	}
+	if units != usage.Units {
```
sanderpick, line 562: If units changed, inform stripe.

core/core.go
```diff
@@ -366,12 +380,17 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
+				powInterceptor(netServiceName, allowedPowMethods[netServiceName], netServiceDesc, powStub, t.pc, t.collections),
+				powInterceptor(ffsServiceName, allowedPowMethods[ffsServiceName], ffsServiceDesc, powStub, t.pc, t.collections),
+				powInterceptor(walletServiceName, allowedPowMethods[walletServiceName], walletServiceDesc, powStub, t.pc, t.collections),
+			),
+			grpcm.WithStreamServerChain(
+				auth.StreamServerInterceptor(t.authFunc),
+				streamServerInterceptor(t.preUsageFunc, t.postUsageFunc),
 			),
-			grpcm.WithStreamServerChain(auth.StreamServerInterceptor(t.authFunc)),
+			grpc.StatsHandler(&StatsHandler{t: t}),
```
sanderpick, line 393: Add a stats handler that track egress, reads, and writes.

core/core.go
```diff
@@ -505,430 +529,6 @@ func (t *Textile) HostID() peer.ID {
 	return t.ts.Host().ID()
 }
 
-func (t *Textile) authFunc(ctx context.Context) (context.Context, error) {
```
sanderpick, line 508: Interceptors now live in their own files.

scripts/gen_js_protos.bash
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
sanderpick, line 1: Script for generating the JS proto files. It's gets pretty manual at the end because the hub case needs to be treated differently (requires billing protos as well).

scripts/protoc_gen_plugin.bash
```diff
@@ -0,0 +1,89 @@
...
+  PROTOC_FLAGS+=("--proto_path=${proto_path}")
+done
+PROTOC_FLAGS+=("--${PLUGIN_NAME}_out=${PLUGIN_OUT}")
+if [ -n "${PLUGIN_OPT}" ]; then
+  PROTOC_FLAGS+=("--${PLUGIN_NAME}_opt=${PLUGIN_OPT}")
+fi
+
+mkdir -p "${PLUGIN_OUT}"
+for proto_path in "${PROTO_PATHS[@]}"; do
+  for dir in $(find "${proto_path}" -type f ! -path './buildtools/*' ! -path '*/node_modules/*' -name '*.proto' -print0  | xargs -0 -n1 dirname | sort | uniq); do
```
sanderpick, line 85: All copied from Powergate minus this change that excludes proto file in `node_modules`.

scripts/publish_js_protos.bash
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
sanderpick, line 1: Publish all JS protos with a given version and npm token.

mongodb/accounts.go
```diff
@@ -29,17 +29,17 @@ func init() {
...
-	PowInfo          *PowInfo
+	Type       AccountType
+	Key        thread.PubKey
+	Secret     thread.Identity
+	Name       string
+	Username   string
+	Email      string
+	Token      thread.Token
+	Members    []Member
+	CustomerID string
```
sanderpick, line 40: `Account` only needs to track a `CustomerID` that is used with the billing API.

mongodb/accounts.go
```diff
@@ -29,17 +29,17 @@ func init() {
...
 type Account struct {
-	Type             AccountType
-	Key              thread.PubKey
-	Secret           thread.Identity
-	Name             string
-	Username         string
-	Email            string
-	Token            thread.Token
-	Members          []Member
-	BucketsTotalSize int64
```
sanderpick, line 40: This is no longer needed here. That means existing accounts will not be penalized existing stored data, but I think that's the cleaner approach anyway, rather than immediately demanding payment.

core/usage_interceptor.go
```diff
@@ -0,0 +1,157 @@
...
+		wrapped.WrappedContext = newCtx
+		err = handler(srv, wrapped)
+		if err != nil {
+			return err
+		}
+		return post(newCtx, info.FullMethod)
+	}
+}
+
+func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Context, error) {
```
sanderpick, line 52: This servers two functions:
- Check for billing status and lock the customer out if needed.
- Check for egress, reads, and writes beyond free quota (see comments in `stats_handler` for more about this).
- Attach data storage usage info the the context that will be used by buckets internally.

core/usage_interceptor.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	cus, err := t.bc.GetCustomer(ctx, account.CustomerID)
+	if err != nil {
+		return ctx, err
+	}
+	if err := common.StatusCheck(cus.Status); err != nil {
+		return ctx, status.Error(codes.FailedPrecondition, err.Error())
+	}
+	if !cus.Billable && cus.NetworkEgress.Free == 0 {
```
sanderpick, line 81: No payment method and free quota exhausted? Error.

core/usage_interceptor.go
```diff
@@ -0,0 +1,157 @@
...
+	// @todo: Account for users after User -> Account migration
+	if account == nil || account.CustomerID == "" {
+		return ctx, nil
+	}
+
+	cus, err := t.bc.GetCustomer(ctx, account.CustomerID)
+	if err != nil {
+		return ctx, err
+	}
+	if err := common.StatusCheck(cus.Status); err != nil {
```
sanderpick, line 78: As mentioned in other places, this will report an error if the customer status is not in good standing.

core/usage_interceptor.go
```diff
@@ -0,0 +1,157 @@
...
+		"/threads.pb.API/WriteTransaction":
+		if !cus.Billable && cus.InstanceWrites.Free == 0 {
+			err = fmt.Errorf("threaddb writes exhausted: %v", common.ErrExceedsFreeUnits)
+			return ctx, status.Error(codes.ResourceExhausted, err.Error())
+		}
+	}
+	return ctx, nil
+}
+
+func (t *Textile) postUsageFunc(ctx context.Context, method string) error {
```
sanderpick, line 125: After all is said and done with this request, bump the stored data from the context.

core/stats_handler.go
```diff
@@ -0,0 +1,203 @@
...
+type StatsHandler struct {
+	t *Textile
+}
+
+var _ stats.Handler = (*StatsHandler)(nil)
+
+var statsTimeout = time.Second * 10
+
+// HandleRPC accounts for customer usage across services.
+func (h *StatsHandler) HandleRPC(ctx context.Context, st stats.RPCStats) {
```
sanderpick, line 22: `StatsHandler` allows you to track data in it's own context (Note: which is not available to other interceptors). However, it does _not_ allow you to return an error, i.e., stop the request. So, really we can just use this for tracking purposes... errors have to be returned from with the request handler or an interceptor. 

So then, what's the point of this? It's the only way we can collect wire message size info for tracking egress. The steam interceptors might sound appealing for this, but you'd have to pop the message from the stream before letting it continue to handlers.  

It's also used here to track reads and writes since we have access to the payloads for unary and stream handlers.

The downside of this approach is that we can't (currently) protect against a request the will _for the first time_ cause a free quota overage. That has to be handled _on the next request_. Keep in mind, this is only the case for egress, reads, and writes. Data storage is handled differently. I have an idea about how to enable this fail-fast behavior for egress, but will save that for next PR.

core/stats_handler.go
```diff
@@ -0,0 +1,203 @@
...
+	if org, ok := mdb.OrgFromContext(ctx); ok {
+		account = org
+	} else if dev, ok := mdb.DevFromContext(ctx); ok {
+		account = dev
+	}
+	// @todo: Account for users after User -> Account migration
+	if account == nil || account.CustomerID == "" {
+		return ctx
+	}
+	return context.WithValue(ctx, statsCtxKey("requestStats"), &requestStats{
```
sanderpick, line 178: Adds a wrapper to the context for tracking stats for the lifetime of the request (chunks from streams, etc.)

sanderpick (commented):
I added some comments to the bits that are worth looking at.

andrewxhill:
Nice! great work dude. just a couple questions 

> Once they hit the ceiling on any product price, the CLI will error and let the user know they can run hub billing portal to launch the customer portal where a payment method can be added.

* Can we flag this for later enhancement to enable a less abrupt outage. I've seen plans start accruing a bill and giving the user 1-week to enable billing and pay or then error the APIs. This would allow a free app to gracefully upgrade if they didn't expect to become popular over night and were using some API that could get blocked. So hit ceiling price -> enter grace period -> pay or get blocked.

> In this case, stored data is not reset, but egress, reads, and writes are reset

* What's your thinking for stored data no longer paid/maintained longer-term?

sanderpick:
>  Once they hit the ceiling on any product price, the CLI will error and let the user know they can run hub billing portal to launch the customer portal where a payment method can be added.
>>  Can we flag this for later enhancement to enable a less abrupt outage. I've seen plans start accruing a bill and giving the user 1-week to enable billing and pay or then error the APIs. This would allow a free app to gracefully upgrade if they didn't expect to become popular over night and were using some API that could get blocked. So hit ceiling price -> enter grace period -> pay or get blocked.

Nice, that sounds good. Should be relatively quick to add that feature. I'll make it an issue.

>  In this case, stored data is not reset, but egress, reads, and writes are reset
>>  What's your thinking for stored data no longer paid/maintained longer-term?

This just means we still account for stored data since that would be the easiest one to game. Since it's not reset, a user would not be able to add 5Gib, delete subscription, re-subscribe, add another 5 Gib.

core/stats_handler.go
carsonfarmer: What's the worst case here though, really? Request comes in, bumps quotas, and then the next one comes in... worst case isn't wild right?

core/stats_handler.go
```diff
@@ -0,0 +1,203 @@
...
+		// Don't charge for usage when customer is trying to add a payment method, cancel subscription, etc.
+		case *hpb.GetBillingSessionResponse:
+			egress = 0
+
+		// Account for threaddb reads and writes
+		case *tpb.CreateReply:
+			if pl.InstanceIDs != nil {
+				writes = int64(len(pl.InstanceIDs))
+			}
+		case *tpb.WriteTransactionReply_CreateReply:
```
carsonfarmer, line 52: Oh interesting, of course this makes sense. But it also makes me realize, we have to be pretty careful with our clients now, so that we're "optimizing" calls on behalf of the user/developer.

core/stats_handler.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+		case *tpb.DeleteReply:
+			if pl.TransactionError == "" {
+				writes = 1
+			}
+		case *tpb.WriteTransactionReply_DeleteReply:
+			if pl.DeleteReply.TransactionError == "" {
+				writes = 1
+			}
+		case *tpb.HasReply:
```
carsonfarmer, line 80: Honest question: Is there a reason you don't just "stack" all these cases and fall through? Is it just because the transaction errors are typed like this?

core/thread_interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+				return nil, err
+			}
+		default:
+			// If we're dealing with an existing thread, make sure that the owner
+			// owns the thread directly or via an API key.
+			threadID, ok := common.ThreadIDFromContext(ctx)
+			if ok {
+				th, err := t.collections.Threads.Get(ctx, threadID, owner)
+				if err != nil && errors.Is(err, mongo.ErrNoDocuments) {
+					// Allow non-owners to interact with a limited set of APIs.
```
carsonfarmer, line 89: Nice

core/usage_interceptor.go
```diff
@@ -0,0 +1,157 @@
...
+		account = org
+	} else if dev, ok := mdb.DevFromContext(ctx); ok {
+		account = dev
+	}
+	// @todo: Account for users after User -> Account migration
+	if account == nil || account.CustomerID == "" {
+		return nil
+	}
+	switch method {
+	case "/api.bucketsd.pb.APIService/Create",
```
carsonfarmer, line 143: Having string literals here and there like this seems error prone? Possible to just group these as consts or something?

mongodb/accounts.go
carsonfarmer: Fantastic, this was my number one question with all this stuff, and you covered it here. This is a good call.

carsonfarmer (approved):
Wow @sanderpick, this is a biggie. I've tried to give at least a cursory glance over everything. Obviously I haven't dug deep into any of the possible edge cases etc, but your higher level comments were super helpful in assessing the overall goals. Two of my primary "concerns" (what to do about existing data, and user-scoped allocations - to come later), are pretty well addressed here (or there are concrete plans to address them.

To follow up on one of Andrew's comments though: What about the following scenario:

Developer creates account, accrues data storage up to free amount and then doesn't upgrade/create account. They leave. Now what? We don't want to hold onto that data forever right? Do we imagine a time limit on keeping it around? Maybe a notification that it will be cleaned out? In theory, they might re-create their account (which you've covered), and would expect their data to still be there for a time.

cmd/buck/cli/push.go
```diff
@@ -45,7 +45,8 @@ var pushCmd = &cobra.Command{
 		size, err := buck.LocalSize()
 		cmd.ErrCheck(err)
 		if size > maxSize*MiB {
-			cmd.Fatal(fmt.Errorf("the bucket size is %dMB which is bigger than accepted limit %dMB", size/MiB, maxSize))
+			cmd.Fatal(fmt.Errorf("bucket size exceeds default --maxsize limit (%dMB): %dMB",
```
jsign, line 48: Nit: `s/MB/MiB`?

asutula (approved):
Wow, not sure what to say other than this is a huge amount of really great work. The high level design and thinking is sound, and the low level details are too many and too complex to have any real feedback, so let's get this thing deployed and see what happens! Seriously awesome work.

core/thread_interceptor.go
```diff
@@ -0,0 +1,193 @@
...
+			owner = dev.Key
+		} else if user, ok := mdb.UserFromContext(ctx); ok {
+			owner = user.Key
+		}
+
+		var newID thread.ID
+		var isDB bool
+		var err error
+		switch method {
+		case "/threads.pb.API/NewDB":
```
jsign, line 52: Unrleated of this PR, idea: we rely on gRPC APIs as strings. If we somewhat rename a gRPC identifier, many of our things will fail silently. Not only here, but other files and also other of our projects.

I'm not aware if there's a dynamic way to get the name using the generated code of the client.

jsign (approved):
Tons of work! :tada:
Left some comments that you might be interested.

core/stats_handler.go
sanderpick: Not wild, but somebody could accrue a bunch of egress by pulling a huge file, which caps at the max bucket size of 4Gib. Not too bad but also not negligible.

api/bucketsd/pb/javascript/package.json
```diff
@@ -9,17 +9,17 @@
   "author": "Textile",
   "license": "MIT",
   "files": [
-    "buckets_pb.js",
-    "buckets_pb_service.js",
-    "buckets_pb.d.ts",
-    "buckets_pb_service.d.ts"
+    "api/bucketsd/pb/bucketsd_pb.js",
```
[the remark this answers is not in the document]
sanderpick: Ah right, didn't occur to me would require a change. Let me know if I can help with anything there.

api/bucketsd/service.go
```diff
@@ -368,61 +358,119 @@ func encryptData(data, currentKey, newKey []byte) ([]byte, error) {
...
+
+// addPinnedBytes adds the provided delta to a running total for context.
+func (s *Service) addPinnedBytes(ctx context.Context, delta int64) context.Context {
+	total, _ := ctx.Value(ctxKey("pinnedBytes")).(int64)
+	ctx = context.WithValue(ctx, ctxKey("pinnedBytes"), total+delta)
+	owner, ok := buckets.BucketOwnerFromContext(ctx)
+	if ok && owner.StorageAvailable != -1 {
+		owner.StorageUsed += delta
+		owner.StorageAvailable -= delta
+		owner.StorageDelta += delta
```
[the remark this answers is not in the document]
sanderpick: Yep, exactly. `StorageUsed` is their total used storage. The `StorageDelta` is the diff from the current request.

api/billingd/service/service.go
sanderpick: The period changes whenever the billing cycle ends, which is currently every day. So, this should happen each day.

api/billingd/service/service.go
```diff
@@ -0,0 +1,708 @@
...
+func (s *Service) handleUsage(
+	ctx context.Context,
+	customerID string,
+	usage Usage,
+	key string,
+	inc, unitSize, freeUnits int64,
+	billable bool,
+) (*pb.Usage, error) {
+	total := usage.Total + inc
+	if total < 0 {
```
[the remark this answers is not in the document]
sanderpick: Not in theory, but seems safer than reporting negative values if we end up with a pinned bytes bug somewhere.

cmd/buck/cli/cli.go
```diff
@@ -39,7 +39,6 @@ func Init(baseCmd *cobra.Command) {
 	pushCmd.Flags().BoolP("force", "f", false, "Allows non-fast-forward updates if true")
 	pushCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
 	pushCmd.Flags().BoolP("quiet", "q", false, "Write minimal output")
-	pushCmd.Flags().Int64("maxsize", buckMaxSizeMiB, "Max bucket size in MiB")
```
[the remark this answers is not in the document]
sanderpick: Oop, this diff is outdated... I re-added this client-side max size restriction

core/stats_handler.go
sanderpick: Yep, for sure.

core/stats_handler.go
sanderpick: Here, `pl` is `interface{}` unless it's in one of the switch statements... so, each one is needed in order to pull params out of the payloads.

core/usage_interceptor.go
sanderpick: Yep, good call. Will make a note to change that.

core/core.go
```diff
@@ -177,11 +187,8 @@ type Config struct {
 	EmailAPIKey        string
 	EmailSessionSecret string
 
-	BucketsMaxSize            int64
-	BucketsTotalMaxSize       int64
-	BucketsMaxNumberPerThread int
-
-	ThreadsMaxNumberPerOwner int
+	MaxBucketSize            int64
```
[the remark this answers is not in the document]
sanderpick: Yep, that's a setting _per bucket_, which can't be controlled by billing.

api/billingd/service/service.go
```diff
@@ -0,0 +1,735 @@
...
+		case s.config.InstanceWritesPriceID:
+			cus.InstanceWrites = Usage{ItemID: item.ID}
+		}
+	}
+	return nil
+}
+
+func (s *Service) GetCustomer(ctx context.Context, req *pb.GetCustomerRequest) (
+	*pb.GetCustomerResponse, error) {
+	lck := s.semaphores.Get(customerLock(req.CustomerId))
```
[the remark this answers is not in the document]
sanderpick: Yep, a nice utility added by Anton at Anytype in `go-threads`

api/billingd/gateway/gateway.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+	return nil
+}
+
+const webhookMaxBodyBytes = int64(65536) // 64 KiB
+
+// webhookHandler handles stripe webhook events
+func (g *Gateway) webhookHandler(c *gin.Context) {
+	c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, webhookMaxBodyBytes)
```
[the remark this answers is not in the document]
sanderpick: Yep, I think it uses the writer to close the request with a "message too large" response.

api/billingd/gateway/gateway.go
```diff
@@ -0,0 +1,193 @@
...
+			log.Errorf("updating customer subscription: %v", err)
+			c.Status(http.StatusBadRequest)
+			return
+		}
+		log.Debugf("%s subscription was updated", sub.ID)
+	default:
+		log.Debugf("unhandled event type: %s\n", event.Type)
+	}
+
+	c.Status(http.StatusOK)
```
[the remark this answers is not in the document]
sanderpick: Yep, they do retry:
> In live mode, Stripe attempts to deliver your webhooks for up to three days with an exponential back off. In the Events section of the Dashboard, you can view when the next retry will occur.

More info [here](https://stripe.com/docs/webhooks/best-practices#events-and-retries). They also have this [event monitoring tool](https://github.com/stripe/stripe-webhook-monitor) that looks like it could be handy at some point.

Out event handling logic is idempotent, so duplicate events should not be a problem.

api/billingd/service/service.go
```diff
@@ -0,0 +1,735 @@
...
+		return nil, i.Err()
+	}
+	for i.Next() {
+		sum = i.UsageRecordSummary()
+	}
+	if sum != nil && sum.Period != nil {
+		return sum, nil
+	}
+	return nil, fmt.Errorf("subscription item %s not found", id)
+}
```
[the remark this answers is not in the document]
sanderpick: Good point. Stored data could get a valid negative increment, but not the others. I'll make a note.

cmd/buck/cli/push.go
sanderpick: Yep, changed.

core/thread_interceptor.go
sanderpick: Yep, pretty fragile. Curious if there's a best practice for this kind of method matching. Will look around.

sanderpick:
> Developer creates account, accrues data storage up to free amount and then doesn't upgrade/create account. They leave. Now what? We don't want to hold onto that data forever right? Do we imagine a time limit on keeping it around? Maybe a notification that it will be cleaned out? In theory, they might re-create their account (which you've covered), and would expect their data to still be there for a time.

@carsonfarmer Good question. At some point we could track account activity and if it goes stale for a year or so, we send an email saying we'll delete the account.

sanderpick:
> Wow, not sure what to say other than this is a huge amount of really great work. The high level design and thinking is sound, and the low level details are too many and too complex to have any real feedback, so let's get this thing deployed and see what happens! Seriously awesome work.

Thanks! Took a good bit of back and forth... it will nice to try it out and see what needs tweaking.

PR #375, opened 2020-10-07
fix(limits): never set account less than 0 on remove

andrewxhill:


sanderpick:
How about just doing a `max(0, newTotalSize)` [here](https://github.com/textileio/textile/blob/master/mongodb/accounts.go#L333), with a note that this is temporary.

andrewxhill:
Okay cool, will do that.

andrewxhill:
updated

sanderpick (approved):

PR #377, opened 2020-10-08
buckets: adds bucket-wide locks

sanderpick:
Adds the semaphore mechanism used in `go-threads` to ensure buckets are not concurrently altered.


jsign (approved):
Sounds like a really good idea.

PR #391, opened 2020-11-05
Consolidate users into accounts

### LINKED ISSUE #315: Consolidate Account and User

sanderpick:
Closes #315

Consolidates the `User` model into `Account` with a new `AccountType` of User.


mongodb/accounts.go
```diff
@@ -99,12 +107,16 @@ func NewAccounts(ctx context.Context, db *mongo.Database) (*Accounts, error) {
 	_, err := a.col.Indexes().CreateMany(ctx, []mongo.IndexModel{
 		{
 			Keys: bson.D{{"username", 1}},
-			Options: options.Index().SetUnique(true).
-				SetCollation(&options.Collation{Locale: "en", Strength: 2}),
+			Options: options.Index().
+				SetUnique(true).
+				SetCollation(&options.Collation{Locale: "en", Strength: 2}).
+				SetSparse(true),
```
jsign, line 113: Q: Does the existing index get this new `SetSparse(true)` attribute? Or only gets considered if the index is created for the first time?

jsign (approved):
Very nice unification!

mongodb/accounts.go
sanderpick: Unfortunately, tt will have to be part of the migration. Running this against the existing DB will result in:
```
(IndexOptionsConflict) Index with name: username_1 already exists with different options
```
So, we'll have to drop and recreate the index.

sanderpick: I'll start on that work now.

jsign: Cool, sound reasonable. Actually, just asking out of curiosity!

PR #392, opened 2020-11-06
Powergate 1.2.1

asutula:
Goal here is to do the bare minimum to integrate the new Powergate API, as few changes as possible. Future PRs will build on the new Powergate API.

Also made some updates to silence compiler warnings about unkeyed mongo structs.


api/bucketsd/service.go
```diff
@@ -2391,116 +2391,125 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
 		if err != nil {
 			return nil, fmt.Errorf("parsing old Cid archive: %s", err)
 		}
 
+		statusName, found := userPb.JobStatus_name[int32(ba.Archives.Current.JobStatus)]
+		if !found {
+			return nil, fmt.Errorf("invalid job status %v", ba.Archives.Current.JobStatus)
+		}
+
+		status := userPb.JobStatus(ba.Archives.Current.JobStatus)
```
asutula, line 2447: Turns out that the `ffs.JobStatus` raw values are the same as the proto enum `userPb.JobStatus` raw values. So that makes it easy to not use the `ffs` version. This is desirable because the Powergate client no longer returns `ffs` types from its API.

buckets/archive/tracker.go
```diff
@@ -146,7 +146,7 @@ func (t *Tracker) Track(
 	dbID thread.ID,
 	dbToken thread.Token,
 	bucketKey string,
-	jid ffs.JobID,
+	jid string,
```
asutula, line 149: Same here, not using `ffs` types anymore, but this should just work because `ffs.JobID` is just an alias for `string`.

core/core.go
```diff
@@ -80,31 +77,18 @@ var (
 		"/threads.pb.API/ListDBs",
 	}
 
-	healthServiceName = "health.rpc.RPCService"
-	netServiceName    = "net.rpc.RPCService"
-	ffsServiceName    = "ffs.rpc.RPCService"
-	walletServiceName = "wallet.rpc.RPCService"
+	powergateServiceName = "powergate.user.v1.UserService"
+
+	// ToDo: Add support for streaming methods and double check the list for completeness.
```
asutula, line 82: Upcoming PR

cmd/hubd/main.go
```diff
@@ -234,6 +238,12 @@ func init() {
 		config.Flags["threadsMaxNumberPerOwner"].DefValue.(int),
 		"Max number threads per owner")
 
+	// Auth tokens
+	rootCmd.PersistentFlags().String(
+		"tokenPowergateAdmin",
+		config.Flags["tokenPowergateAdmin"].DefValue.(string),
+		"Auth token for Powergate admin APIs")
```
asutula, line 245: New flag to provide the powergate admin auth token.

core/core.go
```diff
@@ -194,6 +178,8 @@ type Config struct {
 
 	Hub   bool
 	Debug bool
+
+	TokenPowergateAdmin string
```
asutula, line 182: This gets set from a flag and passed down into wherever uses the powergate admin client.

api/bucketsd/service.go
```diff
@@ -2332,12 +2333,13 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 		return nil, fmt.Errorf("parsing cid path: %s", err)
 	}
 
-	createNewFFS := func() error {
-		id, token, err := s.PowergateClient.FFS.Create(ctx)
+	createNewUser := func() error {
+		ctxAdmin := context.WithValue(ctx, pow.AdminKey, s.PowergateAdminToken)
```
asutula, line 2337: This feels a little spaghetti to have to have the Powergate admin token set on `Service` and then use it here to build a context...

One day I plan on making the Powergate client itself statefuly manage the token and internally create/use the grpc metadata so the client can just be configured once and passed around as needed.

cmd/hubd/main.go
```diff
@@ -116,6 +116,10 @@ var (
 				Key:      "threads.max_number_per_owner",
 				DefValue: 100,
 			},
+			"tokenPowergateAdmin": {
+				Key:      "token.powergate_admin",
```
sanderpick, line 120: How about "powergate.admin_token"? That way the config file will have a top-level `powergate` section. Then the flag becomes `powergateAdminToken`.

mongodb/accounts.go
```diff
@@ -106,20 +106,20 @@ func NewAccounts(ctx context.Context, db *mongo.Database) (*Accounts, error) {
 	a := &Accounts{col: db.Collection("accounts")}
 	_, err := a.col.Indexes().CreateMany(ctx, []mongo.IndexModel{
 		{
-			Keys: bson.D{{"username", 1}},
+			Keys: bson.D{primitive.E{Key: "username", Value: 1}},
```
sanderpick, line 109: Nice, more explicit 👍

sanderpick (approved):
LGTM! One suggestion for the new flag/config param.

jsign:
@asutula , do you mind bumping the `JobStatusPollInterval` in `archive/tracker.go`?
This time is how much we check again for in-progress archives to query if they reached finality.
Current `30s` is very low for mainnet... this should be maybe `30min` or similar.

Issue #366, opened 2020-09-28
hub version showing dirty api version; hub update attempting wrong binary


andrewxhill:
```
~/Downloads/hub_v2.0.2_darwin-amd64                                           ⍉
▶ hub version
> v2.0.2
> The Hub API is running v2.0.3-dirty
> Warning! There is a new hub release. Run hub update to install 2.0.3.

~/Downloads/hub_v2.0.2_darwin-amd64                                            
▶ hub update 
> Error! File 'hub' for the command is not found in https://github.com/textileio/textile/releases/download/v2.0.3/hubd_v2.0.3_darwin-amd64.zip
> Warning! Error: install failed.
> v2.0.2

~/Downloads/hub_v2.0.2_darwin-amd64                                            
▶ hub version
> v2.0.2
> The Hub API is running v2.0.3-dirty
> Warning! There is a new hub release. Run hub update to install 2.0.3.
```

i guess the version command predated the `hubd` build in the releases, so we just need to a stronger filter for that. not sure what happened with dirty on the API.

andrew:
Was just about to report the same thing. Also the `install` script has been removed so for new users or installations they need to know (or go googling) where to put the `hub` binary.

andrew:
Just manually updated to 2.1.3, here's my output for similar commands:
```
$ hub version
> v2.1.3
> The Hub API is running v2.1.3-dirty
> Error! GET https://api.github.com/repos/textileio/textile/releases: 401 Bad credentials []
> Warning! Unable to get latest release.
$ hub update
> Error! GET https://api.github.com/repos/textileio/textile/releases: 401 Bad credentials []
> Warning! Unable to fetch latest public release.
> v2.1.3
$ hub version
> v2.1.3
> The Hub API is running v2.1.3-dirty
> Error! GET https://api.github.com/repos/textileio/textile/releases: 401 Bad credentials []
> Warning! Unable to get latest release.
```

andrewxhill:
okay, thanks. your error is unrelated to this issue so will continue over on #367

PR #394, opened 2020-11-09
fix(cli): fixes hub bin parsing now that hubd in assets

andrewxhill:
Fixes https://github.com/textileio/textile/issues/366


sanderpick (approved):
LGTM

PR #407, opened 2020-11-16
feat(analytics): adds basic signup stream

andrewxhill:


cmd/hubd/main.go
```diff
@@ -96,6 +96,10 @@ var (
 				Key:      "dns.token",
 				DefValue: "",
 			},
+			"segmentKey": {
```
sanderpick, line 99: This should be on the billing daemon

api/billingd/client/client.go
```diff
@@ -54,6 +67,19 @@ func (c *Client) CreateCustomer(ctx context.Context, key thread.PubKey, opts ...
 	if err != nil {
 		return "", err
 	}
+
+	// Create new analytics entry
```
sanderpick, line 71: Shouldn't this happen in the service handler, not the client?

api/billingd/client/client.go
```diff
@@ -54,6 +67,19 @@ func (c *Client) CreateCustomer(ctx context.Context, key thread.PubKey, opts ...
...
+	if c.sc != nil {
+		c.sc.Enqueue(analytics.Identify{
+			UserId: key.String(),
+			Traits: analytics.NewTraits().
+				SetEmail(args.email).
+				Set("parent_key", parentKey).
+				Set("account_type", args.accountType).
```
sanderpick, line 78: Nice, maybe we can put `account_type` on the `Customer` object. Good info.

core/usage_interceptor.go
```diff
@@ -56,6 +56,7 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
 	if t.bc == nil {
 		return ctx, nil
 	}
+
```
sanderpick, line 59: Adding newlines?

sanderpick (commented):
Will be cool to these data come in! Left some comments.

Related to the general approach, instead of collection each event manually, i.e., "customer created", we could just pipe all the [stripe events](https://stripe.com/docs/api/events/types) into segment using the [webhook handler](https://github.com/textileio/textile/blob/master/api/billingd/gateway/gateway.go#L120). Maybe that's what this [stripe integration](https://segment.com/docs/connections/sources/catalog/cloud-apps/stripe/) does.

api/billingd/service/service.go
```diff
@@ -362,15 +376,16 @@ func (s *Service) Stop(force bool) error {
...
-	Status     string `bson:"status"`
-	Balance    int64  `bson:"balance"`
-	Billable   bool   `bson:"billable"`
-	Delinquent bool   `bson:"delinquent"`
-	CreatedAt  int64  `bson:"created_at"`
+	Key         string `bson:"_id"`
+	CustomerID  string `bson:"customer_id"`
+	ParentKey   string `bson:"parent_key"`
+	Email       string `bson:"email"`
+	AccountType string `bson:"account_type"`
```
sanderpick, line 383: How about using the [account type](https://github.com/textileio/textile/blob/master/mongodb/accounts.go#L45) directly?

andrewxhill: what does that look like in the request object? just convert it to int64 and then back?

sanderpick: I think so, yep. Your new client option could take `mdb.AccountType`, the just wrap that it `int64()` in the client method, the when creating the customer, `mdb.AccountType(req.AccountType)`

andrewxhill: k, done + rebased

sanderpick (approved):
LGTM!

PR #408, opened 2020-11-17
Daily evaluated billing, monthly invoicing

sanderpick:
Adds a daily evaluation mechanism and a number of simplifications to the API.


api/billingd/pb/billingd.proto
```diff
@@ -149,8 +118,5 @@ service APIService {
     rpc UpdateCustomerSubscription(UpdateCustomerSubscriptionRequest) returns (UpdateCustomerSubscriptionResponse) {}
     rpc RecreateCustomerSubscription(RecreateCustomerSubscriptionRequest) returns (RecreateCustomerSubscriptionResponse) {}
     rpc DeleteCustomer(DeleteCustomerRequest) returns (DeleteCustomerResponse) {}
-    rpc IncStoredData(IncStoredDataRequest) returns (IncStoredDataResponse) {}
-    rpc IncNetworkEgress(IncNetworkEgressRequest) returns (IncNetworkEgressResponse) {}
-    rpc IncInstanceReads(IncInstanceReadsRequest) returns (IncInstanceReadsResponse) {}
-    rpc IncInstanceWrites(IncInstanceWritesRequest) returns (IncInstanceWritesResponse) {}
+    rpc IncCustomerUsage(IncCustomerUsageRequest) returns (IncCustomerUsageResponse) {}
```
jsign, line 121: Nice, new measures won't require new APIs.

api/billingd/service/service.go
```diff
@@ -159,102 +232,48 @@ func NewService(ctx context.Context, config Config) (*Service, error) {
...
-		s.config.InstanceWritesPriceID, err = s.createPrice(
-			sc,
-			"ThreadDB writes",
-			stripe.PriceRecurringAggregateUsageSum,
-			InstanceWritesFreeUnitsPerInterval,
-			InstanceWritesUnitCostPerInterval,
-		)
-		if err != nil {
-			return nil, err
+	if _, err := s.reporter.AddFunc("@daily", func() {
```
jsign, line 235: Nice

api/billingd/service/service.go
```diff
@@ -740,241 +717,135 @@ func (s *Service) handleStoredData(
...
+	if !ok {
+		return nil, nil
 	}
-	if cus.ParentKey != "" {
-		if _, err := s.handleInstanceReads(ctx, cus.ParentKey, incCount); err != nil {
-			return nil, err
-		}
+	total := usage.Total + incSize
+	if total < 0 {
+		total = 0
```
jsign, line 761: Nit: Maybe some warn

api/billingd/service/service.go
```diff
@@ -740,241 +717,135 @@ func (s *Service) handleStoredData(
...
+	return int64(math.Round(units))
+}
+
+func getPeriodDays() int {
+	// Go to next month
+	d1 := time.Now().AddDate(0, 1, 0)
+	// Zeroing days will cause time to normalize the date to the last day of the current month.
+	d2 := time.Date(d1.Year(), d1.Month(), 0, 0, 0, 0, 0, time.UTC)
+	return d2.Day()
 }
```
jsign, line 851: If we're trying to get the last day of the month, I'm not sure this is correct for a border case: last day of the month of a 31 day month.

Easier to explain [here](https://play.golang.org/p/V_S7EXrCt6a).
There also I propose another solution which I think doesn't have this problem, since day is normalized to day 1 before making the hairy 1-month-sum, so avoiding the 30 vs 31 prob.

jsign (approved):
LGTM, I left some important comment regarding a particular function to double-check. 👍🏼

api/billingd/service/service.go
sanderpick: Ah, good catch. Though I'm not sure the nanosecond addition is needed. This also works:

```
func getPeriodDays(now time.Time) int {
	// Go to next month
	d1 := time.Date(now.Year(), now.Month(), 1, 0, 0, 0, 0, time.UTC).AddDate(0, 1, 0)
	// Zeroing days will cause time to normalize the date to the last day of the current month.
	d2 := time.Date(d1.Year(), d1.Month(), 0, 0, 0, 0, 0, time.UTC)
	return d2.Day()
}
```

Though, I'm not actually using that method since it's not possible to dynamically set the unit price when reporting to stripe. So, kinda wrote that and forgot about it. Will remove.

Currently, monthly-type pricing is determined using 30 days for a month. However, I just had a thought... why not use the average of 30.4167.

jsign: Yeah, I think assuming 30 is fine if 30.4167 might make some manual-calculations unnecessarily hard (if any manual calc makes sense).

sanderpick: Stuck with the average for now.

api/billingd/service/service.go
sanderpick: Added!

api/billingd/pb/billingd.proto
sanderpick: Yep, also cuts out a ton of very similar looking code

PR #412, opened 2020-11-19
Archive Deals Insight

asutula:
- Adds `DealInfo[]` to the Mongo `Archive` object
- Populates `Archive.DealInfo[]` every time we poll Powergate for job status
- Exposes current and historical `Archive` data via a new/simplified API
- Uses configurable "slow" and "fast" job polling intervals to periodically check job status with the interval depending on whether or not the archive contains deals that are sealing or not.


Makefile
```diff
@@ -131,7 +131,7 @@ install-protoc:
 
 PROTOCGENGO=$(shell pwd)/buildtools/protoc-gen-go
 protos: install-protoc clean-protos
-	./scripts/protoc_gen_plugin.bash \
+	PATH=$(PROTOCGENGO):$(PATH) ./scripts/protoc_gen_plugin.bash \
```
asutula, line 134: This wasn't using the `protoc-gen-go` inside the `buildtools` folder. This makes sure it does and results in a newer version of `protoc-gen-go` being used than on @sanderpick's computer, so all the generated code shows updates in this PR.

api/bucketsd/client/client.go
```diff
@@ -370,11 +370,9 @@ func (c *Client) Archive(ctx context.Context, key string, opts ...ArchiveOption)
 	return err
 }
 
-// ArchiveStatus returns the status of a Filecoin bucket archive.
-func (c *Client) ArchiveStatus(ctx context.Context, key string) (*pb.ArchiveStatusResponse, error) {
-	return c.c.ArchiveStatus(ctx, &pb.ArchiveStatusRequest{
-		Key: key,
-	})
+// Archives returns information about current and historical archives.
+func (c *Client) Archives(ctx context.Context, key string) (*pb.ArchivesResponse, error) {
```
asutula, line 374: Combining those two old methods into one here. `Archives` returns an object that looks like:
```
{
  Current: nil|Archive,
  History: Archive[]
}
```
If the bucket has never been archived, both those values are the nil/empty values.
If the bucket has an archive currently processing, it will be listed in `Current` and is updates in near-real-time as new deal information is available. The previously completed archive is moved to the first slot in `History`.

asutula: I added this comment here in the client, but the change is reflected all the way through the stack.

buckets/archive/tracker.go
```diff
@@ -22,8 +23,7 @@ const (
 )
 
 var (
-	CheckInterval         = time.Second * 15
-	JobStatusPollInterval = time.Minute * 30
```
asutula, line 26: This is now configured and passed in here

buckets/archive/tracker.go
```diff
@@ -169,11 +177,10 @@ func (t *Tracker) trackArchiveProgress(
 	jid string,
 	bucketRoot cid.Cid,
 	powInfo *mdb.PowInfo,
-) (bool, string, error) {
+) (time.Duration, string, error) {
```
asutula, line 180: This now returns the duration that should be used for rescheduling. It will be different depending on if we have any deals in sealing state or not. `0` is interpreted as "don't reschedule".

buckets/archive/tracker.go
```diff
@@ -182,27 +189,40 @@ func (t *Tracker) trackArchiveProgress(
...
 		if strings.Contains(err.Error(), "auth token not found") {
-			return false, "", err
+			return 0, "", err
 		}
-		return true, fmt.Sprintf("getting current job %s for bucket %s: %s", jid, buckKey, err), nil
-	}
-
-	if !isJobStatusFinal(res.StorageJob.Status) {
-		return true, "no final status yet", nil
+		return t.jobPollIntervalSlow, fmt.Sprintf("getting current job %s for bucket %s: %s", jid, buckKey, err), nil
```
asutula, line 194: For recoverable errors, we reschedule with the "slow" interval.

buckets/archive/tracker.go
```diff
@@ -182,27 +189,40 @@ func (t *Tracker) trackArchiveProgress(
...
+			return t.jobPollIntervalSlow, fmt.Sprintf("saving deal data in archive: %s", err), nil
 		}
 	}
+	// Step 3: Update status on Mongo for the archive.
 	if err := t.updateArchiveStatus(ctx, buckKey, res.StorageJob, false, ""); err != nil {
-		return true, fmt.Sprintf("updating archive status: %s", err), nil
+		return t.jobPollIntervalSlow, fmt.Sprintf("updating archive status: %s", err), nil
+	}
+
+	rescheduleDuration := t.jobPollIntervalFast
```
asutula, line 209: Use the fast interval by default.

buckets/archive/tracker.go
```diff
@@ -182,27 +189,40 @@ func (t *Tracker) trackArchiveProgress(
...
+	message := "non-final status"
+	if isJobStatusFinal(res.StorageJob.Status) {
+		message = "reached final status"
+		rescheduleDuration = 0
```
asutula, line 213: If the job is final, use duration `0` meaning "don't reschedule"

buckets/archive/tracker.go
```diff
@@ -182,27 +189,40 @@ func (t *Tracker) trackArchiveProgress(
...
+	}
+
+	if rescheduleDuration > 0 {
+		for _, dealInfo := range res.StorageJob.DealInfo {
+			if dealInfo.StateId == storagemarket.StorageDealSealing {
```
asutula, line 218: If we are going to reschedule, inspect the deal states and decide if we should use short or long duration.

I really hate the use of `storagemarkets` here, but it's necessary since we aren't modeling those states ourself at any point in the stack. Maybe we should so we can avoid this and have typed states for deals.

buckets/archive/tracker.go
```diff
@@ -227,26 +247,47 @@ func (t *Tracker) updateArchiveStatus(
...
+	archiveToUpdate.DealInfo = dealInfo
+
 	if err := t.colls.BucketArchives.Replace(ctx, ba); err != nil {
 		return fmt.Errorf("updating bucket archives status: %s", err)
 	}
 	return nil
 }
 
-func (t *Tracker) saveDealsInArchive(
+func (t *Tracker) saveDealsInBucket(
```
asutula, line 290: Rename for clarity.

buckets/local/archive.go
```diff
@@ -176,131 +175,52 @@ type ArchiveMessageType int
...
 	if err != nil {
 		return nil, err
 	}
 	key := b.Key()
-	rep, err := b.clients.Buckets.ArchiveStatus(ctx, key)
+	return b.clients.Buckets.Archives(ctx, key)
+}
+
+// ArchiveWatch delivers messages about the archive status.
+func (b *Bucket) ArchiveWatch(ctx context.Context, watch bool) (<-chan ArchiveStatusMessage, error) {
```
asutula, line 195: I decided to simplify the old pattern of "get status optionally watch by providing a flag" because the types of data returned by status (now `Archives`) and `ArchiveWatch` are very different.

cmd/hubd/main.go
```diff
@@ -124,6 +125,14 @@ var (
 				Key:      "powergate.admin_token",
 				DefValue: "",
 			},
+			"archivesJobPollIntervalSlow": {
+				Key:      "archives.job_poll_interval_slow",
+				DefValue: time.Minute * 30,
+			},
+			"archivesJobPollIntervalFast": {
+				Key:      "archives.job_poll_interval_fast",
+				DefValue: time.Minute * 15,
```
asutula, line 134: Default values for poll intervals.

mongodb/bucketarchives.go
```diff
@@ -19,13 +19,32 @@ type Archives struct {
...
-	FailureMsg string `bson:"failure_msg"`
-	CreatedAt  int64  `bson:"created_at"`
+	Cid        []byte     `bson:"cid"`
+	JobID      string     `bson:"job_id"`
+	Status     int        `bson:"status"`
+	Aborted    bool       `bson:"aborted"`
+	AbortedMsg string     `bson:"aborted_msg"`
+	FailureMsg string     `bson:"failure_msg"`
+	CreatedAt  int64      `bson:"created_at"`
+	DealInfo   []DealInfo `bson:"deal_info"`
```
asutula, line 29: New model to hold deal information. This would just unmarshal to an empty array for pre-existing archives I think? Will look at writing a test for that. Also, maybe we want to consider writing a migration for those pre-existing archives. Not sure how complex that could be.

Makefile
jsign: Glad to see that 9k of code changes are related to this for review :)

api/bucketsd/pb/bucketsd.proto
```diff
@@ -291,7 +298,6 @@ service APIService {
     rpc DefaultArchiveConfig(DefaultArchiveConfigRequest) returns (DefaultArchiveConfigResponse) {}
     rpc SetDefaultArchiveConfig(SetDefaultArchiveConfigRequest) returns (SetDefaultArchiveConfigResponse) {}
     rpc Archive(ArchiveRequest) returns (ArchiveResponse) {}
-    rpc ArchiveStatus(ArchiveStatusRequest) returns (ArchiveStatusResponse) {}
-    rpc ArchiveInfo(ArchiveInfoRequest) returns (ArchiveInfoResponse) {}
+    rpc Archives(ArchivesRequest) returns (ArchivesResponse) {}
```
jsign, line 301: Pointing out that, in theory, this is a breaking change. I'm not sure if Slate uses this or similar so might deserve some preparation for this version release.

buckets/archive/tracker.go
jsign: Nice!

buckets/archive/tracker.go
jsign: Agree. We should eventually have our own states in the API.
Even if we use `storagemarket` here, there's the chance that `storagemarket` version used in Hub might diverge from the one returned by Powergate, if both are running different versions that doesn't break the API. (In fact, that was already the case for a long time since we didn't break APIs in Powergate and didn't update the Powergate client in Hub).

To be fair, I see a really tiny chance that the market's code will be changing the numbers of statuses, so isn't a panicking situation. But I completely agree that the API should be self-contained... not only because is nicer, but because is safer.

Might worth creating an issue?

integrationtest/pg/pg_test.go
```diff
@@ -258,6 +247,8 @@ func setup(t util.TestingTWithCleanup) (context.Context, core.Config, *c.Client,
 	conf.AddrPowergateAPI = powAddr
 	conf.AddrIPFSAPI = util.MustParseAddr("/ip4/127.0.0.1/tcp/5011")
 	conf.AddrMongoURI = "mongodb://127.0.0.1:27027"
+	conf.ArchiveJobPollIntervalFast = time.Second * 5
+	conf.ArchiveJobPollIntervalSlow = time.Second * 10
```
jsign, line 251: You scaled-down respecting semantics :P

mongodb/bucketarchives.go
jsign: Yeah, I'd think the same: unmarshal to empty array for existing things.

Regarding the migration, could be. We could wait a bit to see if existing archives really start asking for this data to be populated, and if justified do it?

jsign (approved):
LGTM!

api/bucketsd/pb/bucketsd.proto
asutula: Yea, I will treat it as a breaking change and do all the necessary things.

buckets/archive/tracker.go
asutula: Yes, done: https://github.com/textileio/powergate/issues/715

integrationtest/pg/pg_test.go
asutula: Haha of course!

mongodb/bucketarchives.go
asutula: Sounds like a plan to me.

jimmylee:
💯

PR #413, opened 2020-11-19
Fix releases

sanderpick:
Github actions changed... can no longer use `set-env` or `add-path` commands without an env override. We'll want to upgrade to [the _new_ way](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) of doing this soon. Overriding for now.


jsign:
Yesterday or the day before I had to update some GH actions in Powergate and Lotus-related ones to newer versions to avoid that flag too.

PR #414, opened 2020-11-19
Monthly usage reporting

sanderpick:
- Adds an API that returns the entire current month's usage (day-to-date unreported usage plus month-to-date reported usage). This was based on some early feedback from @jsonsivar (see https://github.com/textileio/textile/issues/416). 
- Fixes issue where a new stripe customer gets created on every call to `SetupBilling`


api/billingd/client/client.go
```diff
@@ -139,3 +145,10 @@ func (c *Client) IncCustomerUsage(
 		ProductUsage: productUsage,
 	})
 }
+
+func (c *Client) ReportCustomerUsage(ctx context.Context, key thread.PubKey) error {
```
sanderpick, line 149: This is only here to be able to test the existence of reported usage, which is usually done daily by the cron reporter.

api/billingd/client/client_test.go
```diff
@@ -186,6 +186,43 @@ type usageTest struct {
...
+	require.NoError(t, err)
+
+	err = c.ReportCustomerUsage(context.Background(), key)
+	require.NoError(t, err)
+
+	res, err := c.GetCustomerUsage(context.Background(), key)
+	require.NoError(t, err)
+	assert.NotEmpty(t, res.Usage)
+	assert.Equal(t, product.FreeQuotaSize*2, res.Usage[test.key].Total)
+	assert.Equal(t, float64(product.FreeQuotaSize/product.UnitSize)*test.unitPrice, res.Usage[test.key].Cost)
```
sanderpick, line 223: This should be the total amount currently due for the month, assuming no other _daily_ changes.

api/billingd/service/service.go
```diff
@@ -512,19 +515,27 @@ func periodToPb(period Period) *pb.Period {
...
 	}
 	return res
 }
 
-func getUsage(product Product, total int64) *pb.Usage {
-	freeUnits, paidUnits := getDailyUnits(product, total)
+func getCurrentDayBounds() (int64, int64) {
+	now := time.Now()
+	start := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.Local)
+	end := time.Date(now.Year(), now.Month(), now.Day(), 24, 0, 0, 0, time.Local)
```
sanderpick, line 533: Technically this should be minus one second, but it's only for display purposes where it's nice to show just the year, month, and day.

api/billingd/service/service.go
```diff
@@ -755,6 +771,55 @@ func (s *Service) DeleteCustomer(ctx context.Context, req *pb.DeleteCustomerRequ
...
+	return &pb.GetCustomerUsageResponse{
+		Usage: usage,
+	}, nil
+}
+
+func (s *Service) getPeriodUsageItem(id string) (sum *stripe.UsageRecordSummary, err error) {
+	params := &stripe.UsageRecordSummaryListParams{
+		SubscriptionItem: stripe.String(id),
+	}
+	params.Filters.AddFilter("limit", "", "1")
```
sanderpick, line 809: Only return the most recent (current) invoice period.

api/billingd/service/service.go
```diff
@@ -550,6 +562,18 @@ func getUsage(product Product, total int64) *pb.Usage {
 	}
 }
 
+func getUnits(product Product, total int64) (freeUnits, paidUnits int64) {
```
jsign, line 565: Nit: the named returns were mostly as documentation? (since both are the same type, might make sense). 
Asking since names not used in impl and might be a bit weird.

api/billingd/service/service.go
```diff
@@ -755,6 +771,59 @@ func (s *Service) DeleteCustomer(ctx context.Context, req *pb.DeleteCustomerRequ
...
+	ctx context.Context,
+	req *pb.GetCustomerUsageRequest,
+) (*pb.GetCustomerUsageResponse, error) {
+	doc, err := s.getCustomer(ctx, "_id", req.Key)
+	if err != nil {
+		return nil, err
+	}
+	usage := make(map[string]*pb.Usage)
+	for k, u := range doc.DailyUsage {
+		if product, ok := s.products[k]; ok {
```
jsign, line 784: Q: Does a `log.Warnf` for the `!ok` case a good idea? Like, some daily usage report for some unknown product.

api/billingd/service/service.go
```diff
@@ -755,6 +771,59 @@ func (s *Service) DeleteCustomer(ctx context.Context, req *pb.DeleteCustomerRequ
...
+			}
+			paid, err := s.getPeriodUsageItem(u.PaidItemID)
+			if err != nil {
+				return nil, err
+			}
+			total := (free.TotalUsage + paid.TotalUsage) * product.UnitSize
+			// Add current day unreported usage
+			if product.FreeQuotaInterval == FreeQuotaDaily &&
+				product.PriceType == PriceTypeIncremental {
+				// Add current day unreported usage
```
jsign, line 798: Nit: dup comment?

api/billingd/service/service.go
jsign: Not sure about code around this, but is the `OrderBy` criteria `Desc` by default? Asking just in case isn't the oldest.

api/billingd/service/service.go
```diff
@@ -755,6 +771,59 @@ func (s *Service) DeleteCustomer(ctx context.Context, req *pb.DeleteCustomerRequ
...
+		SubscriptionItem: stripe.String(id),
+	}
+	params.Filters.AddFilter("limit", "", "1")
+	i := s.stripe.UsageRecordSummaries.List(params)
+	if i.Err() != nil {
+		return nil, i.Err()
+	}
+	for i.Next() {
+		sum = i.UsageRecordSummary()
+	}
```
jsign, line 820: We need an extra `.Err()` check after the `for` loop, the docs say:
```
// Err returns the error, if any,                                                                                                                                                                                                                                                       
// that caused the Iter to stop.                                                                                                                                                                                                                                                        
/ It must be inspected                                                                                                                                                                                                                                                                 
// after Next returns false.  
```

jsign (approved):
LGTM!, left a particularly important comment out there.

api/billingd/service/service.go
sanderpick: Ah good catch. Hard to tell if we need to check `.Err()` before `.Next()`, but doesn't look like it.

api/billingd/service/service.go
sanderpick: Sure, can add

sanderpick: Actually, don't think it makes sense here, but does when the usage is added, since the caller controls the product key.

sanderpick: Eh, I take that back... thinking too fast this morning. Will add here.

api/billingd/service/service.go
sanderpick: Yep, for documentation, but will clean up the return to use those params so it's less weird.

api/billingd/service/service.go
sanderpick: Yep, "reverse-chronological" order. Docs [here](https://stripe.com/docs/api/usage_records/subscription_item_summary_list).

api/billingd/service/service.go
jsign: Yep, not clear.

PR #423, opened 2020-11-23
Release fixes

### LINKED ISSUE #417: Use org creator's email address for the org customer object
This will enable stripe to send payment failed type emails.

sanderpick:
Closes #417

- Core: Fix NPD when accessing account user in the usage interceptor.
- Billing: Create parent if missing. This can happen when an existing user account beats the dev/org account owner in setting up billing.


api/billingd/client/client.go
```diff
@@ -37,20 +38,32 @@ func (c *Client) CheckHealth(ctx context.Context) error {
...
+) (string, error) {
 	args := &options{}
 	for _, opt := range opts {
 		opt(args)
 	}
-	var parentKey string
+	var parent *pb.CreateCustomerRequest_Params
 	if args.parentKey != nil {
-		parentKey = args.parentKey.String()
+		parent = &pb.CreateCustomerRequest_Params{
```
sanderpick, line 54: Parent will be created if it's missing, which means we need all the parent info to do that.

api/billingd/pb/billingd.proto
```diff
@@ -41,19 +45,20 @@ message GetCustomerResponse {
...
     string parent_key = 3;
     string email = 4;
-    string account_status = 5;
-    string subscription_status = 6;
-    int64 balance = 7;
-    bool billable = 8;
-    bool delinquent = 9;
-    int64 created_at = 10;
-    int64 grace_period_end = 11;
+    int32 account_type = 5;
```
sanderpick, line 48: Added `account_type`

api/billingd/service/service.go
```diff
@@ -621,7 +635,7 @@ func getUnits(product Product, total int64) (freeUnits, paidUnits int64) {
 }
 
 func (s *Service) customerToPb(ctx context.Context, doc *Customer) (*pb.GetCustomerResponse, error) {
-	deps, err := s.cdb.CountDocuments(ctx, bson.M{"parent_id": doc.Key})
+	deps, err := s.cdb.CountDocuments(ctx, bson.M{"parent_key": doc.Key})
```
sanderpick, line 638: Typo

api/billingd/service/service.go
```diff
@@ -1004,8 +1019,12 @@ func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error
 	return nil
 }
 
-func (s *Service) reportUnits(product Product, usage Usage) error {
+func (s *Service) reportUnits(product Product, usage Usage, parentKey string) error {
 	freeUnits, paidUnits := getUnits(product, usage.Total)
+	if parentKey != "" {
+		freeUnits += paidUnits
+		paidUnits = 0
+	}
```
sanderpick, line 1027: I meant to add this already. If this is a dependent account, report all units to the free subscription item. This may have caused an issue in production where some heavy users have had daily usage reported to the paid subscription item, but they're still protected by the grace period. We can do some searching in Stripe to find any affected users.

core/usage_interceptor.go
```diff
@@ -77,7 +78,7 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
 	now := time.Now()
 
 	// Collect new users.
-	if account.User.CreatedAt.IsZero() && account.User.Type == mdb.User {
+	if account.User != nil && account.User.CreatedAt.IsZero() && account.User.Type == mdb.User {
```
sanderpick, line 81: Primary cause of the NPDs seen in production. This commit has already been deployed. No more restarts.

core/usage_interceptor.go
```diff
@@ -99,18 +100,33 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
...
 			}
-			opts = append(opts, billing.WithAccountType(account.Owner().Type))
-			if _, err := t.bc.CreateCustomer(ctx, account.Owner().Key, opts...); err != nil {
+			if _, err := t.bc.CreateCustomer(
+				ctx,
+				account.Owner().Key,
+				email,
+				account.Owner().Type,
+				opts...,
+			); err != nil {
```
sanderpick, line 129: New logic to get the right parent params to send along when creating a new customer.

core/usage_interceptor.go
```diff
@@ -99,18 +100,33 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
 	cus, err := t.bc.GetCustomer(ctx, account.Owner().Key)
 	if err != nil {
 		if strings.Contains(err.Error(), mongo.ErrNoDocuments.Error()) {
-			opts := []billing.Option{
-				billing.WithEmail(account.User.Email),
+			email, err := t.getAccountCtxEmail(ctx, account)
+			if err != nil {
+				return ctx, err
```
jsign, line 105: Is it ok to `return ctx, err` instead of maybe safer `return nil, ...` done in other if-branches below?

jsign (approved):

core/usage_interceptor.go
sanderpick: Yes, but it might be more clear to just return `nil, err` everywhere. I'll make a note of that and do it in another change.

PR #430, opened 2020-11-30
Fix Windows CLI problems

### LINKED ISSUE #428: Multiple write error when initializing a bucket
PS C:\Users\Milena\Desktop\dev\github\intootv-backend\bucket> hub bucket init
Enter a name for your new bucket (optional): intooTV█
Enter a name for your new bucket (optional): intooTV█
x Encrypt bucket contents:
x Encrypt bucket contents:
> Selected threadDB Create new
Enter a name for your new threadDB (optional): intooTV█
Enter a name for your new threadDB (optional): intooTV█
←[31m> Error! ←[90mMultiple write errors: [{write errors: [{E11000 duplicate key error collection: textile.threads index: _id.owner_1_name_1 collation: { locale: "en", caseLevel: false, caseFirst: "off", strength: 2, numericOrdering: false, alternate: "non-ignorable", maxVariable: "punct", normalization: false, backwards: false, version: "57.1" } dup key: { _id.owner: BinData(0, 0801122089738ECC72D60D5943EE98D023695EC3CB8A788734A5AC725DDB078EF1A5758A), name: "0x39434f45454f53010b" }}]}, {<nil>}]←[0m←[0m

jsign:
This fixes multiples Hub CLI problems in Windows. 
There were three main problems:
- We had an assumption of `/` being the path separator, which is now changed to `runtime.OsPathSeparator` to be platform-independent.
- The `uiprogress` component used in pushing files doesn't work in Windows, it tries to use `stty` with error: `←[31m> Error! ←[90mExec: "stty": executable file not found in %!P(MISSING)ATH%!←(MISSING)[0m←[0m`. I changed to detect if that's the case and avoid using the component but explain the event with plain text. (see screenshot below).
- The Windows terminal doesn't support ANSI colors as explained in the `aurora` package we use for coloring terminal output. I changed the `aurora` reference to a custom instance which enables or disables coloring depending on the OS.

Before it printed ANSI Colors characters which were very confusing for UX, like:
`←[31m> Error! ←[90mSome errors or similar←[0m←[0m`

Example of how it looks now:
image
So there you can see:
- No ANSI color characters in outputs.
- Progress bars changed to plaintext explanation of the event.
- `.textile` folder is ignored as expected, and #428 is fixed.

Fixes #428

### DIFF
--- a/buckets/local/diff.go
+++ b/buckets/local/diff.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/ipfs/go-merkledag/dagutils"
-	"github.com/logrusorgru/aurora"
+	aurora2 "github.com/logrusorgru/aurora"
 	"github.com/textileio/textile/v2/buckets"
 	"github.com/textileio/textile/v2/cmd"
 )
+
+var aurora = aurora2.NewAurora(runtime.GOOS != "windows")
 
 // Change describes a local bucket change.
 type Change struct {
@@ -35,7 +38,7 @@ func ChangeType(t dagutils.ChangeType) string {
 }
 
 // ChangeColor returns an appropriate color for the given change type.
-func ChangeColor(t dagutils.ChangeType) func(arg interface{}) aurora.Value {
+func ChangeColor(t dagutils.ChangeType) func(arg interface{}) aurora2.Value {
 	switch t {
 	case dagutils.Mod:
 		return aurora.Yellow
@@ -54,6 +57,7 @@ func (b *Bucket) DiffLocal() ([]Change, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	ctx, cancel := context.WithTimeout(context.Background(), cmd.Timeout)
 	defer cancel()
 	diff, err := b.repo.Diff(ctx, bp)
@@ -73,7 +77,7 @@ func (b *Bucket) DiffLocal() ([]Change, error) {
 				return nil, err
 			}
 			for _, n := range names {
-				p := strings.TrimPrefix(n, bp+"/")
+				p := strings.TrimPrefix(n, bp+string(os.PathSeparator))
 				r, err := filepath.Rel(b.cwd, n)
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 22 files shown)


cmd/buck/cli/util.go
```diff
@@ -37,8 +38,15 @@ func handleProgressBars(p *uiprogress.Progress, events chan local.PathEvent) {
...
 		case local.FileStart:
+			if runtime.GOOS == "windows" {
+				cmd.Message("Uploading %s", e.Path)
+				continue
+			}
 			bars[e.Path] = addBar(p, e.Path, e.Size)
 		case local.FileProgress, local.FileComplete:
+			if runtime.GOOS == "windows" {
+				continue
+			}
```
jsign, line 49: We could be creative here and print some dots on progress, but that might create a weird output if the file is too big and the screen gets full of dots. Another option is crafting a way of always printing the % of progress but I'd guess that requires some careful deletion of already printed values and re-printing; sounds like something a library should do.

Sounds to me that we should explore eventually an `uiprogress` similar component that works well in all OSs, instead of jumping in building our own custom Windows solution.

buckets/local/diff.go
```diff
@@ -4,14 +4,17 @@ import (
...
 	"strings"
 
 	"github.com/ipfs/go-merkledag/dagutils"
-	"github.com/logrusorgru/aurora"
+	aurora2 "github.com/logrusorgru/aurora"
 	"github.com/textileio/textile/v2/buckets"
 	"github.com/textileio/textile/v2/cmd"
 )
 
+var aurora = aurora2.NewAurora(runtime.GOOS != "windows")
```
jsign, line 16: The main trick is basically overriding the `aurora` definition with a custom instance with coloring disabled if appropriate, instead of all places using the static methods of the package directly.

cmd/buck/cli/util.go
textileben: I haven't used it, but this golang based UI toolkit looks pretty nice https://github.com/charmbracelet/bubbletea

jsign: Nice, sounds like a good candidate :)

sanderpick: Cool, that looks nice! We've made a [number of custom tweaks](https://github.com/textileio/uiprogress/commits/master) to the `uiprogress` package. Maybe bubbletea supports the things that we're needed. I think the main one is the ability to replace each bar with a final state, e.g., the pushed / pulled Cid. Will take a look!

sanderpick (approved):
LGTM!

PR #432, opened 2020-12-01
Prerelease to NPM

asutula:
It would be nice to do prereleases in GitHub and publish NPM artifacts to the `next` tag instead of overriding the `latest` tag. These changes should enable that.

### DIFF
--- a/.github/workflows/publish-js-libs.yml
+++ b/.github/workflows/publish-js-libs.yml
@@ -9,11 +9,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v1
-      - name: Get latest tag
-        id: latesttag
-        uses: "WyriHaximus/github-action-get-previous-tag@master"
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       - name: Install build tools
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-js-libs.yml
```diff
@@ -31,11 +26,11 @@ jobs:
...
-          node-version: '12.x'
-          registry-url: 'https://registry.npmjs.org'
+          node-version: "12.x"
+          registry-url: "https://registry.npmjs.org"
       - name: Generate JS libs
         run: |
           make js-protos
       - name: Publish JS libs
         run: |
-          ./scripts/publish_js_protos.bash -v ${{ steps.latesttag.outputs.tag }} -t ${{ secrets.NPM_AUTH_TOKEN }}
```
asutula, line 41: I don't see why we need to use that 3rd party action to get the tag. Since this runs with a release trigger, we can just get the tag directly from the context as well as whether or not this is a prerelease.

sanderpick: I think it was just so that it's testable outside of a release trigger, but SGTM.

sanderpick (approved):
Nice!

PR #433, opened 2020-12-01
go-threads v1.0.2

sanderpick:
- Updates `go-threads` to v1.0.2
- Since we're using real Stripe keys for staging and production, events from either env are received by both. This doesn't hurt anything, but it does mean that the webhook thinks it's unhealthy and is threatening to delete it. So, if the event fails our biz logic, we just log a warning and return 200. I'd like to create a different stripe account for staging so we can use _different_ live keys there. Then we can revert this change.


carsonfarmer (approved):
LGTM!

PR #436, opened 2020-12-02
Filecoin naming

asutula:
Updating client and CLI code to use `fil` and/or `filecoin` instead of `pow`. Left an command alias for `pow`.


andrewxhill (approved):
nice! looks good

PR #437, opened 2020-12-02
billing: add db migrations

sanderpick:
Adds migrations functionality and one migration: _make customer_id index sparse_


carsonfarmer (approved):
👍

PR #438, opened 2020-12-03
api analytics and key events

andrewxhill:


api/billingd/service/service.go
```diff
@@ -948,12 +936,23 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
...
 	if total > product.FreeQuotaSize && !cus.Billable {
 		now := time.Now().Unix()
+
+		summary := map[string]interface{}{
+			product.Key + "_name":       product.Name,
+			product.Key + "_usage":      total,
+			product.Key + "_units":      product.Units,
+			product.Key + "_free_quota": product.FreeQuotaSize,
+		}
+		go s.analytics.Update(cus.Key, cus.Email, false, summary)
```
andrewxhill, line 946: updates it on the user profile

api/billingd/service/service.go
```diff
@@ -948,12 +936,23 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
...
+
 		if cus.GracePeriodStart == 0 {
 			cus.GracePeriodStart = now
 			update["grace_period_start"] = cus.GracePeriodStart
+			go s.analytics.NewEvent(cus.Key, "grace_period_start", false, summary)
 		}
 		deadline := cus.GracePeriodStart + int64(s.config.FreeQuotaGracePeriod.Seconds())
 		if now >= deadline {
+			go s.analytics.NewEvent(cus.Key, "grace_period_end", false, summary)
```
andrewxhill, line 955: an event to fire triggers such as email about status

cmd/hubd/docker-compose-dev.yml
```diff
@@ -19,6 +19,9 @@ services:
       - HUB_ADDR_POWERGATE_API
       - HUB_GATEWAY_SUBDOMAINS
       - HUB_EMAIL_SESSION_SECRET=hubsession
+      - HUB_EMAIL_API_KEY
+      - HUB_SEGMENT_API_KEY
+      - HUB_SEGMENT_PREFIX
```
andrewxhill, line 24: helpful if testing local

cmd/hubd/main.go
```diff
@@ -97,13 +97,13 @@ var (
...
+			"emailInviteTmpl": {
+				Key:      "email.invite_template",
+				DefValue: "2",
 			},
-			"emailDomain": {
-				Key:      "email.domain",
-				DefValue: "email.textile.io",
+			"emailConfirmTmpl": {
+				Key:      "email.confirm_template",
+				DefValue: "3",
```
andrewxhill, line 106: email template values, they can be updated on the service (2 and 3 are currently usable ones), but felt like config flags not hard-coded values in the code

sanderpick:
This is awesome! Will take a look in a bit.

sanderpick (approved):
Looks good! 

- I still think we should rename the billing service to "usage", then we could expose the analytics APIs there. That would keep the segment stuff consolidated there. Especially when Hub becomes just a service runner + interceptors. Could do in a follow up PR if you agree.

analytics/analytics.go
```diff
@@ -0,0 +1,91 @@
...
+	api    segment.Client
+	prefix string
+	debug  bool
+}
+
+// NewClient return a segment client.
+func NewClient(segmentAPIKey, prefix string, debug bool) (*Client, error) {
+	var api segment.Client
+	var err error
+	if segmentAPIKey != "" {
```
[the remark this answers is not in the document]
andrewxhill: i was just letting it gracefully continue. if you don't need to run it, just don't give a key. this is nice so all those future calls to the methods don't have to have checks if it had error previously, they just make the call as usual.

andrewxhill:
okay, i think that's finally gtg. moving it all to 'usage' i think makes sense, sure!

api/billingd/service/service.go
```diff
@@ -991,6 +990,14 @@ func (s *Service) reportUsage() error {
 		if err := cursor.Decode(&doc); err != nil {
 			return fmt.Errorf("decoding customer: %v", err)
 		}
+		go s.analytics.Update(doc.Key, doc.Email, false, map[string]interface{}{
```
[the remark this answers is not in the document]
sanderpick: Nice, sgtm

sanderpick:
Nice, LGTM!

PR #443, opened 2020-12-08
billing: avoid empty customer string

sanderpick:
Of course, an empty string is not considered "sparse" in a document... a downside of using Go structs with the mongo API: No way to omit a string field.

### DIFF
--- a/api/billingd/service/service.go
+++ b/api/billingd/service/service.go
@@ -482,10 +482,10 @@ func (s *Service) createCustomer(
 		AccountType: mdb.AccountType(params.AccountType),
 		CreatedAt:   time.Now().Unix(),
 	}
-	if _, err := s.cdb.InsertOne(ctx, doc); err != nil {
+	if err := s.createSubscription(doc); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


api/billingd/service/service.go
```diff
@@ -455,8 +460,23 @@ func (s *Service) createCustomer(
...
+	customer, err := s.stripe.Customers.New(&stripe.CustomerParams{
+		Email: stripe.String(params.Email),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	doc = &Customer{
 		Key:         params.Key,
+		CustomerID:  customer.ID,
```
sanderpick, line 479: Since this method now takes the per-customer lock, we can safely use `getCustomer` to determine if the customer already exists, vs. relying on a write operation (which needed to come before the Stripe customer creation to avoid creating a bunch of unused Stripe customers in the event the doc already exists).

sanderpick: The result is avoiding this while problem of the empty `customer_id` string.

andrewxhill (approved):

PR #445, opened 2020-12-09
fix(analytics): unix seconds not nano; cron all products

andrewxhill:

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -85,7 +85,7 @@ func (c *Client) NewEvent(userID, email, eventName string, active bool, properti
 	}
 }
 
-// FormatTime converts nanos to string in same format for all analytics requests
-func (c *Client) FormatTime(nanos int64) string {
-	return time.Unix(0, nanos).Format(time.RFC3339)
+// FormatUnix converts seconds to string in same format for all analytics requests
+func (c *Client) FormatUnix(seconds int64) string {
+	return time.Unix(seconds, 0).Format(time.RFC3339)
 }
--- a/api/billingd/service/service.go
+++ b/api/billingd/service/service.go
@@ -643,7 +643,7 @@ func addProductToSummary(summary map[string]interface{}, product Product, total 
 	summary[product.Key+"_total"] = total
 }
 
-func (s *Service) getSummary(cus *Customer) map[string]interface{} {
+func (s *Service) getSummary(cus *Customer, deps int64) map[string]interface{} {
 	summary := map[string]interface{}{
 		"account_status":       cus.AccountStatus(),
 		"billable":             cus.Billable,
@@ -652,6 +652,9 @@ func (s *Service) getSummary(cus *Customer) map[string]interface{} {
 		"invoice_period_end":   s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
 		"invoice_period_start": s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
 		"subscription_status":  cus.SubscriptionStatus,
+	}
+	if deps > 0 {
+		summary["dependents"] = deps
 	}
 	return summary
 }
@@ -970,13 +973,13 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
 		if cus.GracePeriodStart == 0 {
 			cus.GracePeriodStart = now
 			update["grace_period_start"] = cus.GracePeriodStart
-			summary := s.getSummary(cus)
+			summary := s.getSummary(cus, 0)
 			addProductToSummary(summary, product, total)
 			go s.analytics.NewEvent(cus.Key, cus.Email, "grace_period_start", false, summary)
 		}
 		deadline := cus.GracePeriodStart + int64(s.config.FreeQuotaGracePeriod.Seconds())
 		if now >= deadline {
-			summary := s.getSummary(cus)
+			summary := s.getSummary(cus, 0)
 			addProductToSummary(summary, product, total)
 			go s.analytics.NewEvent(cus.Key, cus.Email, "grace_period_end", false, summary)
 			return nil, common.ErrExceedsFreeQuota
@@ -1027,12 +1030,11 @@ func (s *Service) reportUsage() error {
 }
 
 func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error {
-	summary := s.getSummary(cus)
 	deps, err := s.cdb.CountDocuments(ctx, bson.M{"parent_key": cus.Key})
 	if err != nil {
 		return err
 	}
-	summary["dependents"] = deps
+	summary := s.getSummary(cus, deps)
 
 	for k, usage := range cus.DailyUsage {
 		if product, ok := s.products[k]; ok {
@@ -1044,7 +1046,6 @@ func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error 
 			log.Debugf("reported usage for %s: %s=%d", cus.Key, k, usage.Total)
 			if product.FreeQuotaInterval == FreeQuotaDaily &&
 				product.PriceType == PriceTypeIncremental {
-				summary[product.Key+"_daily_usage"] = k
 				if _, err := s.cdb.UpdateOne(ctx, bson.M{"_id": cus.Key}, bson.M{
 					"$set": bson.M{"daily_usage." + k + ".total": 0},
 				}); err != nil {
--- a/api/billingd/service/service.go
+++ b/api/billingd/service/service.go
@@ -654,7 +654,7 @@ func (s *Service) getSummary(cus *Customer, deps int64) map[string]interface{} {
 		"subscription_status":  cus.SubscriptionStatus,
 	}
 	if cus.GracePeriodStart > 0 {
-		summary["grace_period_end"] = cus.GracePeriodStart + int64(s.config.FreeQuotaGracePeriod.Seconds())
+		summary["grace_period_end"] = s.analytics.FormatUnix(cus.GracePeriodStart + int64(s.config.FreeQuotaGracePeriod.Seconds()))
 	}
 	if deps > 0 {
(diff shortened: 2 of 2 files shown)


analytics/analytics.go
```diff
@@ -85,7 +85,7 @@ func (c *Client) NewEvent(userID, email, eventName string, active bool, properti
 	}
 }
 
-// FormatTime converts nanos to string in same format for all analytics requests
-func (c *Client) FormatTime(nanos int64) string {
-	return time.Unix(0, nanos).Format(time.RFC3339)
+// FormatUnix converts seconds to string in same format for all analytics requests
+func (c *Client) FormatUnix(seconds int64) string {
+	return time.Unix(seconds, 0).Format(time.RFC3339)
```
andrewxhill, line 90: fix here as seconds not nanos

api/billingd/service/service.go
```diff
@@ -952,8 +953,11 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
 			product.Key + "_usage":      total,
 			product.Key + "_units":      product.Units,
 			product.Key + "_free_quota": product.FreeQuotaSize,
+			"grace_period_start":        s.analytics.FormatUnix(cus.GracePeriodStart),
+			"invoice_period_end":        s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
+			"invoice_period_start":      s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+			"subscription_status":       cus.SubscriptionStatus,
 		}
-		go s.analytics.Update(cus.Key, cus.Email, false, summary)
```
andrewxhill, line 956: not updating here as it will be updated daily + captured in the grace period start/end events below

api/billingd/service/service.go
```diff
@@ -1036,6 +1054,7 @@ func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error
 		} else {
 			log.Warn("%s has invalid product key: %s", cus.Key, k)
 		}
+		go s.analytics.Update(cus.Key, cus.Email, false, summary)
```
andrewxhill, line 1057: now capturing all products during cron, not just storage

api/billingd/service/service.go
```diff
@@ -1020,13 +1024,27 @@ func (s *Service) reportUsage() error {
...
+			"invoice_period_start": s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+			"subscription_status":  cus.SubscriptionStatus,
+		}
 		if product, ok := s.products[k]; ok {
 			if err := s.reportUnits(product, usage, cus.ParentKey); err != nil {
 				return err
 			}
+
+			summary[product.Key+"_name"] = product.Name
+			summary[product.Key+"_usage"] = usage.Total
```
sanderpick, line 1040: Might be nice to keep this mapping exact like `_total`, and below, `_free_quota_size`

api/billingd/service/service.go
```diff
@@ -952,8 +953,11 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
 			product.Key + "_usage":      total,
 			product.Key + "_units":      product.Units,
 			product.Key + "_free_quota": product.FreeQuotaSize,
+			"grace_period_start":        s.analytics.FormatUnix(cus.GracePeriodStart),
+			"invoice_period_end":        s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
+			"invoice_period_start":      s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+			"subscription_status":       cus.SubscriptionStatus,
```
sanderpick, line 959: Maybe also snag the _account status_? `cus.AccountStatus()`

sanderpick: Another one, maybe snag number of dependents? You'd have to do,
```
deps, err := s.cdb.CountDocuments(ctx, bson.M{"parent_key": cus.Key})
if err != nil {
	return nil, err
}
```

sanderpick (approved):
LGTM!

api/billingd/service/service.go
andrewxhill: oop, yep good call

api/billingd/service/service.go
```diff
@@ -950,13 +950,14 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
 
 		summary := map[string]interface{}{
 			product.Key + "_name":       product.Name,
-			product.Key + "_usage":      total,
+			product.Key + "_total":      total,
 			product.Key + "_units":      product.Units,
 			product.Key + "_free_quota": product.FreeQuotaSize,
```
sanderpick, line 955: Nit: did you want to update this one too? `_free_quota_size`?

api/billingd/service/service.go
andrewxhill: okay, nice. made me realize a consolidation that could happen here for fewer calls. pushed

api/billingd/service/service.go
```diff
@@ -1019,14 +1018,34 @@ func (s *Service) reportUsage() error {
...
+		"invoice_period_end":   s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
+		"invoice_period_start": s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+		"subscription_status":  cus.SubscriptionStatus,
+		"billable":             cus.Billable,
+		"delinquent":           cus.Delinquent,
+		"account_status":       cus.AccountStatus(),
+	}
+	deps, err := s.cdb.CountDocuments(ctx, bson.M{"parent_key": cus.Key})
+	if err == nil {
+		summary["dependents"] = deps
```
andrewxhill, line 1032: let it fail silently and grab it on the daily cron

sanderpick: If there's an error there, would be bad, probably worth returning as normal. If there are no results, `deps` will be 0.

api/billingd/service/service.go
```diff
@@ -1019,14 +1018,34 @@ func (s *Service) reportUsage() error {
 }
 
 func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error {
+	summary := map[string]interface{}{
```
sanderpick, line 1021: Is there a way to simplify these with a single function, `getAnalyticsSummary()`?

sanderpick: That could take a product and the customer object?

api/billingd/service/service.go
andrewxhill: kk, will do

api/billingd/service/service.go
andrewxhill: was thinking the same makin coffee. will update

api/billingd/service/service.go
```diff
@@ -1019,11 +1030,18 @@ func (s *Service) reportUsage() error {
...
+		return err
+	}
+	summary := s.getSummary(cus, deps)
+
 	for k, usage := range cus.DailyUsage {
 		if product, ok := s.products[k]; ok {
 			if err := s.reportUnits(product, usage, cus.ParentKey); err != nil {
 				return err
 			}
+			addProductToSummary(summary, product, usage.Total)
```
andrewxhill, line 1044: split this out as a help function. so instead of getSummary(cus, prod), you just getSummary(cus) and then add any number of products with addProductionToSummary (so here in the loop and earlier one off)

api/billingd/service/service.go
```diff
@@ -636,6 +636,29 @@ func getUnits(product Product, total int64) (freeUnits, paidUnits int64) {
...
+		"account_status":       cus.AccountStatus(),
+		"billable":             cus.Billable,
+		"delinquent":           cus.Delinquent,
+		"grace_period_start":   s.analytics.FormatUnix(cus.GracePeriodStart),
+		"invoice_period_end":   s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
+		"invoice_period_start": s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+		"subscription_status":  cus.SubscriptionStatus,
+	}
+	if deps > 0 {
+		summary["dependents"] = deps
```
andrewxhill, line 657: allows updates to happen when deps not looked up, but wont overwrite dependents field from previous updates during cron

api/billingd/service/service.go
```diff
@@ -636,6 +636,32 @@ func getUnits(product Product, total int64) (freeUnits, paidUnits int64) {
...
+		"account_status":       cus.AccountStatus(),
+		"billable":             cus.Billable,
+		"delinquent":           cus.Delinquent,
+		"grace_period_start":   s.analytics.FormatUnix(cus.GracePeriodStart),
+		"invoice_period_end":   s.analytics.FormatUnix(cus.InvoicePeriod.UnixEnd),
+		"invoice_period_start": s.analytics.FormatUnix(cus.InvoicePeriod.UnixStart),
+		"subscription_status":  cus.SubscriptionStatus,
+	}
+	if cus.GracePeriodStart > 0 {
+		summary["grace_period_end"] = s.analytics.FormatUnix(cus.GracePeriodStart + int64(s.config.FreeQuotaGracePeriod.Seconds()))
```
andrewxhill, line 657: realized that these durations could change so better capture it

api/billingd/service/service.go
sanderpick: Slick

api/billingd/service/service.go
sanderpick: Mentioning one more in case it makes sense: `pb.Usage` has a daily cost param (`Cost`), which is derived from other things: https://github.com/textileio/textile/blob/master/api/billingd/service/service.go#L621

andrewxhill: slipped it in https://github.com/textileio/textile/pull/445/commits/7f60f75b8903cad54a2eb457d8c527e831624ce7

sanderpick: Sweet!

PR #446, opened 2020-12-10
Migrate to go-mondo-ds for threads

sanderpick:
- Migrates `go-threads` usage to `go-ds-mongo` or `go-ds-badger`
- Adds CLI and daemon builds to CI
- No longer need local mongo + IPFS running for tests... all services are started as you'd expect. In CI, GH action services are still used otherwise packages can't be tested in parallel (services started by docker compose lead to conflicts... not sure why)


core/core.go
```diff
@@ -275,14 +289,26 @@ func NewTextile(ctx context.Context, conf Config) (*Textile, error) {
...
+			ctx,
+			args.ThreadsMongoUri,
+			args.ThreadsMongoDB,
+			mongods.WithCollName("eventstore"),
+		)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		t.ts, err = tutil.NewBadgerDatastore(args.ThreadsBadgerRepoPath, "eventstore", false)
```
jsign, line 303: Nit: check err.

jsign (approved):
LGTM!

cmd/hub/cli/cli.go
```diff
@@ -38,12 +38,12 @@ var (
 				Key:      "org",
 				DefValue: "",
 			},
-			"key": {
-				Key:      "key",
+			"apiKey": {
```
sanderpick, line 41: I realized these new hub configuration fields / flags conflict with the bucket `key` configuration flag. They were added in `v2.1.4`... but I'm guessing it hasn't been an issue since the bucket flags are applied after the hub flags (and buckets wouldn't work at all if `key` wasn't reachable from the local config).

buckets/local/pull.go
```diff
@@ -138,9 +141,12 @@ looop:
 			}
 		}
 		eg, gctx := errgroup.WithContext(context.Background())
+		lim := make(chan struct{}, MaxPullConcurrency)
```
sanderpick, line 144: When playing around with the docs bucket, I ran into the `too many files open` error since pull was trying to pull a huge amount of file. This limits how many are pulled concurrently at a given time.

api/bucketsd/service.go
```diff
@@ -2504,11 +2504,11 @@ func (s *Service) Archives(ctx context.Context, req *pb.ArchivesRequest) (*pb.Ar
 	}
 	res := &pb.ArchivesResponse{}
 	if ba.Archives.Current.JobID != "" {
-		archive, err := toPbArchive(ba.Archives.Current)
```
sanderpick, line 2507: lint nit: package name conflict. cc @asutula

PR #457, opened 2020-12-21
bucket archive improvements and limits

asutula:
- Configurable limit to the max rep factor for bucket archives, defaults to 3
- Don't expose `address` as part of archive config... just use the user's only address (we don't allow users to create additional addresses)
- Better json representation of `ArchiveConfig` that is consistent with the JS client. Makes json `ArchiveConfig` data interchangeable and consistent between the two clients.


api/bucketsd/pb/bucketsd.proto
```diff
@@ -193,10 +193,9 @@ message ArchiveConfig {
     repeated string trusted_miners = 4;
     repeated string country_codes = 5;
     ArchiveRenew renew = 6;
-    string addr = 7;
```
asutula, line 196: No more address in the api

api/bucketsd/service.go
```diff
@@ -94,17 +94,18 @@ type ctxKey string
...
+	GatewayURL                string
+	GatewayBucketsHost        string
+	IPFSClient                iface.CoreAPI
+	IPNSManager               *ipns.Manager
+	PowergateClient           *pow.Client
+	PowergateAdminToken       string
+	ArchiveTracker            *archive.Tracker
+	Semaphores                *nutil.SemaphorePool
+	MaxBucketSize             int64
+	MaxBucketArchiveRepFactor int
```
asutula, line 108: This gets set via configuration

api/bucketsd/service.go
```diff
@@ -2349,14 +2355,14 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
 	storageConfig := baseArchiveStorageConfig
 	storageConfig.Cold.Filecoin = toFilConfig(archiveConfig)
-
-	if storageConfig.Cold.Filecoin.Address == "" {
-		// We don't have an address to use, which is the case for a BucketArchive.DefaultArchiveConfig
-		// that is the default value, get the default address from the user
-		storageConfig.Cold.Filecoin.Address = defConfRes.DefaultStorageConfig.Cold.Filecoin.Address
-	}
+	// Get the address from the default storage config for this user.
+	storageConfig.Cold.Filecoin.Address = defConfRes.DefaultStorageConfig.Cold.Filecoin.Address
```
asutula, line 2365: We now always get the address from the user's default Powergate `StorageConfig`, they can't modify this via `ArchiveConfig` anymore.

api/bucketsd/service.go
```diff
@@ -2657,3 +2660,16 @@ func toFilConfig(config *mdb.ArchiveConfig) *userPb.FilConfig {
 		TrustedMiners: config.TrustedMiners,
 	}
 }
+
+func (s *Service) validateArchiveConfig(c *mdb.ArchiveConfig) error {
```
asutula, line 2664: Simple validation of an `ArchiveConfig`

buckets/local/archive.go
```diff
@@ -18,41 +18,39 @@ type ArchiveConfig struct {
...
 
 // ArchiveRenew contains renew configuration for a ArchiveConfig.
 type ArchiveRenew struct {
 	// Enabled indicates that deal-renewal is enabled for this Cid.
-	Enabled bool
+	Enabled bool `json:"enabled"`
 	// Threshold indicates how many epochs before expiring should trigger
 	// deal renewal. e.g: 100 epoch before expiring.
-	Threshold int
+	Threshold int `json:"threshold"`
```
asutula, line 53: Normal json formatting that is also consistent with the JS client.

mongodb/bucketarchives.go
```diff
@@ -66,8 +66,6 @@ type ArchiveConfig struct {
 	CountryCodes []string `bson:"country_codes"`
 	// Renew indicates deal-renewal configuration.
 	Renew ArchiveRenew `bson:"renew"`
-	// Addr is the wallet address used to store the data in filecoin
-	Addr string `bson:"addr"`
```
asutula, line 70: Previously serialized address values will just be ignored from here on out.

asutula:
I added the rep factor limit here in Hub, but also maybe we should add that to Powergate?

cmd/hubd/main.go
```diff
@@ -98,6 +98,10 @@ var (
 				Key:      "buckets.max_size",
 				DefValue: int64(4 * gib),
 			},
+			"bucketsArchiveMaxRepFactor": {
+				Key:      "buckets.max_archive_rep_factor",
```
jsign, line 102: Nit: s/`max_archive_rep_factor`/`archive_max_rep_factor`?

cmd/hubd/main.go
```diff
@@ -378,6 +386,7 @@ var rootCmd = &cobra.Command{
 
 		// Buckets
 		bucketsMaxSize := config.Viper.GetInt64("buckets.max_size")
+		bucketsArchiveMaxRepFactor := config.Viper.GetInt("buckets.max_archive_rep_factor")
```
jsign, line 389: If above changes makes sense, also change here.

jsign (approved):
LGTM.

Considering it has also breaking API changes, this PR should fire some major version bump process?

jsign:
> I added the rep factor limit here in Hub, but also maybe we should add that to Powergate?

Sounds like a good safety net and potentially useful for people running Powergate alone.
I wouldn't rush to implement it, but sounds like a good idea.

cmd/hubd/main.go
asutula: oh yea,

PR #461, opened 2020-12-30
Stripe pagination fix and other nits

### LINKED ISSUE #449: Update billing to show mib or gib instead of bytes
```
                          USAGE     FREE QUOTA          DAILY COST  START      END        
  Network egress (bytes)  33069018  10704349222 (100%)  $0.0000     09-Dec-20  09-Jan-21  
  ThreadDB reads          5522      44478 (89%)         $0.0000     09-Dec-20  09-Jan-21  
  ThreadDB writes         1572      18428 (92%)         $0.0000     09-Dec-20  09-Jan-21  
  Stored data (bytes)     23470248  5345238872 (100%)   $0.0000     09-Dec-20  09-Jan-21  
```

the bytes fields for stored and egress are not human friendly, you end up having to count digits or cut/paste out of terminal. mega org giga would be easier

### LINKED ISSUE #460: Incorrect subscription item summary pagination
`getPeriodUsageItem` get the most recent subscription item summary for a subscription item. From the stripe docs:

> The list is sorted in reverse-chronological order (newest first). The first list item represents the most current usage period that hasn’t ended yet. Since new usage records can still be added, the returned summary information for the subscription item’s ID should be seen as unstable until the subscription billing period ends.

tl;dr, the `sis` ID for the most recent item is _unstable_. Actually, it appears to be ephemeral, since subsequent queries return the same data with a different `sid`. The built-in pagination in the Stripe Go client is using this ephemeral `sis` as a cursor to fetch the next page of results, which errors because it's ephemeral. In other words, it doesn't exist as long term thing until the billing period closes. 

Is this a bug in the API? Or maybe in the Go client? It should know not to use the latest `sid` to fetch a previous page. 

In our case, we only ever want the most recent summary. So, we can just call `Next` on the iterator one time.

Note: This is only seen for accounts that span multiple billing cycles (have multiple summaries per sub. item). In the case of a single summary, the pagination quits based on the `more` response param equaling `false`.

sanderpick:
- Fixes subscription item summary pagination (Closes #460)
- Cleans up billing CLI usage output (better units and sorted list) (Closes #449)
- Makes docker builds "clean", as in, not showing a dirty git status
- Includes the "no-pulling" threads branch... may as well make that officially included until we can fix it


asutula (approved):
Looks good to me 🚀

PR #463, opened 2020-12-31
Update docker images

sanderpick:
I've had a local branch with a bunch of docker edits for awhile, but haven't had the time to make a real PR. Here goes...

- This updates all docker images to newer versions of `tini` and `busybox`.
- Properly sets a non-privileged user instead of root and uses its home dir as the data dir (same dir, just now it's the user home as well).
- Now, doing `make hub-up` or `make buck-up` will leverage new development images that allow for remote debugging by using a `dlv` based `ENTRYPOINT` and some additional `-gcflags`. Start a debug session with [delve](https://github.com/go-delve/delve), or via an IDE (GoLand shown below). Using the dev docker compose files, the `textile` and `buckets` services exposes port `40000` for debugging (since you wouldn't run them at the same time), and `billing` exposes port `40001`. This allows you to spin the whole stack locally with one line and still get debugging.

Screen Shot 2020-12-30 at 9 08 01 PM

- Fixes the log file handling for all binaries. Previously, all log files were titled `threads.log` since a `go-threads` util was being used everywhere. Related, the log file must now be specified, which is a more expected behavior. The default now is just to log to `stdout`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@
 [![Made by Textile](https://img.shields.io/badge/made%20by-Textile-informational.svg?style=popout-square)](https://textile.io)
 [![Chat on Slack](https://img.shields.io/badge/slack-slack.textile.io-informational.svg?style=popout-square)](https://slack.textile.io)
 [![GitHub license](https://img.shields.io/github/license/textileio/textile.svg?style=popout-square)](./LICENSE)
-[![Go Report Card](https://goreportcard.com/badge/github.com/textileio/textile?style=flat-square)](https://goreportcard.com/report/github.com/textileio/textile?style=flat-square)
 [![GitHub action](https://github.com/textileio/textile/workflows/Tests/badge.svg?style=popout-square)](https://github.com/textileio/textile/actions)
 [![standard-readme compliant](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg?style=popout-square)](https://github.com/RichardLitt/standard-readme)
 
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -3,8 +3,7 @@ package analytics
 import (
 	"time"
 
-	logging "github.com/ipfs/go-log"
-	"gopkg.in/segmentio/analytics-go.v3"
+	logging "github.com/ipfs/go-log/v2"
 	segment "gopkg.in/segmentio/analytics-go.v3"
 )
 
@@ -42,7 +41,7 @@ func NewClient(segmentAPIKey, prefix string, debug bool) (*Client, error) {
 // Update updates the user metadata
 func (c *Client) Update(userID, email string, active bool, properties map[string]interface{}) {
 	if c.api != nil && email != "" {
-		traits := analytics.NewTraits()
+		traits := segment.NewTraits()
 		traits.SetEmail(email)
 		for key, value := range properties {
 			traits.Set(key, value)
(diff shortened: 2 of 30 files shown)


README.md
```diff
@@ -3,7 +3,6 @@
 [![Made by Textile](https://img.shields.io/badge/made%20by-Textile-informational.svg?style=popout-square)](https://textile.io)
 [![Chat on Slack](https://img.shields.io/badge/slack-slack.textile.io-informational.svg?style=popout-square)](https://slack.textile.io)
 [![GitHub license](https://img.shields.io/github/license/textileio/textile.svg?style=popout-square)](./LICENSE)
-[![Go Report Card](https://goreportcard.com/badge/github.com/textileio/textile?style=flat-square)](https://goreportcard.com/report/github.com/textileio/textile?style=flat-square)
```
sanderpick, line 6: Removing since it's broken.

analytics/analytics.go
```diff
@@ -3,8 +3,7 @@ package analytics
 import (
 	"time"
 
-	logging "github.com/ipfs/go-log"
-	"gopkg.in/segmentio/analytics-go.v3"
```
sanderpick, line 7: Double import.

analytics/analytics.go
```diff
@@ -3,8 +3,7 @@ package analytics
...
+	logging "github.com/ipfs/go-log/v2"
```
sanderpick, line 6: Switched everything to `v2`.

cmd/buckd/Dockerfile.dev
```diff
@@ -0,0 +1,40 @@
...
+EXPOSE 40000
+
+ENV BUCKETS_PATH /data/buckets
+RUN adduser --home $BUCKETS_PATH --disabled-login --gecos "" --ingroup users buckets
+
+USER buckets
+
+VOLUME $BUCKETS_PATH
+
+ENTRYPOINT ["dlv", "--listen=0.0.0.0:40000", "--headless=true", "--accept-multiclient", "--continue", "--api-version=2", "exec", "/usr/local/bin/buckd"]
```
sanderpick, line 38: `exec` the binary with `dlv` and listen on `40000`. This allows multiple debug sessions to attach, without requiring any.

cmd/buckd/docker-compose-dev.yml
```diff
@@ -3,31 +3,40 @@ services:
...
       - "127.0.0.1:3006:3006"
-      - "3007:3007"
+      - "127.0.0.1:3007:3007"
       - "4006:4006"
       - "127.0.0.1:8006:8006"
+      - "127.0.0.1:40000:40000"
+    security_opt:
+      - "seccomp:unconfined"
+    cap_add:
+      - SYS_PTRACE
```
sanderpick, line 33: Some things needed for remote debugging.

cmd/buckd/main.go
```diff
@@ -230,7 +230,7 @@ var rootCmd = &cobra.Command{
 		debug := config.Viper.GetBool("log.debug")
 		logFile := config.Viper.GetString("log.file")
 		if logFile != "" {
-			err = util.SetupDefaultLoggingConfig(logFile)
+			err = cmd.SetupDefaultLoggingConfig(logFile)
```
jsign, line 233: Q: `SetupDefaultLoggingConfig` seems to have a fixed `ERROR` level. Is that correct considering the `log.debug` flag? Or maybe `log.debug` only applies if `log.file` is empty?

cmd/buckd/docker-compose-dev.yml
jsign: Cool.

jsign (approved):
LGTM! Cool thing for remote debugging.

textileben (approved):
LGTM

cmd/buckd/main.go
sanderpick: This just sets the default on start, and then later on the debug flag is applied to textile modules. Setting anything other than debug here would cause all the libp2p modules to spew lots of logs.

PR #465, opened 2021-01-05
feat(buck): extend the command links with ability to set the output format

mmuenker:
Add the flag `--format`. The flag --format accepts a string to define the output format.
Add support for the format 'json'. If the format is set to json the link struct is printed as parsable JSON object.
Add the flag `--json`. The flag `--json` sets the flag --format to json

With this extension, the link command can be used in an automated environment. For example in a CI script.

I think it makes sense to add the flags for other commands as well.

---
*I completely new at go. I do not know if the changes are compliant with the go standards.*

### DIFF
--- a/cmd/buck/cli/cli.go
+++ b/cmd/buck/cli/cli.go
@@ -77,6 +77,9 @@ func Init(baseCmd *cobra.Command) {
 	archiveCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
 
 	rolesGrantCmd.Flags().StringP("role", "r", "", "Access role: none, reader, writer, admin")
+
+	linksCmd.Flags().BoolP("json", "", false, "Display URL links as json object")
+	linksCmd.Flags().StringP("format", "", "", "Display URL links in the provided format. Options: [json]")
 }
 
 func SetBucks(b *local.Buckets) {
@@ -200,16 +203,30 @@ var linksCmd = &cobra.Command{
 		}
 		links, err := buck.RemoteLinks(ctx, pth)
 		cmd.ErrCheck(err)
-		printLinks(links)
+
+		jsonFormat, err := c.Flags().GetBool("json")
+		cmd.ErrCheck(err)
+		format, err := c.Flags().GetString("format")
+		cmd.ErrCheck(err)
+
+		if jsonFormat == true {
+			format = "json"
+		}
+
+		printLinks(links, format)
 	},
 }
 
-func printLinks(reply local.Links) {
-	cmd.Message("Your bucket links:")
-	cmd.Message("%s Thread link", aurora.White(reply.URL).Bold())
-	cmd.Message("%s IPNS link (propagation can be slow)", aurora.White(reply.IPNS).Bold())
-	if reply.WWW != "" {
-		cmd.Message("%s Bucket website", aurora.White(reply.WWW).Bold())
+func printLinks(reply local.Links, format string) {
+	if format == "json" {
+		cmd.JSON(reply)
+	} else {
+		cmd.Message("Your bucket links:")
+		cmd.Message("%s Thread link", aurora.White(reply.URL).Bold())
+		cmd.Message("%s IPNS link (propagation can be slow)", aurora.White(reply.IPNS).Bold())
+		if reply.WWW != "" {
+			cmd.Message("%s Bucket website", aurora.White(reply.WWW).Bold())
+		}
 	}
 }
 
--- a/cmd/buck/cli/cli.go
+++ b/cmd/buck/cli/cli.go
@@ -20,6 +20,13 @@ var bucks *local.Buckets
 
 var aurora = aurora2.NewAurora(runtime.GOOS != "windows")
 
+type Format string
+
+const (
+	Default Format = "default"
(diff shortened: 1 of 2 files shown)


mmuenker:
I am a little lost.

What triggers the error `You did not provide an API key. You need to provide your API key in the Authorization header`?

How can I run the failed test locally?

sanderpick:
Thanks for doing the DCO.

> What triggers the error `You did not provide an API key. You need to provide your API key in the Authorization header`?
Some tests require keys that are secrets in the Textile Github account, which are pulled in as env var. Your account doesn't have the secret. Don't worry about that test failure though. I can see it's a minimal change. 

> How can I run the failed test locally?
You can run the package test in your IDE or with `go test`: https://www.golang-book.com/books/intro/12

cmd/buck/cli/cli.go
```diff
@@ -77,6 +77,9 @@ func Init(baseCmd *cobra.Command) {
 	archiveCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
 
 	rolesGrantCmd.Flags().StringP("role", "r", "", "Access role: none, reader, writer, admin")
+
+	linksCmd.Flags().BoolP("json", "", false, "Display URL links as json object")
+	linksCmd.Flags().StringP("format", "", "", "Display URL links in the provided format. Options: [json]")
```
sanderpick, line 82: The second param is a shorthand flag. For `json` is might be `j`. However, I would just not allow shorthand for these flags. You can do that by using the `Bool`/`String` method instead of `BoolP`/`StringP`.

sanderpick: Oh, you're setting up two different flags. We could either do `--json` or `--format=json`. I'd say the latter is more flexible. In which case, just delete the `--json` flag. Then use the value of format to compare with a list of supported formats.

cmd/buck/cli/cli.go
```diff
@@ -200,16 +203,30 @@ var linksCmd = &cobra.Command{
...
 	},
 }
 
-func printLinks(reply local.Links) {
-	cmd.Message("Your bucket links:")
-	cmd.Message("%s Thread link", aurora.White(reply.URL).Bold())
-	cmd.Message("%s IPNS link (propagation can be slow)", aurora.White(reply.IPNS).Bold())
-	if reply.WWW != "" {
-		cmd.Message("%s Bucket website", aurora.White(reply.WWW).Bold())
+func printLinks(reply local.Links, format string) {
```
sanderpick, line 220: You could make `format` an enum that has a default value which results in the current behavior, like...
```
type Format string

const(
    Default Format = "default"
    JSON = "json"
)
```

Then the func becomes, `func printLinks(reply local.Links, format Format) {`. You can assert the flag value as `Format` like `Format("json")`.

sanderpick (changes requested):

cmd/buck/cli/cli.go
mmuenker: Is it possible to overload functions with these mechanisms?

If the enum has a default value. Is it still required to pass the `format` function parameter?

cmd/buck/cli/cli.go
mmuenker: I thought if the parameter for the shorthand value is an empty string this functionality is disabled.

> You can do that by using the Bool/String method instead of BoolP/StringP.

I will adapt to these changes.

mmuenker: > I'd say the latter is more flexible. In which case, just delete the --json flag.

Why not both flags?

mmuenker: > Then use the value of format to compare with a list of supported formats.

is it possible to define validation logic to the flag declaration?

sanderpick: > Why not both flags?
Have multiple feels overkill and maybe confusing. Have you seen a similar pattern on other CLI tools?

sanderpick: > is it possible to define validation logic to the flag declaration?

The lib being used here is https://github.com/spf13/cobra. As far as I know, you can only use validation on position arguments.

cmd/buck/cli/cli.go
sanderpick: Go doesn't allow overloading like that. You'd have to explicity call `printLinks(links, Default)`.

sanderpick: It's typically best practice to append the type name to enum values, so the consts would become,
```
type Format string

const(
    DefaultFormat Format = "default"
    JSONFormat = "json"
)
```

cmd/buck/cli/cli.go
mmuenker: > Have multiple feels overkill and maybe confusing. Have you seen a similar pattern on other CLI tools?

yes, that may be overkill. I will remove it.

mmuenker: >  As far as I know, you can only use validation on position arguments.

When the pull request [841](https://github.com/spf13/cobra/pull/841) is merged then it may be possible.

sanderpick:
Thanks you @mmuenker !

sanderpick (approved):

PR #466, opened 2021-01-05
Fix default dns token flag

mmuenker:
While defining the commands `hubd` and `buckd` the default values for the flag `dnsToken` is not correctly access from the Flags map. Currently, the value `dnsDomain` is used as the default for the flag `dnsToken`.


sanderpick:
Thanks @mmuenker! Would you mind adding a DCO signature? https://github.com/textileio/textile/pull/466/checks?check_run_id=1652676122

sanderpick (approved):

PR #468, opened 2021-01-06
Buck init without discarding local changes

sanderpick:
Tricky one requested by a few different users.

- Adds some more `init` flags. From `buck init --help`:
```
By default, if the remote bucket exists, remote objects are pulled and merged with local changes.
Use the '--soft' flag to accept all local changes, including deletions.
Use the '--hard' flag to discard all local changes.
```

### DIFF
--- a/api/bucketsd/client/client.go
+++ b/api/bucketsd/client/client.go
@@ -3,8 +3,7 @@ package client
 import (
 	"context"
 	"io"
-	"runtime"
-	"strings"
+	"path/filepath"
 
 	"github.com/gogo/status"
 	"github.com/ipfs/go-cid"
@@ -73,7 +72,7 @@ func (c *Client) Root(ctx context.Context, key string) (*pb.RootResponse, error)
 func (c *Client) Links(ctx context.Context, key, pth string) (*pb.LinksResponse, error) {
 	return c.c.Links(ctx, &pb.LinksRequest{
 		Key:  key,
-		Path: normalizePath(pth),
+		Path: filepath.ToSlash(pth),
 	})
 }
 
@@ -91,7 +90,7 @@ func (c *Client) ListIpfsPath(ctx context.Context, pth path.Path) (*pb.ListIpfsP
 func (c *Client) ListPath(ctx context.Context, key, pth string) (*pb.ListPathResponse, error) {
 	return c.c.ListPath(ctx, &pb.ListPathRequest{
 		Key:  key,
-		Path: normalizePath(pth),
+		Path: filepath.ToSlash(pth),
 	})
 }
 
@@ -99,7 +98,7 @@ func (c *Client) ListPath(ctx context.Context, key, pth string) (*pb.ListPathRes
 func (c *Client) SetPath(ctx context.Context, key, pth string, remoteCid cid.Cid) (*pb.SetPathResponse, error) {
 	return c.c.SetPath(ctx, &pb.SetPathRequest{
 		Key:  key,
-		Path: normalizePath(pth),
+		Path: filepath.ToSlash(pth),
 		Cid:  remoteCid.String(),
 	})
 }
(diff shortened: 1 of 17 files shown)


api/bucketsd/client/client.go
```diff
@@ -73,7 +72,7 @@ func (c *Client) Root(ctx context.Context, key string) (*pb.RootResponse, error)
 func (c *Client) Links(ctx context.Context, key, pth string) (*pb.LinksResponse, error) {
 	return c.c.Links(ctx, &pb.LinksRequest{
 		Key:  key,
-		Path: normalizePath(pth),
+		Path: filepath.ToSlash(pth),
```
sanderpick, line 75: Switching these back since on windows this `ToSlash` method functions just like the custom func I needed to test the behavior on macOS.

buckets/local/buckets.go
```diff
@@ -228,11 +228,33 @@ func (b *Buckets) NewBucket(ctx context.Context, conf Config, opts ...NewOption)
 
 	// Pull remote bucket contents
 	if !initRemote || args.fromCid.Defined() {
-		if _, err := buck.getPath(ctx, "", cwd, nil, false, args.events); err != nil {
+		if err := buck.repo.Save(ctx); err != nil {
 			return nil, err
 		}
-		if err = buck.repo.Save(ctx); err != nil {
-			return nil, err
+		if args.sync {
```
carsonfarmer, line 234: This is pretty slick! Super useful 👍

cmd/buck/cli/cli.go
```diff
@@ -56,6 +56,7 @@ func Init(baseCmd *cobra.Command) {
 	initCmd.Flags().BoolP("private", "p", false, "Obfuscates files and folders with encryption")
 	initCmd.Flags().String("cid", "", "Bootstrap the bucket with a UnixFS Cid from the IPFS network")
 	initCmd.Flags().BoolP("existing", "e", false, "Initializes from an existing remote bucket if true")
+	initCmd.Flags().Bool("sync", false, "Syncs local state with remote, i.e., discards local changes if true")
```
carsonfarmer, line 59: not a blocker by any means, but it seems like "overwrite" or "force" would be more common terms to use here? but I might be missing some nuance to what sync really means in this case.

carsonfarmer (approved):
This is super nice, we should defs be writing a blog post or at least tweet about this...

```
Buckets bow supports smart sync when initializing from a remote cid.
```
or something like that

carsonfarmer:
Is that race detection just a fluke, and we should re-run the ci tests?

cmd/buck/cli/cli.go
mmuenker: I agree with you there. I tend to "force"

mmuenker:
I have encountered an error:

`Error! Paths containing .textileseed are not allowed`

**Reproduce:**

Run the web app build script
```
foo$ yarn build
```

Init and push the bucket
```
foo$ cd dist
foo/dist$ buck init
foo/dist$ buck push
```
The bucket with the build artifacts was created and is accessible.

Rebuild the project. While rebuilding the dist folder is deleted.
```
foo$ yarn build
```
Init the bucket
```
foo$ cd dist
foo/dist$ buck --thread ... --key ... init
```
Check the status of the bucket
```
foo/dist$ buck status
> modified:  index.html
> modified:  ngsw.json
> deleted:   .textileseed
```

Push the changes to the bucket
```
foo/dist$ buck push
```
When pushing the error is thrown: `Error! Paths containing .textileseed are not allowed`

sanderpick:
Hey @mmuenker, thanks... yeah, there was some unfinished business here.

cmd/buck/cli/cli.go
sanderpick: We use `force` with `pull` in a way that means, "force pull all files, even if they exist locally", which is different that the meaning here... so, I opted to avoid the confusion. However! I have removed the `sync` flag in favor of two other ones that cover more use cases. See PR description.

mmuenker:
> Hey @mmuenker, thanks... yeah, there was some unfinished business here.

Is the solution already implemented?

sanderpick:
Yep, have you had a chance to try the latest release v2.5.0?

PR #469, opened 2021-01-09
feat(analytics): moves package into billing 

andrewxhill:


api/billingd/service/service.go
```diff
@@ -505,10 +505,10 @@ func (s *Service) createCustomer(
 	}
 	log.Debugf("created customer %s with id %s", doc.Key, doc.CustomerID)
 
-	go s.analytics.Update(doc.Key, doc.Email, false, map[string]interface{}{
-		"parent_key":   doc.ParentKey,
-		"customer_id":  doc.CustomerID,
-		"account_type": doc.AccountType,
+	go s.analytics.Identify(doc.Key, doc.AccountType, false, doc.Email, map[string]interface{}{
```
asutula, line 508: Just curious if you're using go routines when making client calls because of possible slow requests? If that is a real concern, maybe it's safer to move the go routine forking to the client itself just so no one forgets to make the calls using a go routine. Another thought is that we should make sure to log ERROR inside of the gRPC client code in the case of a error since we are forking off go routines and have no way of knowing about errors there. If we log ERROR, we can set up some monitoring to let us know if/when our analytics tracking is down (which is quite important if we hang features like grant fil off of it) cc @textileben

api/hubd/service.go
```diff
@@ -315,7 +306,21 @@ func (s *Service) CreateKey(ctx context.Context, req *pb.CreateKeyRequest) (*pb.
 		return nil, status.Errorf(codes.Internal, "mapping key type: %v", key.Type)
 	}
 
-	go s.Analytics.NewEvent(account.User.Key.String(), account.User.Email, "create_key", true, map[string]interface{}{})
+	var event analytics.Event
+	if keyType == mdb.AccountKey {
+		event = analytics.KeyAccountCreated
+	} else {
+		event = analytics.KeyUserGroupCreated
```
asutula, line 313: Above, the only possible values for `keyType` are `mdb.AccountKey` and `mdb.UserKey`. I'm not clear on the naming of things off the top of my head, so just double checking that `mdb.UserKey` should map to `analytics.KeyUserGroupCreated` since "user" and "user group" sound like different things to me.

asutula (commented):
Nice, looking good so far. Left some questions and comments and nits. Not approving for now since it's a draft and I will do a final review once it's all good to go.

api/billingd/analytics/events.go
```diff
@@ -0,0 +1,92 @@
+package analytics
+
+import (
+	"fmt"
+)
+
+// Trait is a bool trait set on user by specific events
+type Trait struct {
+	name  string
+	value bool
```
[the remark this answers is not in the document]
andrewxhill: any type can be set yet, but after writing this i only had bool values on Traits, so just locked it down here. Even though the input downstream can take more than just bools.

api/billingd/service/service.go
andrewxhill: the api requests are in their own routine anyway, since batch management is automatic over there. so just removing this routine and calling it directly.

api/hubd/service.go
andrewxhill: yeah, nice. moved to just KeyUserCreated, no "Group"

api/billingd/Dockerfile.dev
```diff
@@ -20,6 +22,7 @@ LABEL maintainer="Textile <contact@textile.io>"
 
 ENV SRC_DIR /textile
 COPY --from=0 /go/bin/dlv /usr/local/bin/dlv
+COPY --from=0 /etc/ssl/certs /etc/ssl/certs
```
andrewxhill, line 25: couldn't do any local testing of remote apis without these.

api/billingd/analytics/analytics.go
```diff
@@ -38,50 +39,51 @@ func NewClient(segmentAPIKey, prefix string, debug bool) (*Client, error) {
 	return client, err
 }
 
-// Update updates the user metadata
-func (c *Client) Update(userID, email string, active bool, properties map[string]interface{}) {
-	if c.api != nil && email != "" {
+// Identify creates or updates the user traits
+func (c *Client) Identify(userID string, accountType mdb.AccountType, active bool, email string, properties map[string]interface{}) error {
```
sanderpick, line 43: Is _userID_ the customer ID, or something specific to segment?

core/usage_interceptor.go
```diff
@@ -120,10 +121,12 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
 				}
 				opts = append(opts, billing.WithParent(parent.Key, email, parent.Type))
 			}
+			// TODO: Add username, name to Create call
```
sanderpick, line 124: Looks like the TODO is complete?

sanderpick (approved):
LGTM! Nice and extensible for the future 🥳

api/billingd/analytics/analytics.go
andrewxhill: ah, i was just using the segment lingo, but it's just the user pubkey (i.e. `account.Owner().Key`)

andrewxhill: changed to `key`

sanderpick: Cool, sgtm

PR #470, opened 2021-01-11
Fix nil ctx

sanderpick:
Hotfix for these panics: https://grafana-v7.hub.next.textile.io/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22Loki%22,%7B%22expr%22:%22%7Bjob%3D%5C%22hub-production%2Fhub%5C%22%7D%22%7D%5D


carsonfarmer (approved):
Nice quick catch.

PR #480, opened 2021-01-21
FIL Rewards Bookkeeping + APIs + Client + CLI

asutula:
This PR:

- Creates a new internal service, `filrewardsd` which unlocks FIL rewards for orgs and provides APIs to query and claim those rewards.
- `filrewardsd` tracks some new analytics events, `fil_reward`, `fil_claim`, and `fil_finalize_claim`. More info on that below.
- Moves analytics into a dedicated internal service, `analyticsd`
- Integrates the FIL rewards client into `analyticsd` in order to implement processing of analytics events to trigger FIL reward unlocking
- Integrates the FIL rewards client into `hubd` in order to implement user facing APIs for querying and claiming FIL rewards
- Adds sub commands to `hub fil` CLI for interacting with FIL rewards APIs

This PR does not:

- Trigger the transfer of FIL to the destination wallet address when rewards are claimed... That will be the next PR.

## New CLI commands

Only new commands shown below.

```
➜ hub fil --help
Interact with Filecoin related commands.

Usage:
  hub fil [command]

Aliases:
  fil, filecoin, pow

Available Commands:
  claim                       Claim unlocked FIL rewards to the specified org
  claims                     List FIL claims for the specified org
  rewards                  List FIL rewards that have been unlocked for the specified dev and/or org
  rewards-balance   Show FIL rewards balance

Flags:
  -h, --help   help for fil

Global Flags:
      --api string         API target (default "api.hub.textile.io:443")
      --apiKey string      User API key
      --apiSecret string   User API secret
      --identity string    User identity
      --newIdentity        Generate a new user identity
  -o, --org string         Org username
  -s, --session string     User session token
      --token string       User identity token
```

## New Analytics Events

Here are examples of the new Segment events related to FIL rewards which can be used to send notifications to users as well as to build our own analytics:

### fil_reward

```
client.Track(&analytics.Track{
  UserId: "bbaareie7rtlulw6sx3o7ylmavbdrodmrgugrlqbwov2o4gg5p7nghxgela",
  Event: "fil_reward",
  Properties: map[string]interface{}{
    "amount_fil": 1.5,
    "amount_nano_fil": 1500000000,
    "base_nano_fil_reward": 500000000,
    "dev_key": "bbaareiagauglwofxc57figmc4vrprwoumnnrd4qxqcq5ddfgpixefmlspe",
    "factor": 3,
    "type": "REWARD_TYPE_FIRST_ORG_CREATED",
  },
})
```

### fil_claim

```
client.Track(&analytics.Track{
  UserId: "bbaareie7rtlulw6sx3o7ylmavbdrodmrgugrlqbwov2o4gg5p7nghxgela",
  Event: "fil_claim",
  Properties: map[string]interface{}{
    "amount_fil": 0.75,
    "amount_nano_fil": 750000000,
    "claimed_by": "bbaareiagauglwofxc57figmc4vrprwoumnnrd4qxqcq5ddfgpixefmlspe",
    "id": "6025aab698fe5b812a743ea8",
  },
})
```

### fil_finalize_claim

```
client.Track(&analytics.Track{
  UserId: "bbaareie7rtlulw6sx3o7ylmavbdrodmrgugrlqbwov2o4gg5p7nghxgela",
  Event: "fil_finalize_claim",
  Properties: map[string]interface{}{
    "failure_message": "",
    "id": "6025aab698fe5b812a743ea8",
    "state": "CLAIM_STATE_COMPLETE",
    "txn_cid": "todo-some-cid",
  },
})
```


api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+	Unspecified Event = iota
+	FirstKeyAccountCreated
+	FirstKeyUserCreated // does this fit "register first user?"
+	FirstOrgCreated
+	InitialBillingSetup
+	FirstBucketCreated
+	FirstBucketArchiveCreated
+	FirstMailboxCreated
+	FirstThreadDbCreated
+)
```
asutula, line 28: We need to keep a set of reward events that are specific to the FIL rewards code since they are used in a historical record, even if the associated `analytics` event no longer exists. These each have an associated `analytics.Event`.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+const baseAttoFILReward = 1000 // What should this be? Should we read it from mongo?
```
asutula, line 30: This is the amount that will be rewarded for the smallest rewards, and it is multiplied based on the reward data just below.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+var eventRewards = map[Event]reward{
+	FirstKeyAccountCreated:    {factor: 3},
+	FirstKeyUserCreated:       {factor: 1},
+	FirstOrgCreated:           {factor: 3},
+	InitialBillingSetup:       {factor: 1},
+	FirstBucketCreated:        {factor: 2},
+	FirstBucketArchiveCreated: {factor: 2},
+	FirstMailboxCreated:       {factor: 1},
+	FirstThreadDbCreated:      {factor: 1},
+}
```
asutula, line 42: Here we store reward metadata associated with each reward event, currently just `factor` which is the multiplier of `baseAttoFILReward` that would determine the total reward.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+type RewardRecord struct {
+	Key               string          `bson:"key"`
+	AccountType       mdb.AccountType `bson:"account_type"`
+	Event             Event           `bson:"event"`
+	Factor            int             `bson:"factor"`
+	BaseAttoFILReward int             `bson:"base_atto_fil_reward"`
+	CreatedAt         time.Time       `bson:"created_at"`
+}
```
asutula, line 55: This represents a historical record of an achieved reward.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+type FilRewards struct {
+	col          *mongo.Collection
+	rewardsCache map[string]map[Event]struct{}
```
asutula, line 59: We use this cache so we don't have to query Mongo every time an event comes in just to find out if the account already achieved the would-be reward.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("connecting to mongo: %v", err)
+	}
+	db := client.Database(config.DBName)
+	col := db.Collection(config.CollectionName)
+	_, err = col.Indexes().CreateMany(ctx, []mongo.IndexModel{
+		{
+			Keys:    bson.D{primitive.E{Key: "key", Value: 1}, primitive.E{Key: "event", Value: 1}},
+			Options: options.Index().SetUnique(true).SetSparse(true),
+		},
```
asutula, line 80: This is a compound index to ensure that every `key` + `event` combo is unique.

asutula: To put it another way, each account key can achieve each reward event only once.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+		{
+			Keys: bson.D{primitive.E{Key: "key", Value: 1}},
+		},
+		{
+			Keys: bson.D{primitive.E{Key: "event", Value: 1}},
+		},
+		{
+			Keys: bson.D{primitive.E{Key: "created_at", Value: 1}},
+		},
```
asutula, line 89: Some other indexes to support different types of queries we'll do.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+		var rec RewardRecord
+		if err := cursor.Decode(&rec); err != nil {
+			return nil, fmt.Errorf("decoding RewardRecord while building cache: %v", err)
+		}
+		ensureKeyEventCache(cache, rec.Key)
+		cache[rec.Key][rec.Event] = struct{}{}
+	}
+	if err := cursor.Err(); err != nil {
+		return nil, fmt.Errorf("iterating cursor while building cache: %v", err)
+	}
```
asutula, line 112: We populate that rewards cache on startup. We'll want to keep an eye on how long this takes since it will grow literally with the amount of rewards stored over time.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+	defer cursor.Close(ctx)
+	cache := make(map[string]map[Event]struct{})
+	for cursor.Next(ctx) {
+		var rec RewardRecord
+		if err := cursor.Decode(&rec); err != nil {
+			return nil, fmt.Errorf("decoding RewardRecord while building cache: %v", err)
+		}
+		ensureKeyEventCache(cache, rec.Key)
+		cache[rec.Key][rec.Event] = struct{}{}
```
asutula, line 108: We cache the account key + event so we can quickly check what has been achieved.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+	case analytics.BucketArchiveCreated:
+		rewardEvent = FirstBucketArchiveCreated
+	case analytics.MailboxCreated:
+		rewardEvent = FirstMailboxCreated
+	case analytics.ThreadDbCreated:
+		rewardEvent = FirstThreadDbCreated
+	}
+	if rewardEvent == Unspecified {
+		return nil, nil
+	}
```
asutula, line 141: Here is where we map from `analytics.Event` to reward `Event`. If the event is one we're not interested in, just return.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+	ensureKeyEventCache(f.rewardsCache, key)
+
+	if _, exists := f.rewardsCache[key][rewardEvent]; exists {
+		// This reward is already granted so bail.
+		return nil, nil
+	}
```
asutula, line 148: Check the cache and return if this account key has already achieved this reward.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+		Factor:            eventRewards[rewardEvent].factor,
+		BaseAttoFILReward: baseAttoFILReward,
+		CreatedAt:         time.Now(),
+	}
+
+	if _, err := f.col.InsertOne(ctx, rec); err != nil {
+		return nil, fmt.Errorf("inserting RewardRecord: %v", err)
+	}
+
+	f.rewardsCache[key][rewardEvent] = struct{}{}
```
asutula, line 163: Record the reward in the db and in the cache.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// ToDo: Notify the account owner.
```
asutula, line 165: Important `ToDo`.

asutula: Email or something?

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return &rec, nil
+}
+
```
asutula, line 169: Below are some APIs to query `RewardRecords`. They will be used to build and RPC API for devs to call to view and claim their achievements as well as for APIs for us to do reporting. These APIs can evolve as we understand exactly what information we need and how we need it.

api/billingd/filrewards/filrewards.go
```diff
@@ -0,0 +1,217 @@
...
+		return nil, fmt.Errorf("decoding RewardRecord: %v", err)
+	}
+	return rec, nil
+}
+
+func ensureKeyEventCache(keyCache map[string]map[Event]struct{}, key string) {
+	if _, exists := keyCache[key]; !exists {
+		keyCache[key] = map[Event]struct{}{}
+	}
+}
```
asutula, line 217: Just a little util func to make sure we don't call into a `nil` `map`.

api/billingd/filrewards/filrewards.go
jsign: Nit: Maybe log.Warn? Just to identify if we get duplications (which might be fine?) just for knowing this might be happening.

api/billingd/filrewards/filrewards.go
jsign: Nit: I'd add a comment explaining that `*RewardRecord` may be nil without an error. Might not be obvious to clients and might cause an unwanted nil-pointer dereference.

api/billingd/filrewards/filrewards.go
asutula: This also makes it easier to have other sources of events some day if we want to.

api/billingd/filrewards/filrewards.go
asutula: I record the `Factor` and `BaseAttoFILReward` at the time the reward was achieved just in case we want to use that information. Maybe we want to use whatever the current values are, but at least we have the option to use the old values that are recorded here.

api/billingd/filrewards/filrewards.go
andrewxhill: nice

api/billingd/filrewards/filrewards.go
andrewxhill: perfect! `baseAttoFILReward` could just be a startup flag on billing for now. might be nice to set a startup check with a max baseAttoFILReward that hard stops

api/billingd/filrewards/filrewards.go
andrewxhill: cool, makes sense. there was one idea here i'll share on slack. it might be covered but hard to summarize here.

api/billingd/filrewards/filrewards.go
andrewxhill: sweet

api/billingd/filrewards/filrewards.go
andrewxhill: you should just create a new analytics event with the reward info `Track()`. emails can then be linked up from there.

api/billingd/filrewards/filrewards.go
andrewxhill: these look perfect for the Dashboard, were we can show completed actions checked off etc. 

for the dev, probably just a roll-up like, get total, would be great.

api/billingd/filrewards/filrewards.go
asutula: Do you mean to add the reward info as metadata to the segment event? Cool, so you can trigger emails from within segment, or? Just want to understand how that will all work.

andrewxhill: yeah, we can trigger emails there. best to create a specific event type for each reward, then yeah, include any useful metadata (e.g. fil rewarded) and we can automate the email from that

sanderpick:
Nice, looking good 👍

api/analyticsd/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/textileio/go-threads/util"
+	"github.com/textileio/textile/v2/api/analyticsd/service"
+	"github.com/textileio/textile/v2/cmd"
+)
+
+const daemonName = "analyticsd"
```
asutula, line 16: Analytics now as its own service. This helped untangle some things and feels much better all around.

api/analyticsd/pb/analytics.proto
```diff
@@ -0,0 +1,57 @@
...
+  EVENT_ORG_INVITE_CREATED = 8;
+  EVENT_GRACE_PERIOD_START = 9;
+  EVENT_GRACE_PERIOD_END = 10;
+  EVENT_BILLING_SETUP = 11;
+  EVENT_BUCKET_CREATED = 12;
+  EVENT_BUCKET_ARCHIVE_CREATED = 13;
+  EVENT_MAILBOX_CREATED = 14;
+  EVENT_THREAD_DB_CREATED = 15;
+  EVENT_FIL_REWARD_RECORDED = 16;
+}
```
asutula, line 25: Proto analytics events are now the source of truth for event definitions.

api/analyticsd/pb/analytics.proto
```diff
@@ -0,0 +1,57 @@
...
+
+message IdentifyResponse {}
+
+message TrackRequest {
+  string key = 1;
+  AccountType account_type = 2;
+  bool active = 3;
+  Event event = 4;
+  google.protobuf.Struct properties = 5;
```
asutula, line 49: Found this nice proto type `Struct` that models JSON data. In generated Go code, it ends up as `map[string]interface{}` which is nice and consistent with what the Segment API wants as well.

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,200 @@
...
+			},
+		},
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, "enqueuing segment track: %v", err)
+	}
+
+	// Don't process the event that is generated by s.frc.ProcessAnalyticsEvent
+	// when a reward record is created and sent here for analytics tracking.
+	if s.frc != nil && req.Event != pb.Event_EVENT_FIL_REWARD_RECORDED {
+		if _, err := s.frc.ProcessAnalyticsEvent(ctx, req.Key, req.AccountType, req.Event); err != nil {
```
asutula, line 136: Here, we send the analytics event to the filrewards service using the new `filrewards.Client`.

api/billingd/main.go
```diff
@@ -48,6 +48,10 @@ var (
 				Key:      "addr.gateway.host",
 				DefValue: "/ip4/127.0.0.1/tcp/8010",
 			},
+			"addrAnalyticsApi": {
+				Key:      "addr.analytics_api",
+				DefValue: "",
+			},
```
asutula, line 54: Any user of the analytics client needs to be configured to create the analytics client, so here in the billing daemon, we add the information we need to create the analytics client. You'll see this elsewhere in hub as well.

api/billingd/service/service.go
```diff
@@ -231,9 +231,13 @@ func NewService(ctx context.Context, config Config) (*Service, error) {
...
-	ac, err := analytics.NewClient(config.SegmentAPIKey, config.SegmentPrefix, config.Debug)
-	if err != nil {
-		return nil, err
+	var ac *analytics.Client
+	if config.AnalyticsAddr != "" {
+		c, err := analytics.New(config.AnalyticsAddr)
+		if err != nil {
+			return nil, err
+		}
+		ac = c
```
asutula, line 240: we only create the analytics client if an address is configured.

api/billingd/service/service.go
```diff
@@ -505,11 +509,19 @@ func (s *Service) createCustomer(
...
+		doc.AccountType.Pb(),
+		analytics.WithEmail(doc.Email),
+		analytics.WithProperties(map[string]interface{}{
+			"parent_key":  doc.ParentKey,
+			"customer_id": doc.CustomerID,
+			"username":    params.Username,
+		}),
+	); err != nil {
+		log.Errorf("calling analytics identify: %v", err)
+	}
```
asutula, line 524: In the case of the analytics client, you don't have to check if it's `nil` or not because the API implementations do a noop if it is nil. I'll comment on that code.

api/analyticsd/client/client.go
```diff
@@ -0,0 +1,122 @@
...
+	return &Client{
+		asc:  c,
+		conn: conn,
+	}, nil
+}
+
+func (c *Client) Identify(ctx context.Context, key string, accountType pb.AccountType, opts ...Option) error {
+	if c == nil {
+		return nil
+	}
```
asutula, line 61: since these APIs only return `error` I added this noop if the client is `nil`. This way you can just call analytics client apis without having to check if the client was created or not.

api/filrewardsd/client/client.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"time"
+
+	analyticspb "github.com/textileio/textile/v2/api/analyticsd/pb"
+	"github.com/textileio/textile/v2/api/filrewardsd/pb"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type Client struct {
```
asutula, line 14: This filrewards client will be used by hub to implement the all filrewards related APIs. That work is coming tomorrow.

api/filrewardsd/main.go
```diff
@@ -0,0 +1,181 @@
...
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/textileio/go-threads/util"
+	"github.com/textileio/textile/v2/api/filrewardsd/service"
+	"github.com/textileio/textile/v2/cmd"
+)
+
+const daemonName = "filrewardsd"
```
asutula, line 17: Fil rewards as a stand along daemon.

api/filrewardsd/pb/filrewards.proto
```diff
@@ -0,0 +1,83 @@
...
+  REWARD_UNSPECIFIED = 0;
+  REWARD_FIRST_KEY_ACCOUNT_CREATED = 1;
+  REWARD_FIRST_KEY_USER_CREATED = 2; // does this fit "register first user?"
+  REWARD_FIRST_ORG_CREATED = 3;
+  REWARD_INITIAL_BILLING_SETUP = 4;
+  REWARD_FIRST_BUCKET_CREATED = 5;
+  REWARD_FIRST_BUCKET_ARCHIVE_CREATED = 6;
+  REWARD_FIRST_MAILBOX_CREATED = 7;
+  REWARD_FIRST_THREAD_DB_CREATED = 8;
+}
```
asutula, line 18: This is the source of truth for fil reward definitions.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,379 @@
...
+type Service struct {
+	col               *mongo.Collection
+	ac                *analytics.Client
+	rewardsCache      map[string]map[pb.Reward]struct{}
+	baseAttoFILReward int32
+	server            *grpc.Server
+}
+
+type Config struct {
+	Listener            net.Listener
```
asutula, line 63: Notice how the listener is abstracted one level up out of the gRPC service. This is because I found a great listener you can use for testing that allows you to run your client and server in the same process using an in memory listener/connection. Great for debugging and tests in general. See `service_test.go` for this type of use.

api/filrewardsd/service/service_test.go
```diff
@@ -0,0 +1,498 @@
...
+			require.Len(t, res.RewardRecords, 2)
+		}
+		startAtToken = res.MoreStartAt
+		more = res.More
+	}
+	require.Equal(t, 3, numPages)
+}
+
+func requireSetup(t *testing.T, ctx context.Context) (pb.FilRewardsServiceClient, func()) {
+	listener := bufconn.Listen(bufSize)
```
asutula, line 411: This is the listener you can use for testing that I described in the previous comment.

api/filrewardsd/service/service_test.go
```diff
@@ -0,0 +1,498 @@
...
+		MongoCollectionName: "filrewards",
+	}
+	s, err := New(ctx, conf)
+	require.NoError(t, err)
+
+	bufDialer := func(context.Context, string) (net.Conn, error) {
+		return listener.Dial()
+	}
+
+	conn, err := grpc.Dial("bufnet", grpc.WithContextDialer(bufDialer), grpc.WithInsecure())
```
asutula, line 426: Then you connect the client to the same listener like this.

asutula: So this is nice for services like this one that is pretty simple. Basically the whole thing is implemented in the service definition file and this allows you to write a thorough unit/integration test of it, with the added bonus that you are running everything basically just like it will run in production (using the client to make calls, using a `listener` and `conn` etc)

core/usage_interceptor.go
```diff
@@ -248,21 +249,30 @@ func (t *Textile) postUsageFunc(ctx context.Context, method string) error {
...
+				"member_email":    account.User.Email,
+			}),
+		); err != nil {
+			log.Errorf("calling analytics track: %v", err)
 		}
-		t.bc.TrackEvent(ctx, account.Owner().Key, account.Owner().Type, true, tp, map[string]string{
-			"member":          account.User.Key.String(),
-			"member_username": account.User.Username,
-			"member_email":    account.User.Email,
-		})
```
asutula, line 265: This looked like a bug that `TrackEvent` was being called with the default `analytics.Event` value if none of the `switch` cases matched. Fixed in the update code.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,379 @@
...
+		rec.AccountType,
+		analyticspb.Event_EVENT_FIL_REWARD_RECORDED,
+		analytics.WithProperties(map[string]interface{}{
+			"reward":               rec.Reward,
+			"factor":               rewardMeta[rec.Reward].factor,
+			"base_atto_fil_reward": s.baseAttoFILReward,
+		}),
+	); err != nil {
+		log.Errorf("calling analytics track: %v", err)
+	}
```
asutula, line 212: @andrewxhill here is the new event that goes to Segment whenever a new reward record is recorded.

asutula: Decided to go with a single event, adding the specifics as properties.

api/analyticsd/service/service.go
andrewxhill: only funny thing i see here is that if the segment queue barks for some reason the reward wont be unlocked.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,379 @@
...
+	if err := s.ac.Track(
+		ctx,
+		rec.Key,
+		rec.AccountType,
+		analyticspb.Event_EVENT_FIL_REWARD_RECORDED,
+		analytics.WithProperties(map[string]interface{}{
+			"reward":               rec.Reward,
+			"factor":               rewardMeta[rec.Reward].factor,
+			"base_atto_fil_reward": s.baseAttoFILReward,
+		}),
```
andrewxhill, line 209: nice, looks great. i'll need to double check that the email templating engine can do basic math... maybe just toss the total on the payload too.

api/analyticsd/service/service.go
asutula: Yea, that is the issue with relying on segment. But technically we just use email for notifications, so the rewards can still show up in any other UI.

api/filrewardsd/service/service.go
asutula: Added a property for the total amount already calculated.

api/analyticsd/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+var (
+	log = logging.Logger(daemonName)
+
+	config = &cmd.Config{
```
asutula, line 21: @textileben here is the exe config for `analyticsd` which is a new service.

api/analyticsd/main.go
```diff
@@ -0,0 +1,158 @@
...
+				Key:      "segment_api_key",
+				DefValue: "",
+			},
+			"segmentPrefix": {
+				Key:      "segment_prefix",
+				DefValue: "",
+			},
+			"filrewardsAddr": {
+				Key:      "filrewards_addr",
+				DefValue: "",
```
asutula, line 48: This is the address of where the other new service `filrewardsd` is running.

asutula: It's just a simple `host:port` string.

api/analyticsd/pb/analytics.proto
```diff
@@ -0,0 +1,58 @@
...
+  EVENT_ORG_INVITE_CREATED = 8;
+  EVENT_GRACE_PERIOD_START = 9;
+  EVENT_GRACE_PERIOD_END = 10;
+  EVENT_BILLING_SETUP = 11;
+  EVENT_BUCKET_CREATED = 12;
+  EVENT_BUCKET_ARCHIVE_CREATED = 13;
+  EVENT_MAILBOX_CREATED = 14;
+  EVENT_THREAD_DB_CREATED = 15;
+  EVENT_FIL_REWARD = 16;
+  EVENT_FIL_CLAIM = 17;
```
asutula, line 25: New events for fil reward and claim.

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,215 @@
...
+
+	orgKey := ""
+	devKey := ""
+	if req.AccountType == pb.AccountType_ACCOUNT_TYPE_ORG {
+		orgKey = req.Key
+		devKey, _ = req.Properties.AsMap()["member"].(string)
+	} else if req.AccountType == pb.AccountType_ACCOUNT_TYPE_DEV {
+		devKey = req.Key
+		orgKey, _ = req.Properties.AsMap()["org_key"].(string)
+	}
```
asutula, line 141: Some updated logic here @andrewxhill. I'm doing everything I can to get the current org and dev keys using the key, account type and properties passed into the `Track()` call. This feels a bit sketchy since it relies on those hard coded property names `member` and `org_key`.

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Don't process the events that are generated by s.frc.ProcessAnalyticsEvent
+	// when a reward or claim is created and sent here for analytics tracking.
+	// Also be sure that we at least have an org key since that is required.
+	if s.frc != nil && req.Event != pb.Event_EVENT_FIL_REWARD && req.Event != pb.Event_EVENT_FIL_CLAIM && orgKey != "" {
```
asutula, line 146: Having a org key is required for calling `filrewards.ProcessAnalyticsEvent`. It's better to have dev key as well, but some events we are interested in will never have a dev key (like billing usage events), so I let that pass if dev key isn't present. The problem, which I mentioned in slack yesterday, is that js client actions that use an org API key will never have a dev key, which subverts some of the logic we'd like to apply during reward unlocking. I'll comment more about that in the fil rewards service code.

api/billingd/main.go
asutula: @textileben here is a new config option for `billingd`

api/billingd/main.go
```diff
@@ -64,14 +68,6 @@ var (
...
 				DefValue: "",
 			},
-			"segmentApiKey": {
-				Key:      "segment.api_key",
-				DefValue: "",
-			},
-			"segmentPrefix": {
-				Key:      "segment.prefix",
-				DefValue: "",
-			},
```
asutula, line 74: @textileben and here are some `billingd` config options that are now gone (moved to `analyticsd`)

api/filrewardsd/main.go
```diff
@@ -0,0 +1,171 @@
...
+
+var (
+	log = logging.Logger(daemonName)
+
+	config = &cmd.Config{
```
asutula, line 22: @textileben here's the exe config for the new `filrewardsd` service

api/filrewardsd/pb/filrewards.proto
```diff
@@ -0,0 +1,101 @@
...
+  int32 claimed = 2;
+  int32 available = 3;
+}
+
+service FilRewardsService {
+  rpc ProcessAnalyticsEvent(ProcessAnalyticsEventRequest) returns (ProcessAnalyticsEventResponse) {}
+  rpc ListRewards(ListRewardsRequest) returns (ListRewardsResponse) {}
+  rpc Claim(ClaimRequest) returns (ClaimResponse) {}
+  rpc ListClaims(ListClaimsRequest) returns (ListClaimsResponse) {}
+  rpc Balance(BalanceRequest) returns (BalanceResponse) {}
```
asutula, line 100: @andrewxhill in response to our convo, check out the new claim related APIs in here.

- `Balance` returns total rewarded, total claimed, and available to claim amounts for an org.
- `ListClaims` lists completed claims for an org and/or dev.
- `Claim` claims any amount of FIL for an org.

There is still some work to do here with tracking `pending` and `complete` status on claims (which will be used once we actually integrate the transferring of FIL). For the the claim just happens instantaneously and is only reflected in the filrewards database (no FIL transferred to any wallet address)

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,571 @@
...
+
+type Service struct {
+	rewardsCol        *mongo.Collection
+	claimsCol         *mongo.Collection
+	ac                *analytics.Client
+	rewardsCacheOrg   map[string]map[pb.RewardType]struct{}
+	rewardsCacheDev   map[string]map[pb.RewardType]struct{}
+	baseAttoFILReward int32
+	server            *grpc.Server
+	semaphores        *nutil.SemaphorePool
```
asutula, line 82: Using semaphores per org when calculating balances and claiming fil.

asutula: To ensure there is only one writer at a time per org.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,571 @@
...
+		},
+		{
+			Keys: bson.D{primitive.E{Key: "created_at", Value: 1}},
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("creating indexes: %v", err)
+	}
+
+	// Populate caches.
```
asutula, line 136: @sanderpick and @textileben here are those in memory caches we were talking about. This is state that isn't in mongo. It is not required, but just a convenient/fast way to avoid reading from the database in certain cases.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,571 @@
...
+
+	if _, exists := s.rewardsCacheOrg[req.OrgKey][t]; exists {
+		// This reward is already granted to the Org so bail.
+		return &pb.ProcessAnalyticsEventResponse{}, nil
+	}
+
+	if _, exists := s.rewardsCacheDev[req.DevKey][t]; exists {
+		// This reward is already granted to the Dev so bail.
+		return &pb.ProcessAnalyticsEventResponse{}, nil
+	}
```
asutula, line 233: @andrewxhill here is problem with the js client not sending dev key when using org api key... we'd like to be able to check that the dev has not unlocked the reward for _any_ org, but if we don't have a dev key, we can't do this check here.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,571 @@
...
+	if err := s.ac.Track(
+		ctx,
+		r.OrgKey,
+		analyticspb.AccountType_ACCOUNT_TYPE_ORG,
+		analyticspb.Event_EVENT_FIL_REWARD,
+		analytics.WithProperties(map[string]interface{}{
+			"type":                 r.Type,
+			"factor":               rewardTypeMeta[r.Type].factor,
+			"base_atto_fil_reward": s.baseAttoFILReward,
+			"amount":               rewardTypeMeta[r.Type].factor * s.baseAttoFILReward,
```
asutula, line 260: Sending the calculated reward amount now to analytics.

api/usersd/pb/usersd.proto
```diff
@@ -191,4 +235,10 @@ service APIService {
     // Archive Retrieval
     rpc ArchiveRetrievalLs(ArchiveRetrievalLsRequest) returns (ArchiveRetrievalLsResponse){}
     rpc ArchiveRetrievalLogs(ArchiveRetrievalLogsRequest) returns (stream ArchiveRetrievalLogsResponse) {}
+
+    // Fil Rewards
+    rpc ListFilRewards(ListFilRewardsRequest) returns (ListFilRewardsResponse) {}
+    rpc ClaimFil(ClaimFilRequest) returns (ClaimFilResponse) {}
+    rpc ListFilClaims(ListFilClaimsRequest) returns (ListFilClaimsResponse) {}
+    rpc FilRewardsBalance(FilRewardsBalanceRequest) returns (FilRewardsBalanceResponse) {}
```
asutula, line 243: @andrewxhill check out the user api versions of the fil rewards API. It is very similar to the filrewards service, but infers certain information from the auth context (org and dev) instead of allowing the values to be provided by the client.

api/usersd/service.go
```diff
@@ -561,6 +567,65 @@ func (s *Service) ArchiveRetrievalLogs(req *pb.ArchiveRetrievalLogsRequest, serv
...
+
+	if org == nil {
+		return nil, status.Error(codes.Unauthenticated, "no org provided")
+	}
+
+	if user != nil && user.Type != mdb.Dev {
+		return nil, status.Error(codes.Unauthenticated, "provided user must be a dev")
+	}
+
+	opts := []filrewards.ListOption{filrewards.WithOrgKeyFilter(org.Key.String())}
```
asutula, line 583: Example of inferring/validating org and user information from the auth context and setting those values to send to the filrewardsd service.

cmd/hubd/main.go
```diff
@@ -92,6 +92,10 @@ var (
 				Key:      "addr.powergate.api",
 				DefValue: "",
 			},
+			"addrAnalyticsApi": {
```
asutula, line 95: @textileben new `hubd` config item.

cmd/hubd/main.go
```diff
@@ -92,6 +92,14 @@ var (
...
+				Key:      "addr.analytics.api",
+				DefValue: "",
+			},
+			"addrFilrewardsApi": {
```
asutula, line 99: @textileben I missed this one before. Hub connects to the fil rewards service so here is the new address config for that.

api/analyticsd/service/service.go
andrewxhill: looks nasty... but it works? :D

api/billingd/service/service.go
andrewxhill: good stuff

api/usersd/pb/usersd.proto
andrewxhill: looks solid

api/analyticsd/service/service.go
asutula: Works as long as we stick to using those key names!

api/billingd/service/service.go
```diff
@@ -1075,7 +1091,15 @@ func (s *Service) reportCustomerUsage(ctx context.Context, cus *Customer) error
 			log.Warn("%s has invalid product key: %s", cus.Key, k)
 		}
 	}
-	s.analytics.Identify(cus.Key, cus.AccountType, false, "", summary)
+	if err := s.analytics.Identify(
+		ctx,
+		cus.Key,
+		cus.AccountType.Pb(),
+		analytics.WithEmail(cus.Email), // this wasn't being inclided before, do we want to?
```
asutula, line 1098: Just checking if we want to include email since it is available.

api/filrewardsd/pb/filrewards.proto
```diff
@@ -0,0 +1,126 @@
...
+  REWARD_TYPE_FIRST_MAILBOX_CREATED = 7;
+  REWARD_TYPE_FIRST_THREAD_DB_CREATED = 8;
+}
+
+message Reward {
+  string org_key = 1;
+  string dev_key = 2;
+  RewardType type = 3;
+  int64 factor = 4;
+  int64 base_nano_fil_reward = 5;
```
asutula, line 25: Notice we are modeling fil rewards FIL amounts as nano FIL. This is useful because it allows the amounts of FIL we're dealing with to fit easily into an `int64` which is supported by the whole stack from mongo to gRPC to the client.

api/filrewardsd/pb/filrewards.proto
```diff
@@ -0,0 +1,126 @@
...
+  google.protobuf.Timestamp created_at = 6;
+}
+
+enum ClaimState {
+  CLAIM_STATE_UNSPECIFIED = 0;
+  CLAIM_STATE_PENDING = 1;
+  CLAIM_STATE_COMPLETE = 2;
+  CLAIM_STATE_FAILED = 3;
+}
```
asutula, line 34: These claim states will be used depending on how the transfer of FIL goes to the target address during a claim.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,658 @@
...
+var rewardTypeMeta = map[pb.RewardType]meta{
+	pb.RewardType_REWARD_TYPE_FIRST_KEY_ACCOUNT_CREATED:    {factor: 3},
+	pb.RewardType_REWARD_TYPE_FIRST_KEY_USER_CREATED:       {factor: 1},
+	pb.RewardType_REWARD_TYPE_FIRST_ORG_CREATED:            {factor: 3},
+	pb.RewardType_REWARD_TYPE_INITIAL_BILLING_SETUP:        {factor: 1},
+	pb.RewardType_REWARD_TYPE_FIRST_BUCKET_CREATED:         {factor: 2},
+	pb.RewardType_REWARD_TYPE_FIRST_BUCKET_ARCHIVE_CREATED: {factor: 2},
+	pb.RewardType_REWARD_TYPE_FIRST_MAILBOX_CREATED:        {factor: 1},
+	pb.RewardType_REWARD_TYPE_FIRST_THREAD_DB_CREATED:      {factor: 1},
+}
```
asutula, line 43: TODO: Decide what the base fil reward should be and decide what all the factors should be.

asutula: Also TODO: Implement more analytics events to trigger more/new fil rewards events.

api/billingd/service/service.go
andrewxhill: should be safe either way

api/filrewardsd/service/service.go
asutula: I've changed my thinking about this. Basically, if a dev wants to use an org API key in the js client, that org key is free to unlock any rewards that haven't been previously unlocked for that org. It's like a bonus for the user doing something advanced like using the js client.

api/usersd/client/client.go
```diff
@@ -333,3 +334,56 @@ func (c *Client) ArchiveRetrievalLogs(ctx context.Context, id string, ch chan<-
...
+	return res.Claims, res.More, t, nil
+}
+
+func (c *Client) FilRewardsBalance(ctx context.Context) (*pb.FilRewardsBalanceResponse, error) {
+	res, err := c.c.FilRewardsBalance(ctx, &pb.FilRewardsBalanceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("calling fil rewards balance rpc: %v", err)
+	}
+	return res, nil
+}
```
asutula, line 389: New users client methods for fil rewards.

api/filrewardsd/service/service.go
andrewxhill: wouldn't that would allow me to create unlimited orgs for unlimited rewards

cmd/hubd/docker-compose-dev.yml
```diff
@@ -63,6 +65,45 @@ services:
...
+      - FILREWARDS_ANALYTICS_ADDR=analytics:6006
+      - FILREWARDS_BASE_NANO_FIL_REWARD
+    ports:
+      - "127.0.0.1:7006:7006"
+    security_opt:
+      - "seccomp:unconfined"
+    cap_add:
+      - SYS_PTRACE
+    depends_on:
+      - mongo
```
asutula, line 106: Been running this locally, everything seems to work

api/filrewardsd/service/service.go
asutula: Each org can only be rewarded each reward type once no matter who/where the reward type was unlocked from. So it's not unlimited for one org. But yes, you could create unlimited number of orgs and unlock all rewards for each of them using the org API key. This is the problem I was surfacing in my original comment above.

asutula: The only way to prevent that is to somehow not grant fil rewards unless we have the dev key _and_ org key.

api/filrewardsd/pb/filrewards.proto
```diff
@@ -0,0 +1,126 @@
...
+  bool ascending = 4;
+  int64 more_token = 5;
+  int64 limit = 6;
+}
+
+message ListRewardsResponse {
+  repeated Reward rewards = 1;
+  bool more = 2;
+  int64 more_token = 3;
+}
```
asutula, line 71: cc @jsign this is what I settled on for communicating "more" and the needed token for the next requests. Idea is the token is, to the consumer, a meaningless integer they just need to pass in the next request. Turns out I am just using unix nanoseconds for the token, but the api doesn't need to be concerned with that.

jsign: Any reason to not use Skip instead of this opaque token? I'm assuming that you're backed by MongoDB in this daemon.

asutula: Just the potential for repeated records returned in the case that you sort descending (the default) and there are new records created in the time between page requests.

jsign: Sounds good. The sorting criteria is always `CreatedAt` or som fixed and immutable date field?

asutula: In this case, yes, always `CreatedAt`. I'm not claiming this is a universal strategy, just one that works in this case. Things change if more complex queries are supported.

jsign: Yeah, sounds good... just exploring your case to double-check. I think for this scenario that looks fine, at least you can buy the consistency in the "pages" if some new things get created.

The "Skip-Limit" style seems more suited for some more variable sort scenario (despite you can get some drift in pages if things change while fetching next pages). "Skip" is also nice to skip pages and avoid forcing sequential exploration.

Taking this thread as a coffee talk to share ideas about paging patterns.

asutula: Yea makes sense, I agree skip is the tool for general purpose paging, especially when allowing sorting by different keys.

sanderpick: @asutula there are some existing APIs that do a similar thing: https://github.com/textileio/textile/blob/master/api/billingd/pb/billingd.proto#L73. The `next_offset` sounds a lot like your `more_token`. Seems like we should be aiming for consistency for things like this, even if these are internal APIs.

sanderpick: Not suggesting a change here... just something to think about for future APIs

asutula: Yea @sanderpick consistency is worth striving for, I agree, but also needs to be weighed by API correctness/accuracy. In this specific case (listing financial transactions basically), I went for the method that will never miss or skip any records (see comments about "drift" with pages above). If we want to make "skip" the consistent standard, I could remove the option for sorting descending and ascending sorting would be the only option (which could remove the drift problem).

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,224 @@
...
+	select {
+	case <-t.C:
+		s.server.Stop()
+	case <-stopped:
+		t.Stop()
+	}
+	log.Info("gRPC server stopped")
+}
+
+func eventToString(e pb.Event) string {
```
sanderpick, line 183: `e.String()` wouldn't cut it here?

api/filrewardsd/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+# This shared lib (part of glibc) doesn't seem to be included with busybox.
+COPY --from=0 /lib/*-linux-gnu*/libdl.so.2 /lib/
+
+# Copy over SSL libraries.
+COPY --from=0 /usr/lib/*-linux-gnu*/libssl.so* /usr/lib/
+COPY --from=0 /usr/lib/*-linux-gnu*/libcrypto.so* /usr/lib/
+
+# listenAddrs
+EXPOSE 7006
```
sanderpick, line 63: 5000?

api/filrewardsd/Dockerfile.dev
```diff
@@ -0,0 +1,36 @@
...
+FROM debian:buster
+LABEL maintainer="Textile <contact@textile.io>"
+
+ENV SRC_DIR /textile
+COPY --from=0 /go/bin/dlv /usr/local/bin/dlv
+COPY --from=0 /etc/ssl/certs /etc/ssl/certs
+COPY --from=0 $SRC_DIR/filrewardsd /usr/local/bin/filrewardsd
+
+EXPOSE 10006
+EXPOSE 8010
```
sanderpick, line 29: 5000? No gateway

api/filrewardsd/main.go
```diff
@@ -0,0 +1,171 @@
...
+				Key:      "debug",
+				DefValue: false,
+			},
+			"logFile": {
+				Key:      "log_file",
+				DefValue: "", // no log file
+			},
+			"listenAddr": {
+				Key:      "listen_addr",
+				DefValue: "/ip4/127.0.0.1/tcp/7006",
```
sanderpick, line 37: 5000?

api/filrewardsd/pb/filrewards.proto
sanderpick: The sorting options sound useful. All good.

api/usersd/service.go
```diff
@@ -561,6 +566,146 @@ func (s *Service) ArchiveRetrievalLogs(req *pb.ArchiveRetrievalLogsRequest, serv
 	return nil
 }
 
+func (s *Service) ListFilRewards(ctx context.Context, req *pb.ListFilRewardsRequest) (*pb.ListFilRewardsResponse, error) {
+	account, _ := mdb.AccountFromContext(ctx)
+	org := account.Org
+	user := account.User
+
+	if org == nil {
```
sanderpick, line 574: Seems like all these methods should be on the hub service if we're requiring an org. All the existing `userd` methods are accessible to devs and users. Maybe you put them here for something in the future?

mongodb/accounts.go
```diff
@@ -50,6 +51,18 @@ const (
 	User
 )
 
+func (a AccountType) Pb() pb.AccountType {
```
sanderpick, line 54: 👍

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,670 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing claim id: %v", err)
+	}
+
+	res := s.claimsCol.FindOneAndUpdate(
+		ctx,
+		bson.M{"_id": objID},
+		bson.M{"$set": bson.M{"state": state, "txn_cid": req.TxnCid, "failure_message": req.FailureMessage, "updated_at": time.Now()}},
+	)
+	if res.Err() == mongo.ErrNoDocuments {
```
sanderpick, line 441: Nit: `errors.Is()` is a nice way to compare errors.

api/filrewardsd/service/service.go
```diff
@@ -0,0 +1,670 @@
...
+		Claimed:   totalClaimed,
+		Available: totalRewarded - totalPending - totalClaimed,
+	}, nil
+}
+
+func (s *Service) Close() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	if err := s.rewardsCol.Database().Client().Disconnect(ctx); err != nil {
+		log.Errorf("disconnecting mongo client: %s", err)
```
sanderpick, line 562: Return error?

api/filrewardsd/main.go
```diff
@@ -0,0 +1,171 @@
...
+			Debug:             debug,
+		}
+		api, err := service.New(ctx, conf)
+		cmd.ErrCheck(err)
+
+		fmt.Println("Welcome to Hub Filrewards!")
+
+		cmd.HandleInterrupt(func() {
+			api.Close()
+			cancel()
```
sanderpick, line 168: Since the service has a `Close` method, you could create the context internally, then cancel it in `Close`. That way the caller doesn't have to close and cancel.

api/filrewardsd/service/service_test.go
```diff
@@ -0,0 +1,745 @@
+package service
```
sanderpick, line 1: Nice tests!

api/filrewardsd/service/service.go
sanderpick: We could try some of the follow mitigations now or in the future:
- Require that an org be of a certain age before any rewards are unlocked (1 week or similar).
- Require that an org receiving rewards be owned by a dev that has not received awards, directly and/or via a different org they own.

sanderpick (approved):
LGTM! Really nice work @asutula. I left a few nit comments and some questions to think about.

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,224 @@
...
+		devKey, _ = req.Properties.AsMap()["member"].(string)
+	} else if req.AccountType == pb.AccountType_ACCOUNT_TYPE_DEV {
+		devKey = req.Key
+		orgKey, _ = req.Properties.AsMap()["org_key"].(string)
+	}
+
+	// Don't process the events that are generated by s.frc.ProcessAnalyticsEvent
+	// when a reward or claim is created and sent here for analytics tracking.
+	// Also be sure that we at least have an org key since that is required.
+	if s.frc != nil && req.Event != pb.Event_EVENT_FIL_REWARD && req.Event != pb.Event_EVENT_FIL_CLAIM && req.Event != pb.Event_EVENT_FIL_FINALIZE_CLAIM && orgKey != "" {
```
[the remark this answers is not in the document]
asutula: Good call. I'll fix it for now, will look at installing some linter that will be shared for anyone who works on the project.

api/analyticsd/service/service.go
```diff
@@ -0,0 +1,224 @@
...
+			return nil, status.Errorf(codes.Internal, "processing analytics event for filrewards: %v", err)
+		}
+	}
+
+	return &pb.TrackResponse{}, nil
+}
+
+func (s *Service) Close() {
+	if err := s.segment.Close(); err != nil {
+		log.Errorf("closing segment client: %s", err)
```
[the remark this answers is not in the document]
asutula: Always been curious about the right pattern for that. In the case of `Close`, it seems you don't want to immediately return an error. If you did, then all the remaining things that still need to be shut down don't get shut down. I haven't messed with Go error wrapping much, but that seems useful here... You could accumulate all the errors that happen in `Close` into a wrapped error and return that? Maybe i'm not following what you're saying.

api/analyticsd/service/service.go
asutula: It could for sure. The string would look like `EVENT_BILLING_SETUP`. I just kept the strings that @andrewxhill was already using. Any options @andrewxhill?

api/filrewardsd/service/service.go
asutula: Nice ideas. The org age thing is nice. Thoughts on that @andrewxhill?

The second bullet is basically the way it already works... A reward can only be unlocked if the dev hasn't unlocked the same reward for any/all orgs they belong to. And the reward can only be unlocked if it hasn't already been unlocked for the org.

api/filrewardsd/service/service.go
asutula: Same thought/quesion as in the other file where you mentioned this.

api/analyticsd/service/service.go
andrewxhill: for external apis, camelcase or snake is fine. there are some already being collected that are snake though, so would be nice to maintain that unless it's a real drag.

api/analyticsd/service/service.go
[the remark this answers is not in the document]
sanderpick: It's idiomatic to implement `io.Closer`, which returns an error. Then it's up to the implementation to handle which errors should return, e.g., threads `net` logs a gateway shutdown error, and collects errors from other components: https://github.com/textileio/go-threads/blob/master/net/net.go#L165

api/filrewardsd/main.go
asutula: Ok, but what about the context that needs to be passed to `New` on line 158 above?

api/usersd/service.go
asutula: No, I asked you where to put them and you told me `usersd` haha. If it's better on `hubd`, that's great and I'll move them.

api/filrewardsd/main.go
sanderpick: Well, you wouldn't pass a context there. Create the context _inside_ the service.

api/analyticsd/service/service.go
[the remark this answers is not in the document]
asutula: Perfect. Thanks for the example.

api/analyticsd/service/service.go
asutula: I guess using `String()` outputs upper snake case. We could also convert those strings to all lower case. I guess the main issue/question is that the resulting string is slightly different than what we're already collecting. Is it important to keep the exact event names we are already collecting? If not, it would be nice to not have to maintain manually this map of events to names.

api/filrewardsd/main.go
asutula: Ah now I see what you're saying. Cool, will update.

api/usersd/service.go
asutula: Ok will move.

api/analyticsd/service/service.go
andrewxhill: keeping the same would be nice, but if we are going to change them... now is the time for sure. so just do whatever makes sense here and we'll see

andrewxhill: but yeah, avoiding changes going forward (or now) will be helpful

api/filrewardsd/service/service.go
asutula: Updated everywhere.

api/analyticsd/service/service.go
asutula: Ok decided to go with the lower case version of `s.String()` which outputs stuff like `event_billing_setup`. Lower snake case and we don't have to maintain the manual lookup and we make the change now, not later.

api/usersd/service.go
asutula: Moved to hubd service and client.

asutula:
Ok, I think the last outstanding thing here is to possibly implement an account age requirement before beginning to unlock rewards. Waiting on an opinion from @andrewxhill on that.

Heads up @textileben, I updated the `analyticsd` and `filrewardsd` services to listen on port 5000.

asutula:
Closing this for now, will keep the branch around for future use if needed.

PR #481, opened 2021-01-25
Hide progress on windows

sanderpick:
- Don't use progress bar on windows
- Fix `buck init` (`unfreeze` flag used but not defined)

### DIFF
--- a/cmd/buck/cli/init.go
+++ b/cmd/buck/cli/init.go
@@ -73,8 +73,10 @@ Use the '--hard' flag to discard all local changes.
 			cmd.Fatal(errors.New("--cid cannot be used with an existing bucket"))
 		}
 
-		unfreeze, err := c.Flags().GetBool("unfreeze")
-		cmd.ErrCheck(err)
+		// (jsign): re-enable when this feature is usable in mainnet.
+		//unfreeze, err := c.Flags().GetBool("unfreeze")
+		//cmd.ErrCheck(err)
+		var unfreeze bool
 		if unfreeze && xcid == cid.Undef {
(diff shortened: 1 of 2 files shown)


sanderpick:
I'm curious why the local buckets tests never run from cache. All the other one do. It's always run on the same machine.

jsign (approved):

cmd/buck/cli/init.go
```diff
@@ -73,8 +73,10 @@ Use the '--hard' flag to discard all local changes.
 			cmd.Fatal(errors.New("--cid cannot be used with an existing bucket"))
 		}
 
-		unfreeze, err := c.Flags().GetBool("unfreeze")
-		cmd.ErrCheck(err)
+		// (jsign): re-enable when this feature is usable in mainnet.
+		//unfreeze, err := c.Flags().GetBool("unfreeze")
+		//cmd.ErrCheck(err)
+		var unfreeze bool
```
jsign, line 79: I guess we need to include some test to compile all the project to catch these things.

jsign:
> I'm curious why the local buckets tests never run from cache. All the other one do. It's always run on the same machine.

Yeah, weird. I tested locally and should be cached. Nothing popups in my mind... maybe some ghrunner touching some files.

cmd/buck/cli/init.go
sanderpick: We have those tests (added a month ago or so). It did compile, but `init` would throw a runtime error.

jsign: Ah, I read this diff wrong. Too early in the morning.

PR #482, opened 2021-01-25
feat(buck): extend command existing with the format flag

mmuenker:
Add the flag --format. The flag --format accepts a string to define the output format.
Add support for the format 'json'. If the format is set to json the link struct is printed as a parsable JSON object.


sanderpick:

PR #483, opened 2021-01-25
feat(buck): add the non interactive flag to the init command

mmuenker:
Add the flag --non-interactive. Foreach interactive option the default value will be used

The `--non-interactive` flag allows the execution of the init command in a headless environment. Like a CI.

### DIFF
--- a/cmd/buck/cli/cli.go
+++ b/cmd/buck/cli/cli.go
@@ -60,6 +60,7 @@ func Init(baseCmd *cobra.Command) {
 	initCmd.Flags().Bool("hard", false, "Discards all local changes if true")
 	initCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
 	initCmd.Flags().BoolP("quiet", "q", false, "Write minimal output")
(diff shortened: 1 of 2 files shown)


sanderpick:
Hey @mmuenker, you can achieve the same behavior by using the existing flags. For example, supply the `name` flag to skip the name prompt, and use the `private` flag to skip the encryption prompt.

mmuenker:
> Hey @mmuenker, you can achieve the same behavior by using the existing flags. For example, supply the `name` flag to skip the name prompt, and use the `private` flag to skip the encryption prompt.

That is correct, but this is very inconvenient.
Generally, it would be a good idea to introduces the `--non-interactive` flag on a global level. The user could be sure that every prompt is skipped and the default value is used. Else the user has to consult the documentation to finger out which flags must be set to prevent the prompts.

sanderpick:
Hey @mmuenker, I hear you. The problem with a catchall "non interactive" flag, is that if we change any of the defaults, users' scripts will be broken. I would rather users go to the trouble of consulting `--help` to learn how to provide values explicitly.

sanderpick:

PR #484, opened 2021-01-25
feat(buck): extend command init with the format flag

mmuenker:
Add the flag --format. The flag --format accepts a string to define the output format.
Add support for the format 'json'.


sanderpick:
Nice! Instead of adding the `format` flag to each `buck` subcommand, you can add it to the root command, which will make it available everywhere.

mmuenker:
@sanderpick I have updated the pull requests #484 and #482 to support a global `--format` flag

sanderpick (commented):
Hey @mmuenker, sorry for the delay here. This looks good for the init command. If we're not going to handle the format flag for the other commands just yet, I think it makes more sense to add it separately to each sub command.

sanderpick:

PR #489, opened 2021-02-04
fix(bug): rm user since possibly nil

andrewxhill:

### DIFF
--- a/core/usage_interceptor.go
+++ b/core/usage_interceptor.go
@@ -251,7 +251,7 @@ func (t *Textile) postUsageFunc(ctx context.Context, method string) error {
 	if t.bc != nil {
 
 		payload := map[string]string{}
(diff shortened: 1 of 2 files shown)


asutula:
Other option would be to optionally create the properties if user exists?

core/usage_interceptor.go
```diff
@@ -249,20 +249,22 @@ func (t *Textile) postUsageFunc(ctx context.Context, method string) error {
 	}
 
 	if t.bc != nil {
-		var tp analytics.Event
+
+		payload := map[string]string{}
+		if account.User == nil {
```
asutula, line 254: should be `!=`

asutula: above as well

asutula (approved):

PR #493, opened 2021-02-09
chore(ipns): add repub cron

andrewxhill:


api/bucketsd/service.go
```diff
@@ -122,6 +127,20 @@ func (l buckLock) Key() string {
 	return string(l)
 }
 
+// StartRepublishing initializes a key republishing cron
+func (s *Service) StartRepublishing(schedule string, session string) error {
+	s.republisher = cron.New()
```
andrewxhill, line 132: create at time of start so i didn't need to add a NewService method at this time.

api/bucketsd/service.go
```diff
@@ -146,6 +165,30 @@ func (s *Service) List(ctx context.Context, _ *pb.ListRequest) (*pb.ListResponse
...
+			Path: "/",
+		})
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+		s.IPNSManager.Publish(path.New(pth.Item.Cid), key.Cid)
+	}
+	log.Infof("republishing finished %d records => %v", len(keys), time.Since(start))
+	return nil
```
andrewxhill, line 189: crux of it above

core/core.go
```diff
@@ -524,6 +525,8 @@ func NewTextile(ctx context.Context, conf Config, opts ...Option) (*Textile, err
 
 	// Start pulling threads
 	t.tn.StartPulling()
+	// Start republishing ipns keys
+	bs.StartRepublishing("*/5 * * * *", t.internalHubSession)
```
andrewxhill, line 529: will make this a flag so we can tune with the expiration settings

sanderpick: You'll have to stop it too in the core `Close` method. Something like,
```
ctx := cron.Stop()
<-ctx.Done()
log.Info("republisher was shutdown")
```

If you created a `Close` method on `Service` and put something like the above in there, that could be called externally in the core `Close` method.

ipns/ipns.go
```diff
@@ -50,13 +50,15 @@ func NewManager(keys *mdb.IPNSKeys, keyAPI iface.KeyAPI, nameAPI iface.NameAPI,
...
+	m := &Manager{
 		keys:     keys,
 		keyAPI:   keyAPI,
 		nameAPI:  nameAPI,
 		ctxs:     make(map[string]context.CancelFunc),
 		keyLocks: make(map[string]chan struct{}),
-	}, nil
+	}
+
+	return m, nil
```
sanderpick, line 61: Any reason to do that?

ipns/ipns.go
```diff
@@ -120,6 +122,7 @@ func (m *Manager) Publish(pth path.Path, keyID string) {
 			cancel, ok := m.ctxs[keyID]
 			m.ctxsLock.Unlock()
 			if ok {
+				log.Debugf("success path %s: for key %s", pth, keyID)
```
sanderpick, line 125: Success is already logged below in `publishUnsafe`

api/bucketsd/service.go
sanderpick: Hmm, yeah it's a weird thing to do because the service doesn't really "own" the threads. There won't be a session/api key to add to the context since this is an internal operation (not user initiated). 

Rather than using `ListPath`, you can use `s.Buckets.List`:
```
list, err := s.Buckets.List(ctx, threadID, &db.Query{}, &tdb.Bucket{}, tdb.WithToken(dbToken))
if err != nil {
	return nil, err
}
bucks := list.([]*tdb.Bucket)
// loop over buckets
```

We're not providing a thread token here. So, you may be blocked from reading that thread... if you're not, that's only cause there's still some ACL related todos in `go-threads`. 

Another option is to modify the IPNS record mongo collection to store the most recent successfully published path. Then this fix will only work after a bucket it pushed to, which is probably fine. That would prevent future headaches with thread tokens as well.

andrewxhill: yep, cool. was trying to avoid the model change. but looking back now it seems so much cleaner. coming up

ipns/ipns.go
andrewxhill: yeah, these are all the tidy things to wrap up after choosing the right strategy. taking out of wip here in a minute

api/apitest/apitest.go
```diff
@@ -52,6 +52,7 @@ func DefaultTextileConfig(t util.TestingTWithCleanup) core.Config {
 		AddrIPFSAPI:               GetIPFSApiAddr(),
 		AddrGatewayHost:           util.MustParseAddr(fmt.Sprintf("/ip4/127.0.0.1/tcp/%d", gatewayPort)),
 		AddrGatewayURL:            fmt.Sprintf("http://127.0.0.1:%d", gatewayPort),
+		IPNSRepublishCron:         "0 1 * * *",
```
andrewxhill, line 55: could also leave it "" and not run cron if set that way

mongodb/migrations/migrations.go
```diff
@@ -154,12 +154,39 @@ var m003 = migrate.Migration{
 	},
 }
 
+var m004 = migrate.Migration{
+	Version:     4,
+	Description: "set empty path field on all ipnskeys",
```
andrewxhill, line 159: could have detected it missing on every lookup, but felt cleaner to do this once and forget about it

mongodb/ipnskeys.go
```diff
@@ -13,6 +13,7 @@ import (
 type IPNSKey struct {
 	Name      string
 	Cid       string
+	Path      string
 	ThreadID  thread.ID
 	CreatedAt time.Time
 }
```
andrewxhill, line 19: wondering if it would be handy to grab UpdatedAt? could be cool someday to hold a path history to show in the dash, though probably lives on the bucket itself.

ipns/ipns.go
```diff
@@ -169,3 +202,23 @@ func (m *Manager) getSemaphore(key string) chan struct{} {
...
+	ctx, cancel := context.WithTimeout(context.Background(), listKeysTimeout)
+	defer cancel()
+	start := time.Now()
+	keys, err := m.keys.List(ctx)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	withPath := 0
+	for _, key := range keys {
```
sanderpick, line 216: Doing these one at a time will likely take hours, maybe more than 4 hours, which may be the cause of our repub pain in general (assuming IPFS is doing something similar internally). We can at least do them in batches 50 or something (just picked that number). Here's an example of this kind of parallel loop processing in batches: https://github.com/textileio/textile/blob/edbdd05058fccc40e8c23817bf4e44288e59208a/buckets/local/pull.go#L146 (down to line 164).

mongodb/migrations/migrations_test.go
```diff
@@ -149,6 +149,40 @@ func TestMigrations_m003(t *testing.T) {
...
+	// Run up
+	err = migrate.NewMigrate(db, m004).Up(migrate.AllAvailable)
+	require.NoError(t, err)
+
+	res := db.Collection("ipnskeys").FindOne(ctx, bson.M{})
+	require.NoError(t, res.Err())
+	var key bson.M
+	err = res.Decode(&key)
+	require.NoError(t, err)
+	assert.Nil(t, key["path"])
```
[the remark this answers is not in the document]
andrewxhill: good catch

ipns/ipns.go
andrewxhill: just wrap this guy in a batch like your link, or are you thinking i should batch up something lower like this step https://github.com/textileio/textile/pull/493/files#diff-565b03391ddf3dcb3b46674b1eec6d37ca610e29cefb834690fab96955ca89fcR148

sanderpick: Yeah just batch the calls to `m.publish`

andrewxhill: cool, done

cmd/buckd/main.go
```diff
@@ -82,6 +82,10 @@ var (
 				Key:      "addr.powergate.api",
 				DefValue: "",
 			},
+			"ipnsRepublishSchedule": {
+				Key:      "ipns.republish.schedule",
```
sanderpick, line 86: Nit picking here, but this could be `ipns.republish_schedule`, unless we think there will be other sub config values under `ipns.republish`

ipns/ipns.go
```diff
@@ -169,3 +206,41 @@ func (m *Manager) getSemaphore(key string) chan struct{} {
...
 	return ptl
 }
+
+func (m *Manager) republish() error {
+	ctx, cancel := context.WithTimeout(context.Background(), listKeysTimeout)
+	defer cancel()
+	start := time.Now()
+	keys, err := m.keys.List(ctx)
+	if err != nil {
+		log.Error(err)
```
sanderpick, line 216: Looks like this already logged from the caller (in the cron function).

sanderpick (approved):
LGTM! Couple small nits but nothing crazy.

PR #498, opened 2021-02-17
feat(buck): mv command

andrewxhill:


api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+}
+
+// isEqualPath returns true if dest equals src
+func isEqualPath(src, dest string) bool {
+	return filepath.Clean(src) == filepath.Clean(dest)
+}
+
+// isSubPath returns true if chld is a subpath of pth
+func isSubPath(src, dest string) (bool, error) {
+	p, err := filepath.Rel(src, dest)
```
sanderpick, line 1104: You should be able to use `path` here since ipfs paths always use "/"

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("path cleanup failed: %v", err)
+		}
+		for _, chld := range item.Item.Items {
+			if strings.Compare(chld.Name, buckets.SeedName) == 0 || isEqualPath(chld.Path, ipfsDest) {
+				continue
+			}
+
+			sp := stripNodePath(chld.Path)
```
sanderpick, line 1065: Is `sp` the equivalent of `path.Base(chld.Path)`?

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	if buck.IsPrivate() {
+		meta, _, ok := buck.GetMetadataForPath(cleanPth, false)
+		if !ok {
+			return nil, fmt.Errorf("error getting metadata")
+		}
+		buck.SetMetadataAtPath(cleanDst, meta)
+		if err = s.Buckets.Verify(ctx, dbID, buck, tdb.WithToken(dbToken)); err != nil {
+			return nil, fmt.Errorf("verifying bucket update: %v", err)
+		}
+	}
```
sanderpick, line 1019: I think you can skip this since `setPath` is already doing the verification step.

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
 	}
 	dbToken, _ := thread.TokenFromContext(ctx)
 
-	destPath := cleanPath(req.Path)
-	bootCid, err := cid.Decode(req.Cid)
+	if isEqualPath(req.Path, req.Destination) {
```
sanderpick, line 976: Same as `cleanPth` == `cleanDst` below?

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	return filepath.Clean(src) == filepath.Clean(dest)
+}
+
+// isSubPath returns true if chld is a subpath of pth
+func isSubPath(src, dest string) (bool, error) {
+	p, err := filepath.Rel(src, dest)
+	if err != nil {
+		return false, err
+	}
+	return !strings.Contains(p, ".."), nil
```
sanderpick, line 1108: The clients should be only sending absolute paths. I think that would reduce some complexity here.

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	// a nice absolute path to skip Rel errors
+	ipfsPath := gopath.Join("/ipfs", n.Cid().String(), cleanPth)
+	ipfsDest := gopath.Join("/ipfs", n.Cid().String(), cleanDst)
+
+	// If "a/b" => "a/" no cleanup is needed
+	parent, err := isSubPath(ipfsDest, ipfsPath)
+	if err != nil {
+		return nil, fmt.Errorf("path cleanup failed: %v", err)
+	}
+	if parent == true {
```
sanderpick, line 1043: `if parent {`

api/bucketsd/service.go
```diff
@@ -970,16 +973,175 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+		return &pb.MovePathResponse{}, nil
+	}
+
+	// If "a/" => "a/b" cleanup each leaf in "a" that isn't "b" (skipping .textileseed)
+	subdir, err := isSubPath(ipfsPath, ipfsDest)
+	if err != nil {
+		return nil, fmt.Errorf("path cleanup failed: %v", err)
+	}
+	if subdir == true {
+		item, err := s.ListPath(ctx, &pb.ListPathRequest{
```
sanderpick, line 1053: Using this here is nice since it encapsulates the encrypted vs. non-encrypted handling, but will make the method slower by doing more roundtrips to threads. We could leave it for this first pass though.

sanderpick (commented):
I added some comments. Looking good! From a quick pass, I this should work with encrypted and non-encrypted buckets... but could be missing something.

api/bucketsd/service.go
andrewxhill: chld.Path here is of form `/ipfs/sdlkfjs/path/to/file.jpg`. so clean path just strips `/ipfs/sdlkfjs/`

api/bucketsd/service.go
andrewxhill: expand on that, not sure i follow the suggestion

api/bucketsd/service.go
andrewxhill: this function just says, 

1. if two paths, `/ipfs/lskdjfs/a/b/g` and `/ipfs/lskdjfs/a/b/c` 
2. Rel will contain `..` so it isn't a subdir, return false

but 

1. if two paths, `/ipfs/lskdjfs/a/b/` and `/ipfs/lskdjfs/a/b/c` 
2. Rel will be `c/` and not contain `..` so it is a subdir
3. return true

using the `/ipfs` base path works nicely here since they will never error (e.g. `a/b/c` to `b/` will error as no path is known).

api/bucketsd/service.go
andrewxhill:
```
chld.Path = "/ipfs/bafybeihzajh4qqs7tugd53fsbvu52ee33njyas77jmyo4o6n26ysplienm/a/b/file1.jpg"

stripNodePath(chld.Path) => a/b/file2.jpg
path.Base(chld.Path) => file2.jpg
```

api/bucketsd/service.go
sanderpick: Never mind, `path` doesn't have a rel method. Though I think that's probably because you don't need the complexity of rel here... you can just compare prefixes, i.e., if src is a prefix of dest, you're in a subpath.

sanderpick: (`filepath` is for working with files on a filesystem, that can be addressed in complex ways with .., ~, etc. wich vary from os to os)

api/bucketsd/service.go
sanderpick: Ah I see! Okay, but If you skip rel, I think you could avoid creating the ipfs paths.

andrewxhill: yeah, makes sense

andrewxhill: HasPrefix does the trick

api/bucketsd/service.go
```diff
@@ -970,16 +972,143 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	n, err := s.getNodeAtPath(ctx, pth, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = s.setPath(ctx, &pb.SetPathRequest{
+		Key:  req.Key,
+		Path: cleanDst,
+		Cid:  n.Cid().String(),
+	}, dbID, dbToken, cleanDst)
```
andrewxhill, line 1005: there is no way this is going to work on an encrypted bucket, right @sanderpick ? the source CID is for the encrypted dag... 

it will fail down at line 1029

the `err` is `could not resolve path: /ipfs/bafybeieq2ts26nrhfldfdkkafdqssocijpdtie746nwvih54m6zdzy2cya/a/.textileseed`.

if i pop a listPath of the destination in at line 1029 before returning, i see

```
item:{cid:"bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47bgf354" name:"a" path:"/ipfs/bafybeib3l6pbh55k6iparpsebbkmkabzdxtt5vkq3pempnlgsvqpghta7y/a" size:4 is_dir:true metadata:{updated_at:1613596140630032000}} root:{key:"bafzbeiend6uzkwh2dadgl3h6v7smnmlqp2hxkcn7vcwpy4ekaaibuqmf54" path:"/ipfs/bafybeib3l6pbh55k6iparpsebbkmkabzdxtt5vkq3pempnlgsvqpghta7y" created_at:1613596140341415000 updated_at:1613596140630032000 thread:"bafk26pxojma4xalp4x2sqhnmgzuhaa6phywo4qlpwkwxfdmo4ctofzi" owner:"bbaareic6kv6huvr7fbhdajm2tv42iriybrzqo6exi3iylsw7bwthv4az7u" version:1 metadata:{roles:{key:"bbaareic6kv6huvr7fbhdajm2tv42iriybrzqo6exi3iylsw7bwthv4az7u" value:PATH_ACCESS_ROLE_ADMIN} updated_at:1613596140341415000 key:"biq/HrgkeftcYQaWlc52R8guyCgBrBOeGmIc3uJGeGgRgK9AayFrN3L1aYHcxquqXgZiDqU0GnQKeHBgIkSTfg=="} link_key:"5RYdG/U3nhk/Vt7gy1Yq7qAWL0gbmkTNO3Ht/K4QZErPp1L54vPIuzvyjnNJIMmKHapfUa98UEygom5GKoj9Gg==" path_metadata:{key:"" value:{roles:{key:"bbaareic6kv6huvr7fbhdajm2tv42iriybrzqo6exi3iylsw7bwthv4az7u" value:PATH_ACCESS_ROLE_ADMIN} updated_at:1613596140341415000 key:"biq/HrgkeftcYQaWlc52R8guyCgBrBOeGmIc3uJGeGgRgK9AayFrN3L1aYHcxquqXgZiDqU0GnQKeHBgIkSTfg=="}} path_metadata:{key:".textileseed" value:{updated_at:1613596140341415000}} path_metadata:{key:"a" value:{updated_at:1613596140630032000}} path_metadata:{key:"file1.jpg" value:{updated_at:1613596140452829000}} path_metadata:{key:"file2.jpg" value:{updated_at:1613596140463238000}} archives:{current:{}}}
```

so i can see it there in the metadata... but i dunnooo.... maybe result at that node is actually going to be doubly encrypted (original encryption, plus setpath adds another encryption...) not sure. will start walking through the encrypted flow more closely now

sanderpick: I'll take a look in a bit... I think you'll have to use a more internal method (or slightly modified one) to _add the node at the path_, not the public API version (`SetPath`).

sanderpick: Since the node won't be available from the block store, but it _is_ in memery.

api/bucketsd/client/client_test.go
```diff
@@ -712,6 +712,111 @@ func setPath(t *testing.T, private bool) {
...
+	require.NoError(t, err)
+	err = q.AddFile("a/b/c/file1.jpg", "testdata/file1.jpg")
+	require.NoError(t, err)
+	err = q.AddFile("a/b/c/file2.jpg", "testdata/file2.jpg")
+	require.NoError(t, err)
+	for q.Next() {
+		require.NoError(t, q.Err())
+	}
+	q.Close()
+
```
andrewxhill, line 743: worth a look at the mv expectations. matched cmd mv as closely as possible.

api/bucketsd/client/client_test.go
```diff
@@ -712,6 +712,111 @@ func setPath(t *testing.T, private bool) {
...
+	assert.Len(t, li.Item.Items, 1)
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "")
+	require.NoError(t, err)
+	assert.True(t, li.Item.IsDir)
+	assert.Len(t, li.Item.Items, 4)
+
+	// move non existant should fail
+	_, err = client.MovePath(ctx, buck.Root.Key, "x", "a")
+	if private {
```
andrewxhill, line 813: unfortunate difference. i could change the errors to match in the first part of the string

api/bucketsd/client/client_test.go
```diff
@@ -712,6 +712,111 @@ func setPath(t *testing.T, private bool) {
...
+	require.NoError(t, err)
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "")
+	require.NoError(t, err)
+	assert.True(t, li.Item.IsDir)
+	assert.Len(t, li.Item.Items, 4)
+
+	// move root should fail
+	_, err = client.MovePath(ctx, buck.Root.Key, "", "a")
+	require.Error(t, err, "source is root directory")
```
andrewxhill, line 799: not a hard requirement, i had it pretty close. but also felt like a reasonable restriction on the first pass.

api/bucketsd/client/client_test.go
sanderpick: Ah yeah. We could catch the IPFS "no link named" error and throw "could not resolve path" instead. Maybe worth a ticket in general, "Make public and private bucket errors consistent" or something.

api/bucketsd/client/client_test.go
```diff
@@ -741,8 +846,9 @@ func remove(t *testing.T, ctx context.Context, client *c.Client, private bool) {
 
 	err = client.Remove(ctx, buck.Root.Key)
 	require.NoError(t, err)
+
 	_, err = client.ListPath(ctx, buck.Root.Key, "again/file2.jpg")
-	require.Error(t, err)
+	require.NoError(t, err)
```
sanderpick, line 851: No error here anymore?

api/bucketsd/service.go
```diff
@@ -961,7 +1007,64 @@ func (s *Service) createLinks(
...
+	nodes := make([]ipld.Node, len(nmap))
+	i := 0
+	for _, tn := range nmap {
+		nodes[i] = tn.node
+		i++
+	}
+	pn := nmap[rootNode.Cid()].node
+	ctx, dirPath, err := s.insertNodeAtPath(ctx, pn, path.Join(path.New(buck.Path), toPath), buck.GetLinkEncryptionKey())
+	if err != nil {
+		return nil, fmt.Errorf("updating pinned root: %v", err)
```
sanderpick, line 1046: Error message copy pasta?

api/bucketsd/service.go
```diff
@@ -961,7 +1007,64 @@ func (s *Service) createLinks(
...
+	if err != nil {
+		return nil, err
+	}
+
+	// If updating root, add seedfile back to node.
+	if toPath == "" {
+		sn, err := makeSeed(fileKey)
+		ctx, dirPath, err = s.insertNodeAtPath(ctx, sn, path.Join(dirPath, buckets.SeedName), buck.GetLinkEncryptionKey())
+		if err != nil {
+			return nil, fmt.Errorf("updating pinned root: %v", err)
```
sanderpick, line 1059: Error message copy pasta?

api/bucketsd/service.go
```diff
@@ -961,7 +1007,64 @@ func (s *Service) createLinks(
...
+
+	_, err = s.addAndPinNodes(ctx, nodes)
+	if err != nil {
+		return nil, err
+	}
+
+	// If updating root, add seedfile back to node.
+	if toPath == "" {
+		sn, err := makeSeed(fileKey)
+		ctx, dirPath, err = s.insertNodeAtPath(ctx, sn, path.Join(dirPath, buckets.SeedName), buck.GetLinkEncryptionKey())
```
sanderpick, line 1057: Do we need to pin it as well (`addAndPinNodes`)?

api/bucketsd/service.go
```diff
@@ -970,16 +1073,184 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+		}
+		if toItem.IsDir {
+			// from => to becomes cleanDst:
+			// "c" => "b" becomes "b/c"
+			// "a.jpg" => "b" becomes "b/a.jpg"
+			toPth = gopath.Join(toPth, fromItem.Name)
+		}
+	}
+
+	// IPNS node
```
sanderpick, line 1134: Typo IPNS?

api/bucketsd/service.go
```diff
@@ -970,16 +1073,184 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+			}
+		}
+		return &pb.MovePathResponse{}, nil
+	}
+
+	// if a/ => b/ remove a
+	_, err = s.removePath(ctx, dbID, dbToken, &pb.RemovePathRequest{
+		Key:  req.Key,
+		Path: fromPth,
+		// Root: root.Root.Key,
```
sanderpick, line 1217: nit

api/bucketsd/service.go
```diff
@@ -970,16 +1073,184 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	buck.UpdatedAt = time.Now().UnixNano()
+	buck.SetMetadataAtPath(toPth, tdb.Metadata{
+		UpdatedAt: buck.UpdatedAt,
+	})
+	buck.UnsetMetadataWithPrefix(fromPth)
+
+	if err = s.Buckets.Verify(ctx, dbID, buck, tdb.WithToken(dbToken)); err != nil {
+		return nil, fmt.Errorf("verifying bucket update: %v", err)
+	}
+	if err = s.Buckets.Save(ctx, dbID, buck, tdb.WithToken(dbToken)); err != nil {
```
sanderpick, line 1165: This should be done last after all the pins have successfully been updated.

api/bucketsd/service.go
```diff
@@ -970,16 +1073,184 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+	}
+
+	buck.Path = dirPath.String()
+	buck.UpdatedAt = time.Now().UnixNano()
+	buck.SetMetadataAtPath(toPth, tdb.Metadata{
+		UpdatedAt: buck.UpdatedAt,
+	})
+	buck.UnsetMetadataWithPrefix(fromPth)
+
+	if err = s.Buckets.Verify(ctx, dbID, buck, tdb.WithToken(dbToken)); err != nil {
```
sanderpick, line 1162: I think this should be done as soon as you have `toPath` and `fromPth` and before any changes to the UnixFS dir are made. If the token is not authorized to make a change at `toPth` or `fromPth`, this should kick them out before changes are made.

api/bucketsd/service.go
```diff
@@ -970,16 +1073,184 @@ func (s *Service) SetPath(ctx context.Context, req *pb.SetPathRequest) (res *pb.
...
+		}
+	} else {
+		buckPath := path.New(buck.Path)
+		_, dirPath, err = s.setPathFromExistingPath(ctx, buck, buckPath, fromPth, toPth, buckNode.Cid(), nil, nil)
+		if err != nil {
+			return nil, fmt.Errorf("error copying path: %v", err)
+		}
+	}
+
+	buck.Path = dirPath.String()
```
sanderpick, line 1155: This can be done right before `s.Buckets.Save`

sanderpick (commented):
Looking great! I left some nits and a couple important comments about bucket verify vs. save.

api/bucketsd/client/client_test.go
```diff
@@ -712,6 +712,111 @@ func setPath(t *testing.T, private bool) {
...
+	// move and rename file => a/b/c/file2.jpg + a/b/file3.jpg
+	_, err = client.MovePath(ctx, buck.Root.Key, "a/b/c/file1.jpg", "a/b/file3.jpg")
+	require.NoError(t, err)
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "a/b/file3.jpg")
+	require.NoError(t, err)
+	assert.False(t, li.Item.IsDir)
+	assert.Equal(t, li.Item.Name, "file3.jpg")
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "root.jpg")
```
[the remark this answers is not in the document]
andrewxhill: right, should have moved it down a section

api/bucketsd/client/client_test.go
```diff
@@ -712,6 +712,111 @@ func setPath(t *testing.T, private bool) {
...
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "a")
+	require.NoError(t, err)
+	assert.True(t, li.Item.IsDir)
+	assert.Len(t, li.Item.Items, 1)
+
+	li, err = client.ListPath(ctx, buck.Root.Key, "")
+	require.NoError(t, err)
+	assert.True(t, li.Item.IsDir)
+	assert.Len(t, li.Item.Items, 4)
```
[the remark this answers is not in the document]
andrewxhill: c was moved to root up on 786

api/bucketsd/client/client_test.go
andrewxhill: k, will do. maybe we could implement a more detailed error type starting in here, that could return error types etc. it's a bit tricky to parse the ":" delimited stuff nicely. will move to a ticket though

api/bucketsd/client/client_test.go
andrewxhill: weird. cat must have done it

api/bucketsd/service.go
andrewxhill: will bump the one at 1049 below and append nodes if this section happens.

api/bucketsd/service.go
andrewxhill: ah, got it!

api/bucketsd/service.go
andrewxhill: right... i do it here because all the remaining methods are still using the off the shelf/existing functions (e.g. removePath, ListPath) that will pull the buck object again. i'll swap those out now to use the already open buck

api/bucketsd/service.go
```diff
@@ -2337,17 +2694,40 @@ func (s *Service) RemovePath(ctx context.Context, req *pb.RemovePathRequest) (re
...
 	}
+
 	if req.Root != "" && req.Root != buck.Path {
 		return nil, status.Error(codes.FailedPrecondition, buckets.ErrNonFastForward.Error())
 	}
 
+	dirPath, err := s.removePath(ctx, dbID, dbToken, buck, filePath)
+	if err != nil {
+		return nil, err
+	}
```
sanderpick, line 2708:
```
buck.UpdatedAt = time.Now().UnixNano()
buck.UnsetMetadataWithPrefix(filePath)
```
^ needs to happen before Verify is called. Otherwise, we're not verifying anything.

api/bucketsd/service.go
```diff
@@ -961,6 +1007,223 @@ func (s *Service) createLinks(
...
+			return nil, fmt.Errorf("removing path: %v", err)
+		}
+		buck.Path = dirPath.String()
+	}
+
+	buck.UpdatedAt = time.Now().UnixNano()
+	buck.SetMetadataAtPath(toPth, tdb.Metadata{
+		UpdatedAt: buck.UpdatedAt,
+	})
+	buck.UnsetMetadataWithPrefix(fromPth)
```
sanderpick, line 1210: This needs to be done before calling Verify, otherwise it won't have metadata change to verify.

sanderpick (commented):
@andrewxhill looks great but there's still an issue with the use of `Verify` (see comments).

api/bucketsd/service.go
andrewxhill: ahhh, thanks. i misunderstood the role of that. i think it doesn't make sense in the `removePath` but should be called first before any method that calls that.

andrewxhill: done

api/bucketsd/client/client_test.go
```diff
@@ -1052,13 +1232,13 @@ func TestClose(t *testing.T) {
...
+	bconf := apitest.DefaultBillingConfig(t)
+	apitest.MakeBillingWithConfig(t, bconf)
 
 	conf := apitest.DefaultTextileConfig(t)
-	//billingApi, err := tutil.TCPAddrFromMultiAddr(bconf.ListenAddr)
-	//require.NoError(t, err)
-	//conf.AddrBillingAPI = billingApi
+	billingApi, err := tutil.TCPAddrFromMultiAddr(bconf.ListenAddr)
+	require.NoError(t, err)
+	conf.AddrBillingAPI = billingApi
```
sanderpick, line 1241: Re-enabled billing in the bucket tests. I had disabled this on some other PR and forgot to re-enabled.

core/usage_interceptor.go
```diff
@@ -164,6 +164,7 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
 		"/api.bucketsd.pb.APIService/PushPath",
 		"/api.bucketsd.pb.APIService/PushPaths",
 		"/api.bucketsd.pb.APIService/SetPath",
+		"/api.bucketsd.pb.APIService/MovePath",
```
sanderpick, line 167: Needed for usage counting.

api/bucketsd/service.go
```diff
@@ -961,6 +971,247 @@ func (s *Service) createLinks(
...
+	// bucket node
+	buckNode, err := s.getNodeAtPath(ctx, pth, buck.GetLinkEncryptionKey())
+	if err != nil {
+		return nil, fmt.Errorf("getting node: %v", err)
+	}
+
+	var dirPath path.Resolved
+	// Private buckets need to manage keys+encrypted nodes
+	if buck.IsPrivate() {
+		ctx, dirPath, err = s.copyNode(ctx, buck, buckNode, fromPth, toPth)
```
sanderpick, line 1061: Anything that touches the nodes should return a modified context that reflects the storage change.

api/bucketsd/service.go
```diff
@@ -961,6 +971,247 @@ func (s *Service) createLinks(
...
+	}
+
+	if err = s.saveAndPublish(ctx, dbID, dbToken, buck); err != nil {
+		return nil, err
+	}
+
+	return &pb.MovePathResponse{}, nil
+}
+
+func (s *Service) copyNode(
```
sanderpick, line 1134: Moved below where it's used... helps with organization a little bit in this giant file.

api/bucketsd/service.go
```diff
@@ -961,6 +971,247 @@ func (s *Service) createLinks(
...
+	var dirPath path.Resolved
+	// Private buckets need to manage keys+encrypted nodes
+	if buck.IsPrivate() {
+		ctx, dirPath, err = s.copyNode(ctx, buck, buckNode, fromPth, toPth)
+		if err != nil {
+			return nil, fmt.Errorf("copying node: %v", err)
+		}
+	} else {
+		buckPath := path.New(buck.Path)
+		ctx, dirPath, err = s.setPathFromExistingCid(
```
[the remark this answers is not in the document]
andrewxhill: ah yes, i was doing that for private nodes but ended up going with copyPath instead. looks good.

sanderpick (approved):

PR #512, opened 2021-03-03
Sendfil Service

asutula:
This is going to be merged into https://github.com/textileio/textile/pull/480 and integrated into the `filrewardsd` service code, but splitting this out into a new PR just for clarity. 

Ok code all commented. 

This PR includes:

- Service for sending fil, listing transactions, as well as a summary api for reporting.
- Persists all txn data in mongo
- Utilizes lotus `StateWaitMsg` and a ticker to make sure all transaction records are updated with the latest data available from lotus.
- Supports a `wait` param when sending fil and getting a txn record so the caller can be notified when the message is active on chain.

Lots of detail left in the code comments.


api/sendfild/main.go
```diff
@@ -0,0 +1,202 @@
...
+		"MongoDB database name")
+
+	rootCmd.PersistentFlags().Duration(
+		"messageWaitTimeout",
+		config.Flags["messageWaitTimeout"].DefValue.(time.Duration),
+		"Timeout for listening for messages to become active on chain")
+
+	rootCmd.PersistentFlags().Int64(
+		"messageConfidence",
+		config.Flags["messageConfidence"].DefValue.(int64),
```
textileben, line 128: Getting a panic here about this flag.
```
$ docker run --rm textile/sendfil:staging
panic: interface conversion: interface {} is uint64, not int64

goroutine 1 [running]:
main.init.0()
        /textile/api/sendfild/main.go:128 +0xc0c
```

Think it's referring to these not matching? https://github.com/textileio/textile/pull/512/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod&file-filters%5B%5D=.proto&file-filters%5B%5D=.sum#diff-af38636851147b7aa79fdbf910e023fecb38fd9fe6826cff5e2c21111e26da27R67

asutula: Fixed

textileben (approved):
The Docker sections look good to me.

You might want to have someone else review the Filecoin portion as I'm not familiar with how the network messaging works.

api/sendfild/client/client.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/textileio/textile/v2/api/sendfild/pb"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type Client struct {
+	c    pb.SendFilServiceClient
+	conn *grpc.ClientConn
+}
```
asutula, line 16: Nothing too interesting in here, just the client code.

api/sendfild/main.go
```diff
@@ -0,0 +1,210 @@
...
+				DefValue: "", // no log file
+			},
+			"listenAddr": {
+				Key:      "listen_addr",
+				DefValue: "127.0.0.1:5000",
+			},
+			"lotusAddr": {
+				Key:      "lotus_addr",
+				DefValue: "127.0.0.1:7777",
+			},
```
asutula, line 44: Notice these are now just plain `host:port`, no multiaddr. This depends on an [open Powergate PR](https://github.com/textileio/powergate/pull/791).

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  int64 max_nano_fil_sent = 10;
+  int64 min_nano_fil_sent = 11;
+}
+
+service SendFilService {
+  rpc SendFil(SendFilRequest) returns (SendFilResponse) {}
+  rpc Txn(TxnRequest) returns (TxnResponse) {}
+  rpc ListTxns(ListTxnsRequest) returns (ListTxnsResponse) {}
+  rpc Summary(SummaryRequest) returns (SummaryResponse) {}
+}
```
asutula, line 103: Here's the API. Pretty straight forward, lets you send fil, query for a specific transaction, list transactions, and get a summary. The summary is mostly intended for monitoring so we can keep an eye on things.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+
+message ListTxnsResponse {
+  repeated Txn txns = 1;
+  bool more = 2;
+  int64 more_token = 3;
+}
+
+message SummaryRequest {
+  google.protobuf.Timestamp after = 1;
+  google.protobuf.Timestamp before = 2;
```
asutula, line 81: You can generate a summary for a range of time or if you don't provide any constraints, all time.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  int64 count_pending = 2;
+  int64 count_active = 3;
+  int64 count_failed = 4;
+  int64 count_waiting = 5;
+  int64 count_from_addrs = 6;
+  int64 count_to_addrs = 7;
+  int64 total_nano_fil_sent = 8;
+  double avg_nano_fil_sent = 9;
+  int64 max_nano_fil_sent = 10;
+  int64 min_nano_fil_sent = 11;
```
asutula, line 95: We can discuss how these should all be interpreted, but the idea is to have an idea of what "normal" looks like and keep an eye out for any deviations from that. A monitoring system can generate a summary, say every 30 min for the last hour (or more frequently), and we'll get some rolling data which could be nice.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  bool wait = 4;
+}
+
+message SendFilResponse {
+  Txn txn = 1;
+}
+
+message TxnRequest {
+  string message_cid = 1;
+  bool wait = 2;
```
asutula, line 46: Both `SendFil` and `Txn` take a `wait` parameter. If `true`, the response won't come until the transaction is active on chain or times out.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
+syntax = "proto3";
+package api.sendfild.pb;
+option go_package = "github.com/textileio/textile/v2/api/sendfild/pb";
+
+import "google/protobuf/timestamp.proto";
+
+enum MessageState {
+  MESSAGE_STATE_UNSPECIFIED = 0;
+  MESSAGE_STATE_PENDING = 1;
```
asutula, line 9: The message hasn't been detected as active on chain.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  MESSAGE_STATE_ACTIVE = 2;
```
asutula, line 10: Self explanatory.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  MESSAGE_STATE_FAILED = 3;
```
asutula, line 11: This is a final state that can't be recovered from. The message is permanently marked as failed.

asutula: The `Txn.failure_msg` will explain why.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+	bool waiting = 7;
+	string failure_msg = 8;
+	google.protobuf.Timestamp created_at = 9;
+	google.protobuf.Timestamp updated_at = 10;
+}
+
+message SendFilRequest {
+  string from = 1;
+  string to = 2;
+  int64 amount_nano_fil = 3;
```
asutula, line 36: Note the use of nano fil everywhere in this api.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+)
+
+var _ pb.SendFilServiceServer = (*Service)(nil)
+
+var log = logging.Logger("sendfil")
+
+type msgCid struct {
+	Cid       string    `bson:"cid"`
+	CreatedAt time.Time `bson:"created_at"`
+}
```
asutula, line 42: As a future proof thing, we support tracking multiple cids for a single message. This feature doesn't exist yet, but we could do something like: If a message is stuck in the mpool because the max gas fees are too low (or any other reason I'm not yet aware of), we could write code that detects that and automatically replaces the message with one that uses better max gas fee limits, etc. This new message would have an updated cid, and that cid would be added to the list of cids we track on a `txn` record. We then have the ability to look up a `txn` for any cid that has been associated with it over time.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+type txn struct {
+	ID            primitive.ObjectID `bson:"_id"`
+	From          string             `bson:"from"`
+	To            string             `bson:"to"`
+	AmountNanoFil int64              `bson:"amount_nano_fil"`
+	MessageCids   []msgCid           `bson:"message_cids"`
```
asutula, line 49: Here is that list of cids we track for a `txn`.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	MessageState  pb.MessageState    `bson:"message_state"`
+	Waiting       bool               `bson:"waiting"`
```
asutula, line 51: The state of "waiting" is mostly an internal thing, but I do expose it for monitoring purposes because it will be a signal about how healthy the service is... If too many txns are in a waiting state, that is a bad sign.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		return msgCid{}, fmt.Errorf("no message cids found")
+	}
+	return t.MessageCids[len(t.MessageCids)-1], nil
+}
+
+type Service struct {
+	clientBuilder lotus.ClientBuilder
+	col           *mongo.Collection
+	server        *grpc.Server
+	waiting       map[primitive.ObjectID]chan waitResult
```
asutula, line 68: We keep an in memory map of txn mongo ids to channels that complete once lotus tells us the message is on chain (or the message fails or we timeout the waiting)

asutula: This lets us share any message waiting between multiple calls and doesn't create additional calls to lotus to do that.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	config        Config
+	ticker        *time.Ticker
```
asutula, line 70: We use this ticker to periodically query for txns that need to be waited on still (because of a precious timeout or recoverable error)

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	waitingLck    sync.Mutex
```
asutula, line 71: We synchronize access to the `waiting` map using this lock.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		{
+			Keys: bson.D{primitive.E{Key: "amount_nano_fil", Value: 1}},
+		},
+		// MongoDB automatically creates a multikey index if any indexed field is an array;
+		// you do not need to explicitly specify the multikey type.
+		// https://docs.mongodb.com/manual/core/index-multikey/
+		{
+			Keys:    bson.D{primitive.E{Key: "message_cids.cid", Value: 1}},
+			Options: options.Index().SetUnique(true),
+		},
```
asutula, line 116: Cool mongo feature where you can index values nested in a list, in this case, uniquely.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		if err := s.server.Serve(config.Listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
+			log.Errorf("serve error: %v", err)
+		}
+	}()
+
+	if err := s.waitAllQualifying(ctx, true); err != nil {
+		return nil, fmt.Errorf("calling waitAllQualifying: %v", err)
+	}
+
+	s.bindTicker(ctx)
```
asutula, line 153: We bootstrap our message waiting on init and then bind the ticker so we retry message waiting periodically.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+}
+
+func (s *Service) waitAllQualifying(ctx context.Context, isInitialRun bool) error {
+	filter := bson.M{"$and": bson.A{
+		bson.M{"message_state": bson.M{"$ne": pb.MessageState_MESSAGE_STATE_ACTIVE}},
+		bson.M{"message_state": bson.M{"$ne": pb.MessageState_MESSAGE_STATE_FAILED}},
+	}}
+	if !isInitialRun {
+		filter["waiting"] = false
+	}
```
asutula, line 165: To detect txns that need to be waited on, we always ask for all messages that aren't in a final state. If it is the "init" call of this func, that is all we need to do, but if this is a call driven by the ticker, we also want to exclude any txns that are already in a waiting state.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+func (s *Service) bindTicker(ctx context.Context) {
+	go func() {
+		for {
+			<-s.ticker.C
+			if err := s.waitAllQualifying(ctx, false); err != nil {
+				log.Errorf("waitAllQualifying from ticker: %v", err)
+			}
+		}
+	}()
```
asutula, line 191: Pretty simple here, whenever the ticker ticks, call `waitAllQualifying`, passing `false` since it's not the initial run.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing from address: %v", err)
+	}
+	t, err := address.NewFromString(req.To)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing to address: %v", err)
+	}
+	nanoAmount := (&big.Int{}).SetInt64(req.AmountNanoFil)
+	factor := (&big.Int{}).SetInt64(int64(math.Pow10(9)))
+	amount := (&big.Int{}).Mul(nanoAmount, factor)
```
asutula, line 205: Here we convert from nano fil to fil.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		AmountNanoFil: req.AmountNanoFil,
+		MessageCids:   []msgCid{{Cid: sm.Message.Cid().String(), CreatedAt: now}},
+		MessageState:  pb.MessageState_MESSAGE_STATE_PENDING,
+		CreatedAt:     now,
+		UpdatedAt:     now,
+	}
+
+	if _, err = s.col.InsertOne(ctx, &tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "inserting txn into collection: %v", err)
+	}
```
asutula, line 237: Save the txn record I mongo in its initial state.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	wait := s.wait(tx)
```
asutula, line 239: We always call `wait` when `SendFil` is called. `wait` returns a channel that may or may not be read from.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	if req.Wait {
+		res := <-wait
+		if res.err != nil {
+			return nil, status.Errorf(codes.Internal, "waiting for result: %v", res.err)
+		}
+		tx = res.txn
```
asutula, line 246: If requested, we block until we're done waiting and the message in active on chain.

asutula: We update the `txn` we'll return so we get the latest/updated version that came from the wait chan.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+}
+
+func (s *Service) Txn(ctx context.Context, req *pb.TxnRequest) (*pb.TxnResponse, error) {
+	res := s.col.FindOne(ctx, bson.M{"message_cids.cid": req.MessageCid})
+	if res.Err() == mongo.ErrNoDocuments {
+		return nil, status.Error(codes.NotFound, "no txn found for cid")
+	}
+	if res.Err() != nil {
+		return nil, status.Errorf(codes.Internal, "querying for cid txn: %v", res.Err())
+	}
```
asutula, line 264: This API finds a txn by cid.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	if err := res.Decode(&tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "decoding cid txn result: %v", res.Err())
+	}
+
+	if tx.MessageState != pb.MessageState_MESSAGE_STATE_ACTIVE && tx.MessageState != pb.MessageState_MESSAGE_STATE_FAILED && req.Wait {
+		res := <-s.wait(tx)
+		if res.err != nil {
+			return nil, status.Errorf(codes.Internal, "waiting for result: %v", res.err)
+		}
+		tx = res.txn
```
asutula, line 276: If requested, and the txn isn't already in a final state, we wait for the message to be active on chain (or time out or fail). Similar to `SendFil` in this way.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	pbTx, err := toPbTxn(tx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "converting to pb txn: %v", err)
+	}
+
+	return &pb.TxnResponse{Txn: pbTx}, nil
+}
+
+func (s *Service) ListTxns(ctx context.Context, req *pb.ListTxnsRequest) (*pb.ListTxnsResponse, error) {
```
asutula, line 287: Our "list" api. I went pretty deep with the query filter support, so all the below code is a bit hard to read, but it just maps request params to find options and filters. This is all tested pretty well.

asutula: Some combinations of query options are not valid, so this code just parses valid combinations and invalid combinations are no-op. The valid combinations are common sense kinda stuff, not idea feelback to the caller, but hopefully fine for an internal service like this.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		filter["created_at"] = bson.M{comp: *lastCreatedAt}
+		res := s.col.FindOne(ctx, filter)
+		if res.Err() != nil && res.Err() != mongo.ErrNoDocuments {
+			return nil, status.Errorf(codes.Internal, "checking for more data: %v", err)
+		}
+		if res.Err() != mongo.ErrNoDocuments {
+			more = true
+			startAt = lastCreatedAt
+		}
+	}
```
asutula, line 394: I had to do this in other code I've been writing lately, would love to find a better way. But this runs an additional query to see if there is more data available so we can tell the client.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			return nil, status.Errorf(codes.Internal, "converting txn to pb: %v", err)
+		}
+		pbTxns = append(pbTxns, pbTxn)
+	}
+	res := &pb.ListTxnsResponse{
+		Txns: pbTxns,
+		More: more,
+	}
+	if startAt != nil {
+		res.MoreToken = startAt.UnixNano()
```
asutula, line 408: Using unix nanos as the "more token", but not important or meaningful to the client other than it can be passed in the next request.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	}
+	type report struct {
+		All            []entityCount `bson:"all"`
+		ByMessageState []entityCount `bson:"by_message_state"`
+		Waiting        []entityCount `bson:"waiting"`
+		UniqueFrom     []entityCount `bson:"unique_from"`
+		UniqueTo       []entityCount `bson:"unique_to"`
+		SentFilStats   []stats       `bson:"sent_fil_stats"`
+	}
+
```
asutula, line 433: This internal type is populated using the mongo aggregation pipeline below.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	}
+	if req.After != nil {
+		createdAtMatch["$gt"] = req.After.AsTime()
+	}
+	match := bson.M{}
+	if len(createdAtMatch) > 0 {
+		match["created_at"] = createdAtMatch
+	}
+	cursor, err := s.col.Aggregate(ctx, bson.A{
+		bson.M{"$match": match},
```
asutula, line 446: First we filter to any provided date range.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		bson.M{"$facet": bson.M{
```
asutula, line 447: Facet lets you pass the incoming result set to any number of parallel sub pipelines, pretty cool.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			},
+			"sent_fil_stats": bson.A{
+				bson.M{"$group": bson.M{
+					"_id":   nil,
+					"total": bson.M{"$sum": "$amount_nano_fil"},
+					"avg":   bson.M{"$avg": "$amount_nano_fil"},
+					"max":   bson.M{"$max": "$amount_nano_fil"},
+					"min":   bson.M{"$min": "$amount_nano_fil"},
+				}},
+			},
```
asutula, line 472: Using aggregation to calculate various potentially interesting things.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	resp.CountToAddrs = int64(len(r.UniqueTo))
+
+	if len(r.SentFilStats) == 1 {
+		resp.TotalNanoFilSent = r.SentFilStats[0].Total
+		resp.AvgNanoFilSent = r.SentFilStats[0].Avg
+		resp.MaxNanoFilSent = r.SentFilStats[0].Max
+		resp.MinNanoFilSent = r.SentFilStats[0].Min
+	}
+
+	return resp, nil
```
asutula, line 519: We populate our response with the data that came from the aggregation and return it.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+}
+
+type waitResult struct {
+	txn txn
+	err error
+}
+
+func (s *Service) wait(tx txn) chan waitResult {
```
asutula, line 527: Ok here is where the magic is. 

I love that this func simply returns a chan of `txn` (well, wrapped in a `waitResult
` so we can also pass any error) because it makes the resulting chan easy to consume and the chan output is consistent with the type of data needed in the contexts from which `wait` is called. 

What I don't like is that this func is FULL of side effects. It manages the `waiting` map and lock to keep that in memory state up to date. It communicates with Lotus to wait for the message to become active. It updates the in memory `txn` that is sent on the result chan and updates the underlying mongo representation. This doesn't feel great, but it does keep all the internal state consistent and provide an very clean api, so I went with it 🤷🏼‍♂️

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	s.waitingLck.Lock()
+	defer s.waitingLck.Unlock()
+
+	waitCh, found := s.waiting[tx.ID]
+	if found {
+		return waitCh
+	}
```
asutula, line 534: If we're already waiting on this `txn` just return the existing chan.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	ch := make(chan waitResult)
+	s.waiting[tx.ID] = ch
```
asutula, line 537: If not, make a new chan and store it in our map for later.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		resp.MinNanoFilSent = r.SentFilStats[0].Min
+	}
+
+	return resp, nil
+}
+
+type waitResult struct {
+	txn txn
+	err error
+}
```
asutula, line 525: We send this type over the chan returned from `wait`. 

`err` signifies a recoverable error where the txn should be waited on again in the future. 

`txn.MessageState == pb.MessageState_MESSAGE_STATE_FAILED` signifies a final error and the txn should never be waited on again.

`txn.MessageState == pb.MessageState_MESSAGE_STATE_ACTIVE` means the txn message is successfully on chain.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	waitCh, found := s.waiting[tx.ID]
+	if found {
+		return waitCh
+	}
+
+	ch := make(chan waitResult)
+	s.waiting[tx.ID] = ch
+
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), s.config.MessageWaitTimeout)
```
asutula, line 540: We use the configured timeout. Default is 5 min.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		client, closeClient, err := s.clientBuilder(ctx)
+		defer func() {
+			closeClient()
+			close(ch)
+			cancel()
+		}()
+		if err != nil {
+			ch <- waitResult{err: fmt.Errorf("creating lotus client: %v", err)}
+			return
+		}
```
asutula, line 551: Anywhere we come across a temporary or recoverable error, we send that error on the result chan and return. In that case, the txn will be waited again in the future.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			tx.FailureMsg = fmt.Sprintf("decoding latest message cid for txn: %v", err)
+			tx.UpdatedAt = time.Now()
+			log.Warnf("failing txn with err: %s", tx.FailureMsg)
+			if err := s.updateTxn(ctx, tx); err != nil {
+				ch <- waitResult{err: err}
+				return
+			}
+			ch <- waitResult{txn: tx}
+			return
+		}
```
asutula, line 578: If for some strange reason we can't make sense of the cid associated with the txn, that is a fatal/final error condition.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+		tx.Waiting = true
+		if err := s.updateTxn(ctx, tx); err != nil {
+			ch <- waitResult{err: err}
+			return
+		}
```
asutula, line 584: Mark our txn as waiting and save that update to mongo.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+		res, err := client.StateWaitMsg(ctx, c, s.config.MessageConfidence)
+		tx.Waiting = false
```
asutula, line 587: Mark txn as not waiting, this gets saved to mongo below, either in the error case or in following analysis.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		if err != nil {
+			// If for some reason the lotus node doesn't know about the cid, consider that a final error.
+			if strings.Contains(err.Error(), "block not found") {
+				tx.MessageState = pb.MessageState_MESSAGE_STATE_FAILED
+				tx.FailureMsg = err.Error()
+				tx.UpdatedAt = time.Now()
+				log.Warnf("failing txn with err: %s", tx.FailureMsg)
+			}
```
asutula, line 595: Ugly, but if lotus doesn't know about our cid, that is a fatal/final error, so update `txn` to reflect that.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			if err := s.updateTxn(ctx, tx); err != nil {
+				ch <- waitResult{err: err}
+				return
+			}
```
asutula, line 599: Save to mongo.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			if tx.MessageState == pb.MessageState_MESSAGE_STATE_FAILED {
+				ch <- waitResult{txn: tx}
+			} else {
+				ch <- waitResult{err: fmt.Errorf("calling StateWaitMsg: %v", err)}
+			}
+			return
```
asutula, line 605: Depending on if the error was fatal or not, send the appropriate event on the chan.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		}
+
```
asutula, line 607: Ok we're back out into non-error territory. Everything below here assumes that we called `client.StateWaitMsg` successfully.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+				return
+			}
+			ch <- waitResult{txn: tx}
+			if res.Receipt.ExitCode.IsSendFailure() {
+				log.Errorf("received exit code send failure: %s", res.Receipt.ExitCode.String())
+			} else {
+				log.Infof("received exit code error: %s", res.Receipt.ExitCode.String())
+			}
+			return
+		}
```
asutula, line 624: If the receipt data we get from lotus indicates that the exit code is an error exit code, we consider this a final error. We do some extra logging to get a feel for if the exit code is a "send failure" (which would be strange and alarming) or some other more normal failing exit code.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+		tx.MessageState = pb.MessageState_MESSAGE_STATE_ACTIVE
+		tx.UpdatedAt = time.Now()
```
asutula, line 627: If we make it this far, that means the txn message is active on chain, woohoo!

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		for _, msgCid := range tx.MessageCids {
+			if res.Message.String() == msgCid.Cid {
+				isNewCid = false
+				break
+			}
+		}
+
+		if isNewCid {
+			tx.MessageCids = append(tx.MessageCids, msgCid{Cid: res.Message.String(), CreatedAt: time.Now()})
+		}
```
asutula, line 641: Don't know what to think of this. Could leave it as a safety net, but it should never be the case since there is no process replacing messages that are already in the mpool. And once that process does exist, it would be the thing triggering creating the new cid, and would therefore save the new cid to mongo immediately on the front end.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+		if err := s.updateTxn(ctx, tx); err != nil {
+			ch <- waitResult{err: err}
+			return
+		}
```
asutula, line 646: Final call to save `txn` after all the non-error-case updates.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+		s.waitingLck.Lock()
+		delete(s.waiting, tx.ID)
+		s.waitingLck.Unlock()
+
+		ch <- waitResult{txn: tx}
```
asutula, line 652: Remove the chan from the in memory map and send the final txn on the results chan.

api/sendfild/service/service_test.go
```diff
@@ -0,0 +1,646 @@
...
+	"github.com/textileio/powergate/v2/tests"
+	powutil "github.com/textileio/powergate/v2/util"
+	pb "github.com/textileio/textile/v2/api/sendfild/pb"
+	"github.com/textileio/textile/v2/util"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/grpc/test/bufconn"
+)
+
```
asutula, line 29: Tried to be very thorough for these tests. There are an unreasonable number of permutations of all the `ListTxns` options, so just tried to test the obvious ones there.

api/sendfild/service/service_test.go
```diff
@@ -0,0 +1,646 @@
...
+	cleanup := func() {}
+	if os.Getenv("SKIP_SERVICES") != "true" {
+		cleanup = test.StartMongoDB()
+	}
+	exitVal := m.Run()
+	cleanup()
+	os.Exit(exitVal)
+}
+
+func TestRestartWaiting(t *testing.T) {
```
asutula, line 52: This is a very important one. It tests that the waiting state is properly initialized upon service restart.

api/sendfild/service/service_test.go
```diff
@@ -0,0 +1,646 @@
...
+	requireSendFil(t, ctx, c, dAddr.String(), addr1.String(), oneFil, false)
+	requireSendFil(t, ctx, c, dAddr.String(), addr1.String(), oneFil, false)
+	txLast := requireSendFil(t, ctx, c, dAddr.String(), addr1.String(), oneFil, false)
+
+	pageResults := func(ascending bool) {
+		numPages := 0
+		more := true
+		var moreToken int64 = 0
+		for more {
+			req := &pb.ListTxnsRequest{CreatedAfter: txFirst.CreatedAt, CreatedBefore: txLast.CreatedAt, Limit: 3, Ascending: ascending}
```
asutula, line 453: This paging test makes sure that paging works properly when `CreatedAfter` and `CreatedBefore` constraints are also provided, both ascending and descending. This was tricky because paging also internally uses a `created_at` filter via the `MoreToken`... so it works!

api/sendfild/service/service.go
asutula: Important to note that `wait` is a "get or create" kind of API.

api/sendfild/service/service_test.go
```diff
@@ -0,0 +1,646 @@
...
+		last = &t
+	}
+}
+
+func randomCid() cid.Cid {
+	data := make([]byte, 20)
+	rand.Read(data)
+	hash, _ := mh.Sum(data, mh.SHA2_256, -1)
+	return cid.NewCidV1(cid.DagCBOR, hash)
+}
```
jsign, line 646: Nice tests!

jsign (approved):
LGTM, but I left some comments in the review for some consideration. But maybe the plan is merging and continue in other PR maybe?

Some extra question: Are there some plans to include some sanity checks regarding big send fil amounts or similar? As to avoid errors ins getting clients sending too much FIL maybe by error or something out of control?

BTW, I left comments in order as I was thinking things. So if any other does the review, can follow some line of thought, even if some extra comment is discovered later.

api/sendfild/pb/sendfil.proto
asutula: Yea exactly. We use a reasonable timeout while waiting for the message in the lotus api. If that timeout expires, the error returned here will say that is what happened. The client is free to make another request. The service will also retry any previously timed out waits.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+
+message ListTxnsRequest {
+  string from_filter = 1;
+  string to_filter = 2;
+  string involving_filter = 3;
+  int64 amount_nano_fil_lt_filter = 4;
+  int64 amount_nano_fil_gt_filter = 5;
+  int64 amount_nano_fil_lteq_filter = 6;
+  int64 amount_nano_fil_gteq_filter = 7;
+  int64 amount_nano_fil_eq_filter = 8;
```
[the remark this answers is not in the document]
asutula: There are certain invalid combinations which are ignored. Only valid combinations are used.

0 is not a valid or useful value for any of the `amount_nano_fil_...` filters, so we can use those filters only if they are set to something non `0`.

api/sendfild/pb/sendfil.proto
```diff
@@ -0,0 +1,103 @@
...
+  MESSAGE_STATE_UNSPECIFIED = 0;
+  MESSAGE_STATE_PENDING = 1;
+  MESSAGE_STATE_ACTIVE = 2;
+  MESSAGE_STATE_FAILED = 3;
+}
+
+enum WaitingFilter {
+  WAITING_FILTER_UNSPECIFIED = 0;
+  WAITING_FILTER_WAITING = 1;
+  WAITING_FILTER_NOT_WAITING = 2;
```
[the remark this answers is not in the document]
asutula: It means the txn is pending AND the service actively has a `StateWaitMsg` call to louts running. The concept of "waiting" is really an internal thing, but I thought it would be a useful metric for knowing the health of the service, so I exposed it in the API.

api/sendfild/client/client.go
```diff
@@ -0,0 +1,199 @@
...
+		c:    c,
+		conn: conn,
+	}, nil
+}
+
+type SendFilOption = func(*pb.SendFilRequest)
+
+func SendFilWait() SendFilOption {
+	return func(req *pb.SendFilRequest) {
+		req.Wait = true
```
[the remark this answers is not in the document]
asutula: Ah good point. I think it is normal and fine to want to wait here, and there is a chance that the underlying lotus call to `StateMsgWait` will timeout. This doesn't meant the transaction failed or timed out, just that our monitoring of it timed out. We need to make that clear to the caller. I think providing a proper error message in this case is the key.

api/sendfild/client/client.go
```diff
@@ -0,0 +1,199 @@
...
+
+type TxnOption = func(*pb.TxnRequest)
+
+func TxnWait() TxnOption {
+	return func(req *pb.TxnRequest) {
+		req.Wait = true
+	}
+}
+
+func (c *Client) Txn(ctx context.Context, messageCid string, opts ...TxnOption) (*pb.Txn, error) {
```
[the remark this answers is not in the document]
asutula: 👍🏼

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	MessageState  pb.MessageState    `bson:"message_state"`
+	Waiting       bool               `bson:"waiting"`
+	FailureMsg    string             `bson:"failure_msg"`
+	CreatedAt     time.Time          `bson:"created_at"`
+	UpdatedAt     time.Time          `bson:"updated_at"`
+}
+
+func (t txn) latestMsgCid() (msgCid, error) {
+	if len(t.MessageCids) == 0 {
+		return msgCid{}, fmt.Errorf("no message cids found")
```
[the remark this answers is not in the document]
asutula: It should never be possible. Would love not to have to return `error` here. What's the best way to deal with situations like this? I see the use of `panic` that many people would do, but I've never understood why you would want to crash your entire service intentionally, if you can just return an error or log error instead.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	CreatedAt time.Time `bson:"created_at"`
+}
+
+type txn struct {
+	ID            primitive.ObjectID `bson:"_id"`
+	From          string             `bson:"from"`
+	To            string             `bson:"to"`
+	AmountNanoFil int64              `bson:"amount_nano_fil"`
+	MessageCids   []msgCid           `bson:"message_cids"`
+	MessageState  pb.MessageState    `bson:"message_state"`
```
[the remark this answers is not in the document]
asutula: Yea, figured it would never change.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	s.bindTicker(ctx)
+
+	return s, nil
+}
+
+func (s *Service) waitAllQualifying(ctx context.Context, isInitialRun bool) error {
+	filter := bson.M{"$and": bson.A{
+		bson.M{"message_state": bson.M{"$ne": pb.MessageState_MESSAGE_STATE_ACTIVE}},
+		bson.M{"message_state": bson.M{"$ne": pb.MessageState_MESSAGE_STATE_FAILED}},
```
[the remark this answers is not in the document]
asutula: Good call, I'll change it to pending.

api/sendfild/service/service.go
asutula: Pending - we have not yet detected the message is active on chain
Waiting - this service has an active call to lotus `StateWaitMsg` running

api/sendfild/service/service.go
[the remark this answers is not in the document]
jsign: I think erroring is fine 👍🏼 . I was mostly thinking if you ever save a `txn` with an empty array or not at this point.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	err error
+}
+
+func (s *Service) wait(tx txn) chan waitResult {
+	s.waitingLck.Lock()
+	defer s.waitingLck.Unlock()
+
+	waitCh, found := s.waiting[tx.ID]
+	if found {
+		return waitCh
```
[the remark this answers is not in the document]
asutula: Oh interesting. I think my understanding of how channels work is wrong then. I wrongly assumed they are like a fan out kind of subscription. i.e. supporting multiple subscribers. Will definitely need to change this then.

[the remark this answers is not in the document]
asutula: Guess I need to build some sort of broadcaster/listener infrastructure then?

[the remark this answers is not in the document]
jsign: Yeah, for that kind of functionality you need that.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	log.Infof("found %v txns to initiate waiting on", len(txns))
+	for _, txn := range txns {
+		s.wait(txn)
+	}
+	return nil
+}
+
+func (s *Service) bindTicker(ctx context.Context) {
+	go func() {
+		for {
```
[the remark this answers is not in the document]
asutula: Like this?

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	if res.Err() != nil {
+		return nil, status.Errorf(codes.Internal, "querying for cid txn: %v", res.Err())
+	}
+
+	var tx txn
+	if err := res.Decode(&tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "decoding cid txn result: %v", res.Err())
+	}
+
+	if tx.MessageState != pb.MessageState_MESSAGE_STATE_ACTIVE && tx.MessageState != pb.MessageState_MESSAGE_STATE_FAILED && req.Wait {
```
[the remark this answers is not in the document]
asutula: Ok, using `== PENDING` now. 

I think the case of the txn being pending and `req.Wait` is `false`, just means the client is asking for the current status of the txn (to the best of the service's knowledge), so we just return it. Maybe I'm not understanding your question.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		res := <-s.wait(tx)
+		if res.err != nil {
```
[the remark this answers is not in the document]
asutula: I'm going to defer the chan related concerns until after I refactor to broadcast updates to all listeners. That will change all of these types of issues.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+			close(ch)
+			cancel()
+		}()
+		if err != nil {
+			ch <- waitResult{err: fmt.Errorf("creating lotus client: %v", err)}
+			return
+		}
+
+		lastestMsgCid, err := tx.latestMsgCid()
+		if err != nil {
```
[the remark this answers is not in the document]
asutula: In practice, a txn should always have at least one cid. I will log error if it doesn't so we are aware of some strange situation that shouldn't be happening.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+				log.Warnf("failing txn with err: %s", tx.FailureMsg)
+			}
+			if err := s.updateTxn(ctx, tx); err != nil {
+				ch <- waitResult{err: err}
+				return
+			}
+			if tx.MessageState == pb.MessageState_MESSAGE_STATE_FAILED {
+				ch <- waitResult{txn: tx}
+			} else {
+				ch <- waitResult{err: fmt.Errorf("calling StateWaitMsg: %v", err)}
```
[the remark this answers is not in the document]
asutula: Correct

api/sendfild/service/service.go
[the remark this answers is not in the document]
jsign: Mmm, what is "this"? :P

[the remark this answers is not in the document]
asutula: The `select` clause below.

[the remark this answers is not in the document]
jsign: Which line? In this file?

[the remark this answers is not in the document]
asutula: This file, line 191 just below this comment thread. I think maybe you're viewing the diff at some outdated commit.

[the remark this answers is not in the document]
jsign: Dunno, I'm still confused. I see this and seems to be the latest code:
image
Line 191 doesn't have a `select`. ❓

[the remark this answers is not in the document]
jsign: Had to do a hard-refresh 🤷🏼‍♂️ 
Now I see it. Yep, that's what I was referring to 👍🏼

api/sendfild/service/service.go
asutula: I agree. Will change in upcoming PR.

asutula:
Ok @jsign I fixed the basic things based on your feedback. I'm going to merge this into `asutula/fil-rewards-bookkeeping` and follow up with another PR that addresses the more complex issues of:

- removing `more` from list results
- dealing with no nanosecond precision in mongo for `moreToken`
- correctly using channels to notify all interested callers of updates to messages

api/sendfild/service/service.go
asutula: This is resolved in later comments about side effects etc. This will change in the next PR when I implement waiting and notifying properly.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	}
+
+	if _, err = s.col.InsertOne(ctx, &tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "inserting txn into collection: %v", err)
+	}
+
+	wait := s.wait(tx)
+
+	if req.Wait {
+		res := <-wait
```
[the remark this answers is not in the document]
asutula: Yes, need to handle that case for sure. Will add in the next PR that updates wait/notify logic and I'll be sure to use a good error message that explains its a timeout, not a failure.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "converting to pb txn: %v", err)
+	}
+
+	return &pb.TxnResponse{Txn: pbTx}, nil
+}
+
+func (s *Service) ListTxns(ctx context.Context, req *pb.ListTxnsRequest) (*pb.ListTxnsResponse, error) {
+	findOpts := options.Find()
+	if req.Limit > 0 {
```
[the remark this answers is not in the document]
asutula: Was thinking the trusted client thing made this not a big problem, but will consider adding this limit in the next PR.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	}
+
+	// Apply paging info
+	comp := "$lt"
+	if req.MoreToken != 0 {
+		if req.Ascending {
+			comp = "$gt"
+		}
+		t := time.Unix(0, req.MoreToken)
+		ands = append(ands, bson.M{"created_at": bson.M{comp: &t}})
```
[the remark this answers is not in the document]
asutula: This all makes sense. I'll think a bit more about how to move forward here and implement a fix in the upcoming PR.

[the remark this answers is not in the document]
asutula: If you think I'm leaning too much into the idea of not using `skip`, let me know. My motivation was to support ascending and descending sorting while never having result set drift (missing or repeating records when dealing with financial transactions seems unacceptable to me). We can remove descending sort option and use `skip`. Or we can use a composite key like you suggested (time, cid).

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+		filter["$and"] = ands
+	}
+
+	cursor, err := s.col.Find(ctx, filter, findOpts)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "querying txns: %v", err)
+	}
+	defer cursor.Close(ctx)
+	var txns []txn
+	err = cursor.All(ctx, &txns)
```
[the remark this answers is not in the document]
asutula: 👍🏼 address the related comment above.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "decoding txns query results: %v", err)
+	}
+
+	more := false
+	var startAt *time.Time
+	if len(txns) > 0 {
+		lastCreatedAt := &txns[len(txns)-1].CreatedAt
```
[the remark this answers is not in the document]
asutula: I'll try to think of a better name. My thinking was that it is the "last" item in the slice, no matter which order the slice is sorted in.

api/sendfild/service/service.go
asutula: This will change in the next PR when I implement waiting and notifying properly. I think the actual waiting will be done in one place as be more of an internal implementation detail. API calls that want to wait will call a subscription type of API and won't have to deal with this currently opaque and poorly named func.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), s.config.MessageWaitTimeout)
+
+		client, closeClient, err := s.clientBuilder(ctx)
+		defer func() {
+			closeClient()
+			close(ch)
+			cancel()
+		}()
```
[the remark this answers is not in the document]
asutula: Flagged this as something to handle properly when I implement wait/notify correctly in the next PR.

api/sendfild/service/service.go
```diff
@@ -0,0 +1,713 @@
...
+	ch := make(chan waitResult)
+	s.waiting[tx.ID] = ch
+
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), s.config.MessageWaitTimeout)
+
+		client, closeClient, err := s.clientBuilder(ctx)
+		defer func() {
+			closeClient()
+			close(ch)
```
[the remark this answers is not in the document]
asutula: Yes, will fix with upcoming wait/notify changes.

api/sendfild/service/service.go
asutula: I think it will be fine. It will depend on my final wait/notify implementation in the next PR, so I'll be sure the multiple callers things is covered. It will be cleaner because multiple callers will be adding themselves as subscribers for the result of a single underlying call to lotus. That is a more typical and understandable API.

api/sendfild/service/service.go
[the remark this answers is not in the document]
jsign: Whatever you think is best here. My point was mostly about this _millisecond_ precision at the `Date` level in MongoDB.
The decision around Skip-Offset might be more related to potentially include other sorting criteria in the future.

PR #513, opened 2021-03-03
feat(cli): adds flag based config setting

andrewxhill:
so this just opens it to three ways to set the config.

1. `hub buck archive set-default-config ../storage-config.json`
2. `cat ../storage-config.json | hub buck archive set-default-config --`
3. `hub buck archive set-default-config --rep-factor=3 --fast-retrieval --verified-deal --trusted-miners=f08240,f023467,f09848`


cmd/buck/cli/cli.go
```diff
@@ -80,6 +80,17 @@ func Init(baseCmd *cobra.Command) {
 	archiveCmd.Flags().StringP("file", "f", "", "Optional path to a file containing archive config json that will override the default")
 	archiveCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
 
+	setDefaultArchiveConfigCmd.Flags().IntP("rep-factor", "r", 1, "Target number of active deals")
+	setDefaultArchiveConfigCmd.Flags().Int64P("deal-min-duration", "d", 518400, "Minimum duration for the deal")
+	setDefaultArchiveConfigCmd.Flags().StringSliceP("excluded-miners", "x", nil, "Miner addresses that should not be used")
+	setDefaultArchiveConfigCmd.Flags().StringSliceP("trusted-miners", "t", nil, "Miner addresses that must be used")
+	setDefaultArchiveConfigCmd.Flags().StringSliceP("country-codes", "c", nil, "Select miners with specific countries")
+	setDefaultArchiveConfigCmd.Flags().Uint64P("max-price", "p", 0, "Maximum price that will be spent per RepFactor in AttoFIL/GiB per Epoch")
```
andrewxhill, line 88: what to set this now?

jsign: I'll move the semantics to "update" rather than set.
I'll continue to work it on this branch.

api/bucketsd/service.go
```diff
@@ -92,6 +92,7 @@ var (
 		FastRetrieval:   true,
 		DealStartOffset: 72 * 60 * 60 / powUtil.EpochDurationSeconds, // 72hs
 		MaxPrice:        100_000_000_000,
+		VerifiedDeal:    false,
```
jsign, line 95: In these cases, I prefer to set explicitly false (even if that's the case as it's bool default), to be completely clear that we didn't miss to configure this.

api/bucketsd/service.go
```diff
@@ -3354,6 +3355,7 @@ func fromPbArchiveConfig(pbConfig *pb.ArchiveConfig) *mdb.ArchiveConfig {
 			MaxPrice:        pbConfig.MaxPrice,
```
jsign, line 3355: Some extra cascading to use the new `VerifiedDeal` field you added to:
- The struct that we keep in Hub saving the storage-config (mdb.ArchiveConfig)
- Map the `VerifiedDeal` to the storage-config being prepared to send to powergate when pushing the Cid (line 3387).

cmd/buck/cli/archive.go
```diff
@@ -35,21 +35,30 @@ var defaultArchiveConfigCmd = &cobra.Command{
...
-	Short:   "Set the default archive storage configuration for the specified Bucket.",
-	Long:    `Set the default archive storage configuration for the specified Bucket from a file, stdin, or options.`,
+	Use:   "set-default-config [(optional)file]",
+	Short: "Set the default archive storage configuration for the specified Bucket.",
+	Long: `Set the default archive storage configuration for the specified Bucket from a file, stdin, or flags.
+
+If flags are specified, this command updates the current default storage-config with the *explicitely set* flags. 
+Flags that aren't explicitely set won't set the default value, and thus keep the original value in the storage-config.
+
+If a file or stdin is used, the storage-config will be completely overriden by the provided one.`,
```
jsign, line 45: Explaining a bit more how this works.

cmd/buck/cli/archive.go
```diff
@@ -35,21 +35,30 @@ var defaultArchiveConfigCmd = &cobra.Command{
...
 	Example: "hub buck archive set-default-config --rep-factor=3 --fast-retrieval --verified-deal --trusted-miners=f08240,f023467,f09848",
 	Args:    cobra.RangeArgs(0, 1),
 	Run: func(c *cobra.Command, args []string) {
-		config := local.ArchiveConfig{}
-		stdIn, err := c.Flags().GetBool("")
+		var config local.ArchiveConfig
+		conf, err := bucks.NewConfigFromCmd(c, ".")
+		cmd.ErrCheck(err)
+
+		stdIn, err := c.Flags().GetBool("stdin")
```
jsign: stdin flag name was missing.

cmd/buck/cli/archive.go
```diff
@@ -63,47 +72,71 @@ var setDefaultArchiveConfigCmd = &cobra.Command{
...
 		} else {
-			repFactor, err := c.Flags().GetInt("rep-factor")
+			ctx, cancel := context.WithTimeout(context.Background(), cmd.Timeout)
+			defer cancel()
+			buck, err := bucks.GetLocalBucket(ctx, conf)
 			cmd.ErrCheck(err)
-			config.RepFactor = repFactor
-
-			dealMinDuration, err := c.Flags().GetInt64("deal-min-duration")
+			config, err = buck.DefaultArchiveConfig(ctx)
```
jsign, line 82: We assign to `config` the current default storage-config, then we mutate if flags are explicitly set.

cmd/buck/cli/cli.go
```diff
@@ -81,15 +81,15 @@ func Init(baseCmd *cobra.Command) {
...
 	setDefaultArchiveConfigCmd.Flags().StringSliceP("excluded-miners", "x", nil, "Miner addresses that should not be used")
 	setDefaultArchiveConfigCmd.Flags().StringSliceP("trusted-miners", "t", nil, "Miner addresses that must be used")
 	setDefaultArchiveConfigCmd.Flags().StringSliceP("country-codes", "c", nil, "Select miners with specific countries")
 	setDefaultArchiveConfigCmd.Flags().Uint64P("max-price", "p", 0, "Maximum price that will be spent per RepFactor in AttoFIL/GiB per Epoch")
 	setDefaultArchiveConfigCmd.Flags().BoolP("fast-retrieval", "f", true, "Created deals should enable the fast retrieval feature")
 	setDefaultArchiveConfigCmd.Flags().BoolP("verified-deal", "v", false, "Deal is originating from a verified client with DataCap")
-	setDefaultArchiveConfigCmd.Flags().Int64P("deal-start-offset", "e", 8640, "Epochs in the future impose a deadline for deals to be on-chain")
-	setDefaultArchiveConfigCmd.Flags().BoolP("", "i", false, "Set config from stdin JSON")
+	setDefaultArchiveConfigCmd.Flags().Int64P("deal-start-offset", "e", 0, "Epochs in the future impose a deadline for deals to be on-chain")
```
jsign: Since not being explicit about flags means _do nothing_, we avoid the problem of introducing more default-values here. (default-values doesn't matter).

cmd/buck/cli/archive.go
andrewxhill: ah no, this _was_ right i think because it was accepting a flag called `--` or -i`. so `GetBool("")` would grab that naked double hyphen. but if changed to `stdin`, makes sense now

cmd/hub/cli/fil_index.go
```diff
@@ -289,7 +289,14 @@ var filCalculateDealPrice = &cobra.Command{
 		cmd.Message("Duration in epochs: %d", res.DurationEpochs)
 		cmd.RenderTable([]string{"miner", "cost", "verified-client cost"}, data)
 
-		fmt.Printf("%s\n", aurora.Bold(aurora.Green("-- Lotus CLI --")))
+		fmt.Printf("%s\n", aurora.Bold(aurora.Green("-- Buckets --")))
```
jsign, line 292: Explain in `calculate`.

andrewxhill: 💥 so nice!

jsign (approved):

PR #517, opened 2021-03-09
docs(hub): fix hub bucket add description

sebastinez:
Hi folks,

A little typo I think I stumbled upon, the add command is described as `Add adds a UnixFs DAG...` but I think it should say `Adds a UnixFs DAG...` to avoid confusion.

```
Available Commands:
  add         Add adds a UnixFs DAG locally at path
  archive     Create a Filecoin archive
  cat         Cat bucket objects at path
```

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ Usage:
   buck [command]
 
 Available Commands:
-  add         Add adds a UnixFs DAG locally at path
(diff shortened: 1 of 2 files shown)


sanderpick (approved):
Thanks! Yep, better wording 🙏

PR #518, opened 2021-03-09
Improve UX and add verified-client awareness in bucket archiving

jsign:
This PR:
- Fixes a missing mapping that would make `hub buck set-default-config` not save the correct value for `verifiedDeals`.
- The `Archive(...)` API checks if the bucket size is above a defined minimum (can be tuned via a new flag/env).
- The maximum archive size was changed from a `const` to a flag/env. The default value was changed from 64GiB to 32GiB, as nobody in the network is using 64GiB sector-sizes yet.
- `hub fil addrs` also shows information about verified-clients (i.e: remaining data-cap). 
- `hub buck archive` command now is smarter and guides the user to avoid common pitfalls:
  - By default, if the wallet address of the account is verified, it will auto-tune the storage-config used for _this_ archive to do a verified-deal. This behavior can be disabled using `--skip-verified-deal-override`, to let the user use whatever value they configured in their default storage-config.
  - If the wallet address is verified, by the remaining data-cap is zero, it will be warned with a stdout message, and it won't autotune the `verifiedDeal` flag of the storage-config. (In a point below, it will be warned of making an unverified deal, so this won't go unnoticed without a prompt).
  - If the wallet address is unverified, it will be warned with a prompt and suggested to click a link to get automatically verified using its GH account.
  - If the default storage-config doesn't have `trustedMiners`, it will be warned also giving a suggestion to use the Miner Index commands which can help to select miners and auto-creating the CLI commands to copy-paste and fix it.
  - If the wallet address has zero balance, it will fail-fast since it doesn't make sense to make a deal.
  - If the storage-config has `maxPrice=0` (not bounded), it will warn saying that's dangerous.
  - If the bucket size is below the minimum accepted, it will fail-fast at the API response.
  - If the local bucket has unpushed changes, it will warn the user. `hub buck archive` archives the last pushed bucket, so this sounds good to surface to be sure that the user will archive what's intending.
  - If somehow the user enabled `verifiedDeal` attribute by `hub buck archive set-default-config` but its wallet address isn't verified, also warn about that. Most probably the user wasn't aware that being able to make a verified deal isn't just about that flag, but also getting a verified wallet address.
  - If by any case, the resulting scenario will be triggering an unverified deal, it will be warned and asked if it's ok to continue. This situation can happen through various cases/branches:
    - Simply having an unverified wallet address.
    - Having a verified wallet address, but 0 remaining data-cap
    - Having a verified wallet address, but using the `--skip-verified-deal-override`.
    - Maybe more combinations!

A worst-case scenario would look like this:
image
(The zero-balance case isn't shown here because it will fail the CLI command before other checks, but assume it's there too).

Also, the `Archive(..)` API by default auto-tunes the storage-config used for the archive if the wallet-address is a verified client.
If the caller wants to avoid this logic, it can use a new request attribute `SkipAutomaticVerifiedDeal`.

This PR sits on top of: https://github.com/textileio/powergate/pull/795


api/bucketsd/service.go
```diff
@@ -50,8 +50,6 @@ import (
 const (
 	// chunkSize for get file requests.
 	chunkSize = 1024 * 32 // 32 KiB
-	// maxArchiveSize is the max bucket size that can be archived to filecoin.
```
jsign, line 53: @sanderpick , I think you created this constant. I realized that there was also some `MaxBucketArchiveSize` (see line 112), but was unused and lacked the wiring to flags/envs.
So I deleted this const, wire things, and let this be configurable via flags.

api/bucketsd/service.go
```diff
@@ -3343,6 +3345,7 @@ func toPbArchiveConfig(config *mdb.ArchiveConfig) *pb.ArchiveConfig {
 			MaxPrice:        config.MaxPrice,
 			FastRetrieval:   config.FastRetrieval,
 			DealStartOffset: config.DealStartOffset,
+			VerifiedDeal:    config.VerifiedDeal,
```
jsign, line 3348: Missing mapping mentioned in the PR description!

go.mod
```diff
@@ -75,7 +76,7 @@ require (
 	github.com/textileio/go-assets v0.0.0-20200430191519-b341e634e2b7
 	github.com/textileio/go-ds-mongo v0.1.5-0.20201230201018-2b7fdca787a5
 	github.com/textileio/go-threads v1.0.3-0.20201216032729-f7b034a0de80
-	github.com/textileio/powergate/v2 v2.2.0
+	github.com/textileio/powergate/v2 v2.2.1-0.20210308170300-1912f81755cf
```
jsign, line 79: Whenever I merge and tag https://github.com/textileio/powergate/pull/795, I'll update to the final tag here.
I'll test things in staging before merging and tagging, then I can wrap up all things.

cmd/buck/cli/archive.go
```diff
@@ -4,9 +4,11 @@ import (
 	"bytes"
```
jsign, line 4: Here we're doing mostly what was described in the PR description about more checks and prompts to avoid common pitfalls.

api/bucketsd/service.go
sanderpick: Nice, good call

api/bucketsd/service.go
sanderpick: 👍

cmd/buck/cli/archive.go
```diff
@@ -197,25 +211,104 @@ var archiveCmd = &cobra.Command{
...
+		opts = append(opts, local.WithArchiveConfig(config))
 
-		conf, err := bucks.NewConfigFromCmd(c, ".")
+		addrs, err := buck.Addresses(ctx)
 		cmd.ErrCheck(err)
-		ctx, cancel := context.WithTimeout(context.Background(), cmd.Timeout)
-		defer cancel()
-		buck, err := bucks.GetLocalBucket(ctx, conf)
+		if len(addrs.Addresses) != 1 {
+			cmd.Error(fmt.Errorf("There should be exactly one wallet address but are there're %d", len(addrs.Addresses)))
```
sanderpick, line 231: "There should be exactly one wallet address but there are %d"

sanderpick (approved):
Very nice UX flow! LGTM.

cmd/buck/cli/archive.go
jsign: Oops, thanks!

PR #521, opened 2021-03-10
Wait manager/runner

asutula:
This PR refactors the previously overly complex and problematic sendfil `service.go` into a few components that are more simple, easy to understand, and also hopefully work.

- `store/store.go` - Encapsulation of mongo datastore operations, exposes an API with clean functionality supporting required operations on/around `Txn` structs.
- `waitrunner/waitrunner.go` - A concurrent worker that waits for a single message to become active on chain, fail, or times out. Uses `store`.
- `waitmanager/waitmanager.go` - Manages the async concurrent processes of monitoring many `Txn`'s underlying filecoin message status. Allows subscription and notifies interested callers of updates. Uses `waitrunner` and `store`.
- `service.go` - The gRPC service implementation which now includes only minimal logic.

This PR addresses other smaller feedback from the previous PR, I'll make note of those things in the code.


api/sendfild/service/store/store.go
```diff
@@ -0,0 +1,435 @@
...
+		log.Error("no document matched calling ReplaceOne to update txn")
+		return ErrNotFound
+	}
+	return nil
+}
+
+func (s *Store) ListTxns(ctx context.Context, req *pb.ListTxnsRequest) ([]*Txn, error) {
+	findOpts := options.Find()
+	findOpts = findOpts.SetLimit(req.PageSize)
+	findOpts = findOpts.SetSkip(req.PageSize * req.Page)
```
asutula, line 261: Using page and page size here to convert to limit and skip. Page is 0 based.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,152 @@
...
+type WaitManager struct {
+	clientBuilder lotus.ClientBuilder
+	store         *store.Store
+	confidence    uint64
+	waitTimeout   time.Duration
+	waiting       map[primitive.ObjectID]*WaitRunner
+	waitingLck    sync.Mutex
+	ticker        *time.Ticker
+	mainCtx       context.Context
+	mainCtxCancel context.CancelFunc
```
asutula, line 26: Using contexts quite carefully here and passing them down into the wait runners to ensure things are cleaned up properly.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (w *WaitManager) Subscribe(objID primitive.ObjectID, messageCid string, listener chan WaitResult) (CancelListenerFunc, error) {
+	runner, err := w.getOrCreateRunner(objID, messageCid)
+	if err != nil {
+		return nil, fmt.Errorf("getting wait runner: %v", err)
+	}
+	cancel := runner.AddListener(listener)
+	runner.Start()
+	return cancel, nil
+}
```
asutula, line 70: These two funcs are similar but `Subscribe` adds the provided listener before starting the runner and of course returns a cancelation function.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,152 @@
...
+			case <-listener:
+				err = w.deleteWaitRunner(objID)
+			case <-w.mainCtx.Done():
+				err = w.deleteWaitRunner(objID)
+			}
+			if err != nil {
+				log.Errorf("deleting wait runner: %v", err)
+			}
+		}()
+	}
```
asutula, line 104: If we are creating a new runner, we initialize it by adding it to our in memory map of runners, and we add a listener that is used to `Close`/delete the runner when the runner finishes its work or when this wait manager is `Close`d.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,152 @@
...
+		latestMessageCid, err := txn.LatestMsgCid()
+		if err != nil {
+			return err
+		}
+		if err := w.RegisterTxn(txn.ID, latestMessageCid.Cid); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
asutula, line 124: On startup and whenever the ticker ticks, we query for all pending txns and register them for monitoring.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,152 @@
...
+	w.waitingLck.Lock()
+	defer w.waitingLck.Unlock()
+	runner, ok := w.waiting[objID]
+	if !ok {
+		return nil
+	}
+	err := runner.Close()
+	delete(w.waiting, objID)
+	return err
+}
```
asutula, line 152: When deleting a runner from our map of runners, we also `Close` it.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/filecoin-project/lotus/api/apistruct"
+	"github.com/ipfs/go-cid"
+	"github.com/textileio/powergate/v2/lotus"
+	"github.com/textileio/textile/v2/api/sendfild/service/store"
+)
+
+type WaitResult struct {
+	LatestMessageCid string
+	Err              error
+}
```
asutula, line 19: This is the type we notify subscribers with.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+	waitTimeout      time.Duration
+	store            *store.Store
+	lotusClient      *apistruct.FullNodeStruct
+	closeLotusClient func()
+	listeners        []chan WaitResult
+	lock             sync.Mutex
+	mainCtx          context.Context
+	mainCtxCancel    context.CancelFunc
+	waitCtx          context.Context
+	waitCtxCancel    context.CancelFunc
```
asutula, line 33: We need a main context and a wait context, will be clear why below.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return w, nil
+}
+
+func (w *WaitRunner) Start() {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if w.waitCtxCancel != nil {
+		return
+	}
```
asutula, line 62: `Start` is meant to be a no-op if the runner is already started. We use `w.waitCtxCancel` property as a flag to detect if we're started or not. We synchronize access to that check to be safe when calls from many goroutines are coming in.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+	w.waitCtx, w.waitCtxCancel = context.WithTimeout(w.mainCtx, w.waitTimeout)
+	w.waitAndNotify()
```
asutula, line 64: The process of waiting is triggered exactly once here.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	w.closeLotusClient()
+	w.waitCtxCancel()
+	w.mainCtxCancel()
+	return nil
+}
+
+func (w *WaitRunner) waitAndNotify() {
+	go func() {
```
asutula, line 101: We spin this off in a goroutine. This is pretty straight forward really. We call lotus and wait for the result, updating the `Txn` state in the `Store` accordingly along the way, and notifying all listeners as needed.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+			w.notify(w.messageCid)
+			return
+		}
+
+		if err := w.store.SetWaiting(w.mainCtx, w.messageCid, true); err != nil {
+			w.notifyErr(fmt.Errorf("setting txn to waiting: %v", err))
+			return
+		}
+
+		res, err := w.lotusClient.StateWaitMsg(w.waitCtx, c, w.confidence)
```
asutula, line 117: Notice the use of `waitCtx` here where elsewhere we use `mainCtx`. This is because the `waitCtx` is set to a reasonable timeout, and we want that, and in the case it does time out, we need the `mainCtx` to do things like update the `Store`.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+				log.Errorf("calling StateWaitMsg block not found: %s", err.Error())
+				if err := w.store.FailTxn(w.mainCtx, w.messageCid, "block not found in lotus"); err != nil {
+					w.notifyErr(fmt.Errorf("failing txn in store: %v", err))
+					return
+				}
+				w.notify(w.messageCid)
+				return
+			}
+			// ugly but seems to be the only way to detect this
+			if err.Error() == "context canceled" {
```
asutula, line 136: Not cool that they swallow the `context.DeadlineExceeded` error in the lotus client and return this one. I saw it also contains a `Code: 1`, so maybe there are some error types defined in their json rpc client code that we can use to make comparisons?

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+				w.notifyErr(fmt.Errorf("waiting for txn status timed out, but txn is still processing, query txn again if needed"))
```
asutula, line 137: This is the "useful" error to let clients know that the error is a timeout and that the txn is still processing.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+			w.notifyErr(fmt.Errorf("calling StateWaitMsg: %v", err))
+			return
+		}
+
+		if res.Receipt.ExitCode.IsError() {
+			if res.Receipt.ExitCode.IsSendFailure() {
+				log.Errorf("received exit code send failure: %s", res.Receipt.ExitCode.String())
+			} else {
+				log.Infof("received exit code error: %s", res.Receipt.ExitCode.String())
+			}
```
asutula, line 149: Just some logging for informational purposes.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+				return
+			}
+
+			cidToNotify := c.String()
+			if res.Message.Defined() {
+				cidToNotify = res.Message.String()
+			}
+
+			w.notify(cidToNotify)
+			return
```
asutula, line 161: Depending on the type of receipt exit code error, I wasn't sure if some cases would still result in a message cid, so just doing this check just in case and calling back to the known message cid our runner was created with.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		if err := w.store.ActivateTxn(w.mainCtx, w.messageCid, res.Message.String()); err != nil {
+			w.notifyErr(fmt.Errorf("activating txn in store: %v", err))
+			return
+		}
+		w.notify(res.Message.String())
```
asutula, line 168: If we got this far, the message is active, so we update the txn to reflect that and notify listeners.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (w *WaitRunner) notify(messageCid string) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	for _, l := range w.listeners {
+		l <- WaitResult{LatestMessageCid: messageCid}
+		close(l)
+	}
+}
```
asutula, line 188: Just some helpers.

asutula:
@jsign I noticed a couple more tests I want to write (the `Store` especially), but I think this is ready for some 👀. Feeling pretty good about it, but I'm sure there are some things to discuss.

jsign (approved):
LGTM, left some comments but definitely much better than the previous version.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+func (w *WaitManager) Subscribe(objID primitive.ObjectID, messageCid string, listener chan WaitResult) (CancelListenerFunc, error) {
+	runner, err := w.getOrCreateRunner(objID, messageCid)
+	if err != nil {
+		return nil, fmt.Errorf("getting wait runner: %v", err)
+	}
+	cancel := runner.AddListener(listener)
+	runner.Start()
```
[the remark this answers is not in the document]
asutula: Just because I wanted a way to add a listener before the runner is started (as in two lines above). In practice, I guess there is probably no good justification for this. If we did start the runner on creation, It is very unlikely the runner will finish before the first listener is added.

api/sendfild/service/waitmanager/waitmanager.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if err := w.RegisterTxn(txn.ID, latestMessageCid.Cid); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (w *WaitManager) bindTicker(ctx context.Context) {
+	go func() {
```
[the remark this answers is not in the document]
asutula: Yes. The ticker handles half of that responsibility. The same logic is also applied in `getOrCreateRunner` where, in the case of creating a new runner, we add a listener to watch/write to the store, even if there are no other listeners. You can think of the ticker's responsibility as being to follow up on any txns that previously timed out or had a recoverable error. It will do this work even if no external listener is added.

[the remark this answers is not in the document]
asutula: Just to be clear, there are three ways for a txn to be monitored:
1. Our internal listener that is always created when a new runner is created.
2. If an external listener is added and we aren't already monitoring the txn.
3. If the ticker detects a pending txn that we aren't already monitoring.

api/sendfild/service/waitmanager/waitmanager.go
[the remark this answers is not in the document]
asutula: I think I'll simplify this as you suggest.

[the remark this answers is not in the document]
asutula: Ok now starting the runner in the constructor.

api/sendfild/service/waitmanager/waitrunner.go
```diff
@@ -0,0 +1,188 @@
...
+	if w.waitCtxCancel != nil {
+		return
+	}
+	w.waitCtx, w.waitCtxCancel = context.WithTimeout(w.mainCtx, w.waitTimeout)
+	w.waitAndNotify()
+}
+
+type CancelListenerFunc = func()
+
+func (w *WaitRunner) AddListener(listener chan WaitResult) CancelListenerFunc {
```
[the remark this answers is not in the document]
asutula: Oh yea good call, done.

api/sendfild/service/waitmanager/waitrunner.go
asutula: Done.

api/sendfild/service/store/store.go
```diff
@@ -0,0 +1,435 @@
...
+		col:           col,
+		mainCtxCancel: cancel,
+	}
+
+	return s, nil
+}
+
+func (s *Store) NewTxn(ctx context.Context, messageCid, from, to string, amountNanoFil int64) (*Txn, error) {
+	now := time.Now()
+
```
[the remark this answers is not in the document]
asutula: Done.

api/sendfild/service/store/store.go
```diff
@@ -0,0 +1,435 @@
...
+			Keys: bson.D{primitive.E{Key: "updated_at", Value: 1}},
+		},
+	}); err != nil {
+		cancel()
+		return nil, fmt.Errorf("creating collection indexes: %v", err)
+	}
+
+	s := &Store{
+		col:           col,
+		mainCtxCancel: cancel,
```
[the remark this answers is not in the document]
asutula: Updated.

api/sendfild/service/service.go
```diff
@@ -513,169 +223,36 @@ func (s *Service) Summary(ctx context.Context, req *pb.SummaryRequest) (*pb.Summ
...
-		delete(s.waiting, tx.ID)
-		s.waitingLck.Unlock()
-
-		ch <- waitResult{txn: tx}
-	}()
-
-	return ch
-}
-
 func (s *Service) Close() error {
```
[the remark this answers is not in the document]
asutula: Updated.

api/sendfild/service/service.go
```diff
@@ -694,26 +271,36 @@ func (s *Service) Close() error {
...
 	if err != nil {
-		log.Errorf("calling ReplaceOne to update txn: %v", err)
-		return err
+		return nil, status.Errorf(codes.Internal, "subscribing: %v", err)
 	}
-	if res.MatchedCount == 0 {
-		log.Error("no document matched calling ReplaceOne to update txn")
-		return fmt.Errorf("no matched txn document to replace")
+	select {
+	case res := <-listener:
```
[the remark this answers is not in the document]
asutula: Added.

PR #522, opened 2021-03-10
buckets: PathsPaths: avoid bucket race

sanderpick:
- Fixes races in `PushPaths` handler when accessing `buck` 
- Fixes `buck` CLI Miner Index client gRPC options when `api` is local


api/bucketsd/service.go
```diff
@@ -490,6 +490,10 @@ func (s *Service) pinBlocks(ctx context.Context, nodes []ipld.Node) (context.Con
 
 	// Check context owner's storage allowance
 	owner, ok := buckets.BucketOwnerFromContext(ctx)
+	if ok {
+		log.Debugf("pinBlock: storage: %d used, %d available, %d delta",
+			owner.StorageUsed, owner.StorageAvailable, owner.StorageDelta)
+	}
```
sanderpick, line 496: Some logging around available storage that I'll use in a follow up PR addressing that issue.

api/bucketsd/service.go
```diff
@@ -1950,6 +1958,7 @@ func (s *Service) PushPaths(server pb.APIService_PushPathsServer) error {
 	if buckRoot != "" && buckRoot != buck.Path {
 		return status.Error(codes.FailedPrecondition, buckets.ErrNonFastForward.Error())
 	}
+	verifyBuck := buck.Copy()
```
sanderpick, line 1961: Copy the bucket instance for use in the queue. When a file is added to the queue, we just need to check if the requestor is allowed to make the change. We can re-add the changes serially when the file is done being added to IPFS, i.e., we get a result in `addedCh`.

api/bucketsd/service.go
```diff
@@ -2024,43 +2033,43 @@ func (s *Service) PushPaths(server pb.APIService_PushPathsServer) error {
...
 		}
 		return err
 	}
 
 	for {
 		select {
 		case res := <-addedCh:
-			fn, err := s.IPFSClient.ResolveNode(ctx, res.resolved)
+			ctxLock.RLock()
+			newCtx := ctx
```
sanderpick, line 2051: Ensuring `ctx` is not accessed anywhere during processing of `res`.

api/bucketsd/service.go
```diff
@@ -2069,16 +2078,17 @@ func (s *Service) PushPaths(server pb.APIService_PushPathsServer) error {
...
-				ctxLock.Lock()
-				ctx = ctx2
-				ctxLock.Unlock()
 			}
 			buck.Path = dir.String()
 			buck.UpdatedAt = time.Now().UnixNano()
+			buck.SetMetadataAtPath(res.path, tdb.Metadata{
+				UpdatedAt: buck.UpdatedAt,
+			})
+			buck.UnsetMetadataWithPrefix(res.path + "/")
```
sanderpick, line 2091: Since processing of `res` is handled serially, we can safely alter the bucket right after the UnixFS DAG is altered.

cmd/cmd.go
```diff
@@ -123,17 +123,15 @@ func NewClients(hubTarget string, isHub bool, minerIndexTarget string) *Clients
...
 			creds := credentials.NewTLS(&tls.Config{})
-			mindexOpts = append(hubOpts, grpc.WithTransportCredentials(creds))
-			auth.Secure = true
+			minerIndexOpts = append(minerIndexOpts, grpc.WithTransportCredentials(creds))
 		} else {
-			mindexOpts = append(mindexOpts, grpc.WithInsecure())
+			minerIndexOpts = append(minerIndexOpts, grpc.WithInsecure())
 		}
 
-		c.MinerIndex, err = mi.NewClient(minerIndexTarget, mindexOpts...)
```
sanderpick, line 136: This was causing the client to be configured with `grpc.WithTransportCredentials` and `grpc.WithInsecure` in the case of a local hub. Since the miner index is not authenticated, there's no need grab the `grpc.WithPerRPCCredentials` from the hub client options.

core/usage_interceptor.go
```diff
@@ -172,10 +172,18 @@ func (t *Textile) preUsageFunc(ctx context.Context, method string) (context.Cont
...
+			// Customer is paying for storage, quota is unbounded
 			owner.StorageAvailable = int64(math.MaxInt64)
 		} else if now.Unix() < cus.GracePeriodEnd {
+			// Customer is in grace period, quota is grace
+			owner.StorageAvailable = cus.DailyUsage["stored_data"].Grace
+		} else if cus.GracePeriodEnd == 0 {
+			// Customer has not started grace period, but there's no way to start the grace period _after_
+			// a successful request, so we use grace quota here and let the post usage handler start the
+			// grace period in the event the request exceeds the free quota
 			owner.StorageAvailable = cus.DailyUsage["stored_data"].Grace
```
sanderpick, line 184: Use grace quota before grace starts... see comments for more

api/bucketsd/service.go
jsign: Cool, so making a copy to avoid concurrent access.

Maybe it would be a good idea to later rename this variable to "readOnlyBuck" or something that can hint a future development that it shouldn't use this *copy* for anything but reading. Now that in the same scope are two buckets (one copied and one the real one to save), it might be error-prone to do *new* changes in the wrong one.

jsign (approved):

PR #525, opened 2021-03-11
gateway: use an io.Reader to serve Cid data

jsign:
Use an `io.Reader` instead of loading all the content in a slice, which can lead to high RAM usage and potential OOM crashes.
The mime detection still works crafting a new reader to return the full output.

### DIFF
--- a/gateway/buckets.go
+++ b/gateway/buckets.go
@@ -4,10 +4,8 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"mime"
 	"net/http"
 	"path"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -17,6 +15,7 @@ import (
 	"github.com/textileio/go-threads/core/thread"
 	"github.com/textileio/go-threads/db"
 	"github.com/textileio/textile/v2/api/bucketsd/client"
+	bucketsclient "github.com/textileio/textile/v2/api/bucketsd/client"
 	"github.com/textileio/textile/v2/api/common"
 	"github.com/textileio/textile/v2/buckets"
 	mdb "github.com/textileio/textile/v2/mongodb"
@@ -84,9 +83,12 @@ func (g *Gateway) renderBucketPath(c *gin.Context, ctx context.Context, threadID
 		return
 	}
 	if !rep.Item.IsDir {
-		if err := g.buckets.PullPath(ctx, buck.Key, pth, c.Writer); err != nil {
+		contentType, r, err := getContentTypeFromPullPath(ctx, g.buckets, buck.Key, pth)
+		if err != nil {
 			render404(c)
 		}
+		c.Writer.Header().Set("Content-Type", contentType)
+		io.Copy(c.Writer, r)
 	} else {
 		var base string
 		if g.subdomains {
@@ -132,7 +134,7 @@ func (g *Gateway) renderBucketPath(c *gin.Context, ctx context.Context, threadID
 type serveBucketFS interface {
 	GetThread(ctx context.Context, key string) (thread.ID, error)
 	Exists(ctx context.Context, bucket, pth string) (bool, string)
-	Write(ctx context.Context, bucket, pth string, writer io.Writer) error
+	GetPathWithContentType(ctx context.Context, bucket, pth string) (string, io.Reader, error)
 	ValidHost() string
 }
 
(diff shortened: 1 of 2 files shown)


gateway/ipfs.go
```diff
@@ -79,23 +80,30 @@ func (g *Gateway) renderIPFSPath(c *gin.Context, base, pth string) {
...
-		contentType := http.DetectContentType(data)
-		c.Writer.Header().Set("Content-Type", contentType)
-		c.Render(200, render.Data{Data: data})
+
+		renderError(c, http.StatusBadRequest, err)
+		return
+	}
+	defer f.Close()
+
+	var buf [512]byte
```
jsign, line 91: We reserve 512 bytes as the maximum amount of space needed for MIME detection.

gateway/ipfs.go
```diff
@@ -79,23 +80,30 @@ func (g *Gateway) renderIPFSPath(c *gin.Context, base, pth string) {
...
+	n, err := io.ReadAtLeast(f, buf[:], len(buf))
+	if err != nil && err != io.ErrUnexpectedEOF {
+		renderError(c, http.StatusInternalServerError, err)
+		return
 	}
```
jsign, line 96: Try to read at least 512 bytes.
The only possible error that we can come up here is reading less than 512 bytes if the content is less than that size. That's OK.

gateway/ipfs.go
```diff
@@ -79,23 +80,30 @@ func (g *Gateway) renderIPFSPath(c *gin.Context, base, pth string) {
...
+	contentType := http.DetectContentType(buf[:])
+	c.Writer.Header().Set("Content-Type", contentType)
+	c.Render(200, render.Reader{ContentLength: -1, Reader: io.MultiReader(bytes.NewReader(buf[:n]), f)})
```
jsign, line 99: Since we're returning a reader, we can't pre-calculate the content-length of the response.

That's fine, but we should be explicit with content-length -1 since I looked up the `c.Render` impl, and that's the only way to signal `net/http` server that this return should be chunk encoded.

gateway/ipfs.go
```diff
@@ -79,23 +80,30 @@ func (g *Gateway) renderIPFSPath(c *gin.Context, base, pth string) {
...
 }
 
-func (g *Gateway) openPath(ctx context.Context, pth path.Path) ([]byte, error) {
+func (g *Gateway) openPath(ctx context.Context, pth path.Path) (io.ReadCloser, error) {
 	f, err := g.ipfs.Unixfs().Get(ctx, pth)
 	if err != nil {
 		return nil, err
 	}
-	defer f.Close()
```
jsign, line 98: This is now the responsibility of the caller. 
Can be found in line 89.

sanderpick (approved):
LGTM! There's a number of spots in https://github.com/textileio/textile/blob/master/gateway/buckets.go where we do `g.buckets.PullPath` by directly passing the request writer. Worth adding the handling there?

gateway/buckets.go
```diff
@@ -84,9 +83,12 @@ func (g *Gateway) renderBucketPath(c *gin.Context, ctx context.Context, threadID
 		return
 	}
 	if !rep.Item.IsDir {
-		if err := g.buckets.PullPath(ctx, buck.Key, pth, c.Writer); err != nil {
+		contentType, r, err := getContentTypeFromPullPath(ctx, g.buckets, buck.Key, pth)
+		if err != nil {
 			render404(c)
 		}
+		c.Writer.Header().Set("Content-Type", contentType)
+		io.Copy(c.Writer, r)
```
jsign, line 91: The main idea is always relying on `getContentTypeFromPullPath` which returns the detected content type, and an `io.Reader` to copy to the response writer.

gateway/buckets.go
```diff
@@ -165,24 +167,23 @@ func serveBucket(fs serveBucketFS) gin.HandlerFunc {
 		exists, target := fs.Exists(ctx, key, c.Request.URL.Path)
 		if exists {
 			c.Writer.WriteHeader(http.StatusOK)
-			ctype := mime.TypeByExtension(filepath.Ext(c.Request.URL.Path))
-			if ctype == "" {
-				ctype = "application/octet-stream"
-			}
-			c.Writer.Header().Set("Content-Type", ctype)
-			if err := fs.Write(ctx, key, c.Request.URL.Path, c.Writer); err != nil {
```
jsign, line 173: We don't use content-type detection by extension anymore.

gateway/buckets.go
```diff
@@ -165,24 +167,23 @@ func serveBucket(fs serveBucketFS) gin.HandlerFunc {
...
+			contentType, r, err := fs.GetPathWithContentType(ctx, key, c.Request.URL.Path)
+			if err != nil {
 				renderError(c, http.StatusInternalServerError, err)
 			} else {
+				c.Writer.Header().Set("Content-Type", contentType)
+				io.Copy(c.Writer, r)
```
jsign, line 175: Simply call `fs.GetPathWithContentType(..)`, set that content-type and write to the response writer.

gateway/buckets.go
```diff
@@ -254,16 +259,35 @@ func (g *Gateway) renderWWWBucket(c *gin.Context, key string) {
...
+		}
+	}()
+
+	contentType, r, err := detectReaderContentType(pr)
+	if err != nil {
+		return "", nil, fmt.Errorf("detecting content-type: %s", err)
+	}
+
+	return contentType, r, nil
+}
```
jsign, line 289: The magic is concentrated only here.
We give `PullPath` a writer from a Pipe, use the reader to let `detectReaderContentType` figure out the content-type and return a new reader (multireader).

gateway/ipfs.go
```diff
@@ -88,15 +88,25 @@ func (g *Gateway) renderIPFSPath(c *gin.Context, base, pth string) {
 	}
 	defer f.Close()
```
jsign, line 90: A bit weird diff, but this is the same code as pre-review, just that the content-type detection logic was extracted to a new `detectReaderContentType`, that now is also leveraged by `gateway/buckets.go`, so no repeated code.

gateway/buckets.go
jsign: I believe this is the correct solution. Adding more bloat to `PullPath` API doesn't sound correct to me since not necessarily all clients are interested in content-type detection.

gateway/buckets.go
```diff
@@ -84,9 +83,12 @@ func (g *Gateway) renderBucketPath(c *gin.Context, ctx context.Context, threadID
 		return
 	}
 	if !rep.Item.IsDir {
-		if err := g.buckets.PullPath(ctx, buck.Key, pth, c.Writer); err != nil {
+		contentType, r, err := getContentTypeFromPullPath(ctx, g.buckets, buck.Key, pth)
+		if err != nil {
 			render404(c)
```
sanderpick, line 88: I could be reading the diff wrong, but `return` needed after `render404`?

sanderpick (approved):
LGTM! 🎉

gateway/buckets.go
jsign: Oh yeah, I'll add that and merge 👍🏼

PR #527, opened 2021-03-15
Allowed from addrs list

asutula:
Pretty straight forward. If you don't want to be forced to provide an address list, you can opt out explicitly using another flag.

### DIFF
--- a/api/sendfild/main.go
+++ b/api/sendfild/main.go
@@ -73,10 +73,6 @@ var (
 				Key:      "allowed_from_addrs",
 				DefValue: []string{},
 			},
-			"allowEmptyFromAddrs": {
-				Key:      "allow_empty_from_addrs",
-				DefValue: false,
-			},
 		},
 		EnvPre: "SENDFIL",
 		Global: true,
@@ -145,11 +141,7 @@ func init() {
 	rootCmd.PersistentFlags().StringSlice(
 		"allowedFromAddrs",
 		config.Flags["allowedFromAddrs"].DefValue.([]string),
-		"A list of filecoin address allowed to be used as transaction from addresses")
-	rootCmd.PersistentFlags().Bool(
-		"allowEmptyFromAddrs",
-		config.Flags["allowEmptyFromAddrs"].DefValue.(bool),
-		"Allow not specifying --allowedFromAddrs list")
+		"A comma separated list of filecoin address allowed to be used as transaction from addresses or * for all addresses")
 
 	err := cmd.BindFlags(config.Viper, rootCmd, config.Flags)
 	cmd.ErrCheck(err)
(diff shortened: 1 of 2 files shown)


api/sendfild/main.go
```diff
@@ -69,6 +69,14 @@ var (
...
 				DefValue: time.Minute,
 			},
+			"allowedFromAddrs": {
+				Key:      "allowed_from_addrs",
+				DefValue: []string{},
+			},
+			"allowEmptyFromAddrs": {
+				Key:      "allow_empty_from_addrs",
+				DefValue: false,
+			},
```
asutula, line 79: We default to an empty list which means every call to send fil will fail by default. If you want to allow an empty list which is interpreted as "allow all from addresses", you have to explicitly set the `--allowEmptyFromAddrs` flag (which defaults to `false`).

api/sendfild/service/service.go
```diff
@@ -62,6 +64,10 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	if !config.AllowEmptyFromAddrs && len(config.AllowedFromAddrs) == 0 {
+		return nil, fmt.Errorf("empty allowed from addrs not allowed")
+	}
```
asutula, line 69: Error if we detect prohibited conditions.

api/sendfild/service/service.go
```diff
@@ -91,6 +97,17 @@ func New(config Config) (*Service, error) {
...
+	isAllowedFromAddr := s.config.AllowEmptyFromAddrs
+	for _, allowedFromAddr := range s.config.AllowedFromAddrs {
+		if allowedFromAddr == req.From {
+			isAllowedFromAddr = true
+			break
+		}
+	}
+	if !isAllowedFromAddr {
+		return nil, status.Error(codes.InvalidArgument, "from address not allowed")
+	}
```
asutula, line 109: Check our allowed list, or accept all from addresses if the config says that is ok.

api/sendfild/main.go
textileben: How about a single, required, setting. It can be a list of allowed addresses, or special case `*` which allows from any address?

jsign: >How about a single, required, setting. It can be a list of allowed addresses, or special case * which allows from any address?

Cool, that sounds good to me.

jsign (approved):

api/sendfild/main.go
asutula: Sure that works.

api/sendfild/main.go
```diff
@@ -186,14 +196,16 @@ var rootCmd = &cobra.Command{
...
-			Debug:              debug,
+			Listener:            listener,
+			ClientBuilder:       cb,
+			MongoUri:            mongoUri,
+			MongoDbName:         mongoDb,
+			MessageWaitTimeout:  messageTimeout,
+			MessageConfidence:   messageConfidence,
+			RetryWaitFrequency:  retryWaitFrequency,
+			AllowedFromAddrs:    allowedFromAddrs,
+			AllowEmptyFromAddrs: len(allowedFromAddrs) == 1 && allowedFromAddrs[0] == "*",
```
asutula, line 207: Ok @textileben here it is. I view `*` as CLI sugar kind of thing, so that still needs to be translated into real data here to be passed into the `Service` constructor.

textileben (approved):
lgtm

PR #533, opened 2021-03-18
Enable Filecoin retrievals in Buckets

jsign:
This PR sits on top of https://github.com/textileio/powergate/pull/807

We had this feature hidden because of a [problem](https://github.com/filecoin-project/lotus/issues/5361) when configuring Lotus when using `IpfsOnlineMode=true`, which didn't make retrievals possible. The Powergate PR workaround this problem, and unblocks the feature.

This PR also added some extra fixes and sections in the Miner Index CLI.

### DIFF
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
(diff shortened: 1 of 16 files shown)


sanderpick (approved):
LGTM! 🎉

PR #536, opened 2021-03-18
Github actions to build docker images for CD

textileben:
On pushes to master, generate 2 container images

- update `latest` tag with master commit (I will remove from Docker builds)
- create new image tag, `sha-{{ git short sha}}` , that will be used for a staging deployment

Signed-off-by: Ben Wilson <ben@textile.io>

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,7 @@ on:
       - master
 jobs:
   docker-build-push:
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         imageRepo:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,48 @@
+name: Docker image master branch
+on:
+  push:
+    branches:
+      - master
+jobs:
+  docker-build-push:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,48 @@
...
+          - textile/mindexd
+        include:
+          - package: textile/hub
+            dockerFile: cmd/hubd/Dockerfile
+          - package: textile/billing
+            dockerFile: api/billingd/Dockerfile
+          - package: textile/buckets
+            dockerFile: cmd/buckd/Dockerfile
+          - package: textile/mindexd
+            dockerFile: api/mindexd/Dockerfile
```
jsign, line 24: 💯

.github/workflows/docker.yml
```diff
@@ -0,0 +1,48 @@
+name: Docker image master branch
+on:
+  push:
+    branches:
+      - master
+jobs:
+  docker-build-push:
+    runs-on: self-hosted
```
jsign, line 8: I think this could run on a free-tier from GH... but maybe more work needed for Docker layer caching.
Looks good to me on self-hosted too.

jsign (approved):

PR #537, opened 2021-03-18
hub: fix update cmd when build not available

sanderpick:
`hub update` was panicing on my M1 machine because the code here was assuming the release would always be available for current os/arch. I got sucked into some more improvements. 

I spent a little time actually adding the arm64 builds, but ran into an issue with bingo on go 1.16 and some random build errors while testing locally:
```
--> darwin/arm64 error: exit status 2
Stderr: # github.com/textileio/textile/v2/cmd/hub
/usr/local/go/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
ld: warning: ignoring file /var/folders/bp/7ygrmpvn1779kd81k0ltmqkm0000gn/T/go-link-502520635/go.o, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Will resume when other things are less pressing.

### DIFF
--- a/cmd/hub/cli/version.go
+++ b/cmd/hub/cli/version.go
@@ -2,6 +2,8 @@ package cli
 
 import (
 	"context"
+	"fmt"
+	"runtime"
 
 	"github.com/blang/semver"
 	"github.com/caarlos0/spin"
@@ -12,7 +14,7 @@ import (
 )
 
 func getLatestRelease() (*su.Release, error) {
-	s := spin.New("%s Checking latest Hub CLI release")
+	s := spin.New("%s Checking for new Hub CLI version")
 	s.Start()
 	defer s.Stop()
 	config := su.Config{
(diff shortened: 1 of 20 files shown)


cmd/hub/cli/version.go
```diff
@@ -26,9 +28,12 @@ func getLatestRelease() (*su.Release, error) {
 	}
 
 	latest, found, err := updater.DetectLatest(cmd.Repo)
-	if err != nil || !found {
+	if err != nil {
 		return nil, err
+	} else if !found {
+		return nil, fmt.Errorf("%s_%s build not found", runtime.GOOS, runtime.GOARCH)
```
andrewxhill, line 34: 👍

andrewxhill (approved):

PR #551, opened 2021-04-02
Old Bucket Archive Fixes

asutula:
Took over this PR to implement the actual fixes for the following issues uncovered recently:

- The bucket archive bson key `job_status` was changed to `status` months ago, but no migration was done, leaving affected buckets in a state where their `unspecified` current archive status was preventing a new archive from being triggered.
- Hub's Powergate was reset at some point, loosing old archive data. This resulted in old bucket archive storage info from being looked up in Powergate during the creation of a new archive, so the process would fail.

There is one more issue I'm hoping to figure out today, old buckets not deleting, but still need more info from Cake and to do more research there. Maybe we just move forward with these current fixes for now.

### DIFF
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -3218,12 +3218,12 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 			}
 		} else { // Case 2.
 			res, err := s.PowergateClient.Data.CidInfo(ctxPow, oldCid.String())
-			isKnownToPowergate := err == nil || status.Convert(err).Code() != codes.NotFound
-			if err != nil && isKnownToPowergate {
+			isNotKnownToPowergate := err != nil && status.Convert(err).Code() == codes.NotFound
+			if err != nil && !isNotKnownToPowergate {
 				return nil, fmt.Errorf("looking up old storage config: %v", err)
 			}
 
-			if isKnownToPowergate && cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
+			if !isNotKnownToPowergate && cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
 				// Old storage config is the same as the new so use replace.
 				res, err := s.PowergateClient.Data.ReplaceData(ctxPow, oldCid.String(), p.Cid().String())
 				if err != nil {
@@ -3232,7 +3232,7 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 				jid = res.JobId
 			} else {
 				// New storage config.
-				if isKnownToPowergate {
+				if !isNotKnownToPowergate {
 					// Remove previous storage config.
 					_, err = s.PowergateClient.StorageConfig.Apply(
 						ctxPow,
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -3202,7 +3202,7 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 			case userPb.JobStatus_JOB_STATUS_EXECUTING, userPb.JobStatus_JOB_STATUS_QUEUED:
 				return nil, fmt.Errorf("there is an in progress archive")
 			// Case 1.c.
-			case userPb.JobStatus_JOB_STATUS_FAILED, userPb.JobStatus_JOB_STATUS_CANCELED:
+			case userPb.JobStatus_JOB_STATUS_FAILED, userPb.JobStatus_JOB_STATUS_CANCELED, userPb.JobStatus_JOB_STATUS_UNSPECIFIED:
 				res, err := s.PowergateClient.StorageConfig.Apply(
 					ctxPow,
 					p.Cid().String(),
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -3202,7 +3202,7 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 			case userPb.JobStatus_JOB_STATUS_EXECUTING, userPb.JobStatus_JOB_STATUS_QUEUED:
 				return nil, fmt.Errorf("there is an in progress archive")
 			// Case 1.c.
-			case userPb.JobStatus_JOB_STATUS_FAILED, userPb.JobStatus_JOB_STATUS_CANCELED, userPb.JobStatus_JOB_STATUS_UNSPECIFIED:
+			case userPb.JobStatus_JOB_STATUS_FAILED, userPb.JobStatus_JOB_STATUS_CANCELED:
 				res, err := s.PowergateClient.StorageConfig.Apply(
 					ctxPow,
 					p.Cid().String(),
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -3218,11 +3218,12 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 			}
 		} else { // Case 2.
 			res, err := s.PowergateClient.Data.CidInfo(ctxPow, oldCid.String())
-			if err != nil {
+			isKnownToPowergate := err == nil || status.Convert(err).Code() != codes.NotFound
+			if err != nil && isKnownToPowergate {
 				return nil, fmt.Errorf("looking up old storage config: %v", err)
 			}
 
-			if cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
+			if isKnownToPowergate && cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
 				// Old storage config is the same as the new so use replace.
 				res, err := s.PowergateClient.Data.ReplaceData(ctxPow, oldCid.String(), p.Cid().String())
 				if err != nil {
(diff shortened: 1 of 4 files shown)


jsign (approved):

asutula:
Looking like this is probably not needed. Instead a data migration could be used. Will leave this open just in case, close tomorrow once we figure out a plan.

api/bucketsd/service.go
```diff
@@ -3218,32 +3218,37 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
 			}
 		} else { // Case 2.
 			res, err := s.PowergateClient.Data.CidInfo(ctxPow, oldCid.String())
-			if err != nil {
+			isKnownToPowergate := err == nil || status.Convert(err).Code() != codes.NotFound
```
asutula, line 3221: Here we identify if Powergate is aware of the old cid by inspecting the error. I verified the Powergate client is just passing the gRPC error along here and that the Powergate server does return `codes.NotFound` in the case where the cid is unknown.

api/bucketsd/service.go
```diff
@@ -3218,32 +3218,37 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+			if err != nil && isKnownToPowergate {
 				return nil, fmt.Errorf("looking up old storage config: %v", err)
 			}
 
-			if cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
+			if isKnownToPowergate && cmp.Equal(&storageConfig, res.CidInfo.LatestPushedStorageConfig) {
```
asutula, line 3226: We can only use `ReplaceData` if Powergate is aware of the cid.

api/bucketsd/service.go
```diff
@@ -3218,32 +3218,37 @@ func (s *Service) Archive(ctx context.Context, req *pb.ArchiveRequest) (*pb.Arch
...
+						pow.WithStorageConfig(&userPb.StorageConfig{}),
+						pow.WithOverride(true),
+					)
+					if err != nil {
+						return nil, fmt.Errorf("pushing config to disable hot and cold storage: %v", err)
+					}
+					_, err = s.PowergateClient.StorageConfig.Remove(ctxPow, oldCid.String())
+					if err != nil {
+						return nil, fmt.Errorf("removing old cid storage: %v", err)
+					}
```
asutula, line 3249: In this case, we are applying a new storage config AND the cid is known to powergate, so we can do the remove step.

asutula: Then below, outside of this `if`, we always apply the new storage config.

mongodb/migrations/migrations.go
```diff
@@ -180,13 +181,63 @@ var m004 = migrate.Migration{
...
+	Description: "rename any bucket archive job_status to status",
+	Up: func(db *mongo.Database) error {
+		log.Info("migrating 005 up")
+		ctx, cancel := context.WithTimeout(context.Background(), migrateTimeout)
+		defer cancel()
+		if _, err := db.Collection("bucketarchives").UpdateMany(ctx, bson.M{}, bson.M{
+			"$rename": bson.M{"archives.current.job_status": "archives.current.status"},
+		}); err != nil {
+			return err
+		}
```
asutula, line 195: As a first step, rename any `archives.current.job_status` to `archives.current.status`. This probably could be wrapped into the aggregation below, but this is simple and I didn't want to complicate the aggregation any more.

mongodb/migrations/migrations.go
```diff
@@ -180,13 +181,63 @@ var m004 = migrate.Migration{
...
+						},
+					},
+				},
+			}}},
+			bson.D{
+				primitive.E{Key: "$out", Value: "bucketarchives"},
+			},
+		}); err != nil {
+			return err
+		}
```
asutula, line 224: You can't use `$rename` on objects that are nested in arrays, so you have to use `Aggregate`, writing the results back to the collection. This is pretty next level stuff, I'm not going to try to explain it all because it was some copy/paste/modify. I leave it to the very thorough migration test to verify this does indeed work.

mongodb/migrations/migrations.go
```diff
@@ -180,13 +181,63 @@ var m004 = migrate.Migration{
...
+		return nil
+	},
+	Down: func(db *mongo.Database) error {
+		log.Info("nothing to do for 005 down")
+		return nil
+	},
```
asutula, line 230: The state we are migration from is not a valid state for the database... it was a bug. So it doesn't make any sense to provide `Down`.

mongodb/bucketarchives.go
```diff
@@ -36,7 +36,7 @@ type DealInfo struct {
 	Miner       string `bson:"miner"`
 
 	PieceCID string `bson:"piece_cid"`
-	Size     uint64
+	Size     uint64 `bson:"size"`
```
asutula, line 39: I noticed this tag was missing. No problem because my understanding is the default bson behavior is to use the lowercase version of the property name, so nothing changes. Just did this for completeness.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
 	assert.Nil(t, key2["path"])
 }
 
+func TestMigrations_m005(t *testing.T) {
+	ctx := context.Background()
+	db := setup(t, ctx)
+	_, err := db.Collection("bucketarchives").InsertMany(ctx, []interface{}{
```
asutula, line 189: Tried to insert many different permutations and combinations of data here to make sure we test all possibilities.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
...
+						"created_at": 1000,
+					},
+					bson.M{
+						"job_id":     "job1",
+						"status":     1,
+						"created_at": 1000,
+					},
+				},
+			},
+		},
```
asutula, line 211: This represents a "new" object with all the correct `status` fields.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
...
+						"created_at": 1000,
+					},
+					bson.M{
+						"job_id":     "job4",
+						"job_status": 1,
+						"created_at": 1000,
+					},
+				},
+			},
+		},
```
asutula, line 233: This is an "old" one with the wrong fields using `job_status`.

asutula: The remainder below here are various combinations.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
...
+	require.NoError(t, err)
+	defer cursor.Close(ctx)
+
+	validateItem := func(item bson.M) {
+		require.NotEmpty(t, item["job_id"])
+		require.Contains(t, item["job_id"], "job")
+		require.Equal(t, 1000, int(item["created_at"].(int32)))
+		require.Nil(t, item["job_status"])
+		require.Equal(t, 1, int(item["status"].(int32)))
+	}
```
asutula, line 325: Little func that validates a bucket archive object. Most important thing is to make sure the id still looks valid, `job_status` no longer exists and `status` does exist.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
...
+			require.NotNil(t, item["archives"].(bson.M)["history"])
+		}
+		if item["_id"] == "id5" {
+			require.Nil(t, item["archives"].(bson.M)["current"])
+			require.NotNil(t, item["archives"].(bson.M)["history"])
+		}
+		if item["_id"] == "id6" {
+			require.NotNil(t, item["archives"].(bson.M)["current"])
+			require.Nil(t, item["archives"].(bson.M)["history"])
+		}
```
asutula, line 343: Just checking some specific items to make sure the structure of the data hasn't changed where it shouldn't have.

mongodb/migrations/migrations_test.go
```diff
@@ -183,6 +183,176 @@ func TestMigrations_m004(t *testing.T) {
...
+		if item["archives"].(bson.M)["current"] != nil {
+			validateItem(item["archives"].(bson.M)["current"].(bson.M))
+		}
+		if item["archives"].(bson.M)["history"] != nil {
+			for _, item := range item["archives"].(bson.M)["history"].(bson.A) {
+				validateItem(item.(bson.M))
+			}
+		}
```
asutula, line 351: Validate all the contained archive infos under `current` and the `history` array.

api/bucketsd/service.go
jsign: I'm a bit confused by this line, two cases:
- if `err == nil`: ok, the easy case makes sense.
- If `err != nil && status != NotFound`:  I'm not sure this means that "is known to Powergate". If the error is "timeout" due to some network problem, then we're assuming "is known to powergate"?

Maybe I'm confused.

Thinking about what you're trying to do... I can't think of some way to answer this question.

asutula: Maybe it would make more sense if I frame it from the other angle: is not known to powergate. That is a value with an easy to understand meaning and calculation. Change coming up.

asutula: Ok just pushed the update. The subsequent use is a little awkward double negative reading, but I think it's worth it to have a straight forward calculation and meaning of `isNotKnownToPowergate`, no ambiguity.

asutula: Also, just to clarify the need for this in more general terms... This is addressing the case where the `oldCid` from the latest bucket archive is not known to Powergate (because we reset Powergate at some point). This change detects the `NotFound` error, and in that case, doesn't return an error (Previously it would return any error, so the bucket would be unable to be archived). Additionally, this information must also be used to decide if we can remove a previous storage config in the case of applying a new storage config.

sanderpick (approved):
Nice, LGTM. I'm trying to remember when we nuked powergate in the hub... was that recent?

jsign:
> Nice, LGTM. I'm trying to remember when we nuked powergate in the hub... was that recent?

I don't recall ever nuking Powergate in any environment. I'm pretty sure that never happened since would be a bit headache regarding wallets and similars.

asutula:
Yea hard to say. I don't remember specifically, but I did write code to deal with that possibility, so it must have been on our minds (see [here](https://github.com/textileio/textile/blob/master/api/bucketsd/service.go#L3079) and [here](https://github.com/textileio/textile/blob/master/core/pow_interceptor.go#L108)). That code is basically the headache @jsign mentions... already dealt with haha.

PR #552, opened 2021-04-02
Add hub staging deploy step for CD

textileben:
Adding a Github Actions step to deploy the docker images build on pushes to master.

This step will receive the short sha used to tag the images and send it to the deployment endpoint.


textileben:
No longer needed

PR #554, opened 2021-05-18
feat(api): adds content-types for .svg and .json

burk3:
in our app, .svg's were not displaying in img tags due to the
content-type returned by the server being `text/plain` instead of
`image/svg+xml`.

To fix, the gateway will return the appropriate content-type,
`image/svg+xml`, for .svg files. As a bonus, .json files will
return `application/json`.

referenced mime types from https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types


burk3:
let me close, sign-off, and re-open.

PR #555, opened 2021-05-18
feat(api): adds content-types for .svg and .json

burk3:
in our app, .svg's were not displaying in img tags due to the
content-type returned by the server being `text/plain` instead of
`image/svg+xml`.

To fix, the gateway will return the appropriate content-type,
`image/svg+xml`, for .svg files. As a bonus, .json files will
return `application/json`.

referenced mime types from https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

### DIFF
--- a/gateway/ipfs.go
+++ b/gateway/ipfs.go
@@ -119,6 +119,10 @@ func detectReaderContentType(r io.Reader, pth string) (string, io.Reader, error)
 		return "text/css", reader, nil
 	case ".js":
 		return "application/javascript", reader, nil
+	case ".json":
+		return "application/json", reader, nil
+	case ".svg":
+		return "image/svg+xml", reader, nil
 	case ".txt", ".md":
 		return "text/plain", reader, nil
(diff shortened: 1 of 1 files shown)


burk3:
Looks like a Stripe API key is missing from the test runner, which I assume isn't something this change affected. While we're talking about these tests... are there any tests I can update somewhere while I'm at it to exercise this functionality? I didn't see any tests in the package where the change happened, but I guess there might be integration tests somewhere that I could integrate with. I'll take a look in a little in case nobody sees this any time soon.

Also, should someone be assigned to this so they can track? @andrewxhill since you pointed me in the right direction, or @carsonfarmer  since you blessed my message with your emoji and because I hear from my team that you are the cool person? Haven't participated enough to get a feel for how y'all operate.

andrewxhill:
Nope, looks good! We'll get this deployed next week.

PR #561, opened 2021-07-02
fix(errs): report current current delta.

andrewxhill:

### DIFF
--- a/api/bucketsd/service.go
+++ b/api/bucketsd/service.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"math/big"
 	gopath "path"
 	"regexp"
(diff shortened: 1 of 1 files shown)


api/bucketsd/service.go
```diff
@@ -511,7 +512,10 @@ func (s *Service) addPinnedBytes(ctx context.Context, delta int64) context.Conte
 	owner, ok := buckets.BucketOwnerFromContext(ctx)
 	if ok {
 		owner.StorageUsed += delta
-		owner.StorageAvailable -= delta
+		// int64(math.MaxInt64) indicates that the user has no current cap so don't deduct
+		if owner.StorageAvailable < int64(math.MaxInt64) {
+			owner.StorageAvailable -= delta
```
andrewxhill, line 517: seems logical to not try and decrease storage available if the user has uncapped storage

api/bucketsd/service.go
```diff
@@ -2401,7 +2405,7 @@ func (s *Service) updateOrAddPin(ctx context.Context, from, to path.Path) (conte
 	owner, ok := buckets.BucketOwnerFromContext(ctx)
 	if ok {
 		log.Debugf("updateOrAddPin: storage: %d used, %d available, %d delta",
-			owner.StorageUsed, owner.StorageAvailable, owner.StorageDelta)
+			owner.StorageUsed, owner.StorageAvailable, deltaSize)
```
andrewxhill, line 2408: it's reporting some value out of memory which is only some previous delta (usually 0). here, reporting the current incoming delta

api/bucketsd/service.go
jsign: Oh, didn't know `int64(math.MaxInt64)` exactly mean uncapped.

PR #564, opened 2021-08-11
Billing rabbit hole

sanderpick:
- Includes prior hot fixes from other branches.
- Updates threads to https://github.com/textileio/go-threads/pull/548 (will wait for that to be merged and released before merging this).
- Fix for the user account overage problem.
- Fixes billing issue where we were collecting 100x less than expected 💩.
- Various usage tracking fixes (see comments).
- Adds new commands for only interacting with a remote bucket:
  - `buck remote add`: add an entire file or folder to a bucket path while ignoring the local repo.
  - `buck remote rm`: remove an entire file or folder from a bucket path while ignoring the local repo.
- Starts the IPFS client in "offline mode": never fetch block from the network. This is an extra measure to protect against abuse of the public gateway. 

See https://docs.google.com/spreadsheets/d/1caZm6OVj-U48BVohQ2guKgVyJlEqGVkAxkHpUlHkurw/edit#gid=0 for a breakdown of hub billing calcs.


buckets/local/bucket.go
```diff
@@ -105,6 +105,29 @@ func (b *Bucket) Path() (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	ctx = common.NewThreadIDContext(ctx, id)
+	if b.auth != nil {
+		ctx = b.auth(ctx)
+	}
+	return ctx, nil
+}
+
```
sanderpick, line 130: Some extra methods needed for the remote commands.

api/billingd/service/service.go
```diff
@@ -987,7 +1000,8 @@ func (s *Service) handleUsage(ctx context.Context, cus *Customer, product Produc
 	}
 	update := bson.M{"daily_usage." + product.Key + ".total": total}
 
-	if total > product.FreeQuotaSize && !cus.Billable {
+	var err error
+	if total > product.FreeQuotaSize && !cus.Billable && cus.ParentKey == "" {
```
sanderpick, line 1004: Never error when checking usage for a child. If their parent is not billable and over a free quota, they will see the error when their usage is accrued on the parent, but it makes no sense to also error here... and was causing the slate issue: the parent account was billable, but the child accounts were still erroring (since they are never billable).

cmd/buck/cli/cli.go
```diff
@@ -94,6 +95,15 @@ func Init(baseCmd *cobra.Command) {
...
 	linksCmd.Flags().String("format", "default", "Display URL links in the provided format. Options: [default,json]")
+
+	remoteCmd.AddCommand(remoteAddCmd)
+	remoteCmd.AddCommand(remoteRemoveCmd)
+
+	remoteAddCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
+	remoteAddCmd.Flags().BoolP("quiet", "q", false, "Write minimal output")
+
+	remoteRemoveCmd.Flags().BoolP("yes", "y", false, "Skips the confirmation prompt if true")
+	remoteRemoveCmd.Flags().BoolP("quiet", "q", false, "Write minimal output")
```
sanderpick, line 106: New `remote add` and `remove rm` commands that function like the original `buck` command: ignore the local repo. This is useful for very large bucket or buckets with many small files where calculating local diffs takes a very long time.

cmd/hub/cli/billing.go
```diff
@@ -97,11 +97,17 @@ Use the --user flag to get usage for a dependent user.`,
...
+				ends := time.Unix(cus.GracePeriodEnd, 0).Format("Mon Jan _2 15:04:05 MST")
+				cmd.Warn(
+					"Grace period ending soon. Please add a payment method with `%s` before %s.",
+					aurora.Cyan("hub billing portal"),
+					aurora.Bold(ends))
+			} else {
+				cmd.Warn(
+					"Grace period ended. Please add a payment method with `%s` to continue using your account.",
+					aurora.Cyan("hub billing portal"))
+			}
```
sanderpick, line 110: More descriptive messages.

core/auth_interceptor.go
```diff
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
+	"github.com/golang-jwt/jwt"
```
sanderpick, line 8: Fix for the JWT security issue.

core/core.go
```diff
@@ -230,7 +236,12 @@ func NewTextile(ctx context.Context, conf Config, opts ...Option) (*Textile, err
 	}
 
 	// Configure clients
-	ic, err := httpapi.NewApi(conf.AddrIPFSAPI)
+	ipfsapi, err := httpapi.NewApi(conf.AddrIPFSAPI)
+	if err != nil {
+		return nil, err
+	}
+	// Don't allow using textile as a gateway to non-bucket files
+	ic, err := ipfsapi.WithOptions(caopts.Api.FetchBlocks(false))
```
sanderpick, line 244: Make the IPFS node "offline": don't ever ask the network for non-local (non-bucket) blocks.

core/core.go
```diff
@@ -471,6 +474,17 @@ func NewTextile(ctx context.Context, conf Config, opts ...Option) (*Textile, err
...
+				MaxConnectionIdle:     time.Hour * 24,
+				MaxConnectionAge:      time.Hour * 24,
+				MaxConnectionAgeGrace: time.Hour * 24,
+				Time:                  time.Hour * 2,
+				Timeout:               time.Hour * 2,
+			}),
+			grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
+				MinTime:             time.Minute * 5,
+				PermitWithoutStream: true,
+			}),
```
sanderpick, line 487: Some gRPC params to ensure our backend is not the cause of the `PROTOCOL ERROR` errors. See https://grpc.github.io/grpc/core/md_doc_keepalive.html

core/core.go
```diff
@@ -479,6 +493,7 @@ func NewTextile(ctx context.Context, conf Config, opts ...Option) (*Textile, err
 		return nil, err
 	}
 	go func() {
+		healthpb.RegisterHealthServer(t.server, health.NewServer())
```
sanderpick, line 496: Health check registration needed for my GKE experiment

core/usage_interceptor.go
```diff
@@ -38,9 +39,13 @@ func unaryServerInterceptor(pre preFunc, post postFunc) grpc.UnaryServerIntercep
 		}
 		res, err := handler(newCtx, req)
 		if err != nil {
+			perr := post(newCtx, info.FullMethod)
+			if perr != nil {
+				log.Errorf("unaryServerInterceptor postFunc: %v", err)
+			}
```
sanderpick, line 45: In the event there's an error returned from the gRPC handler, we still want to call the post usage function that submits usage info from the context.

core/usage_interceptor.go
```diff
@@ -60,8 +65,11 @@ func streamServerInterceptor(pre preFunc, post postFunc) grpc.StreamServerInterc
 		}
 		wrapped := grpcm.WrapServerStream(stream)
 		wrapped.WrappedContext = newCtx
-		err = handler(srv, wrapped)
-		if err != nil {
+		if err := handler(srv, wrapped); err != nil {
+			perr := post(newCtx, info.FullMethod)
+			if perr != nil {
+				log.Errorf("streamServerInterceptor postFunc: %v", err)
+			}
```
sanderpick, line 72: Same thing here, but for streams.

core/usage_interceptor.go
```diff
@@ -247,14 +255,17 @@ func (t *Textile) postUsageFunc(ctx context.Context, method string) error {
...
 		if _, err := t.bc.IncCustomerUsage(
-			ctx,
+			sctx,
 			account.Owner().Key,
 			map[string]int64{
 				"stored_data": owner.StorageDelta,
 			},
 		); err != nil {
-			return err
+			log.Infof("postUsageFunc: %v", err)
```
sanderpick, line 268: There's no point in returning a usage error to the user here since they have already "used" the resources. They will get kicked out during the next request.

cmd/buck/cli/cli.go
carsonfarmer: oh this is really nice!

core/core.go
carsonfarmer: Do we need to make an announcement or anything about this? It _will_ likely cause issues for _some_ folks I'm guessing?

carsonfarmer (approved):
This looks awesome to me @sanderpick, but I do think it needs a second pair of eyes, as I'm not totally familiar with all the stuff going on here.

core/core.go
sanderpick: Yep, good call. We'll have to note a couple things in public slack about this PR:
- offline mode
- billing undercharging
- health check return code is now 200 not 204 (I think slate makes use of the health check)

PR #572, opened 2021-12-26
Use latest go mongo driver

sbs2001:
The current version of mongo driver `v1.7.1`   has a [critical data race bug](https://jira.mongodb.org/browse/GODRIVER-2109). This also produces a big warning while installing hub. This PR uses newer version of the driver which is free of this bug.


sanderpick (approved):
nice, thank you 👍 some of the tests require secret keys, which your fork doesn't have access to, but everything else looks fine

